PR #8, opened 2017-06-20
chore(provider): use schema.Schema Description attribute

nicolai86:
I've seen other providers use this attribute and I guess this will be used in the future to generate the documentation. In any case having the Description inline doesn't hure


stack72:
Nice addition @nicolai86 :) This LGTM!

nicolai86:
🎉 ty

PR #10, opened 2017-06-21
chore(deps): downgrade ssh command to remove utils dependency

nicolai86:
restored the deps from terraform v0.9.5


stack72:
LGTM as a temporary solution to the failed build - we need to address this for the next version of the provider IMO

nicolai86:
Completely agreed , though I'm not 100% sure which direction this will go; a properly maintained fork of the scaleway sdk is probably the best option to allow the scaleway cli to evolve independently.

PR #15, opened 2017-08-10
vendor: github.com/hashicorp/terraform/...@v0.10.0

radeksimko:
```
govendor fetch github.com/hashicorp/terraform/...@v0.10.0
```


radeksimko:
@nicolai86 It's rather a regular maintenance across all providers. 😉 

The main motivation is https://github.com/hashicorp/terraform/pull/15562 (to prevent anyone from adding non compliant fields in the future) and bumping version in `User-Agent` (if there's such thing in the provider code) to 0.10.0. Ideally we should be reporting provider version, but until then `0.10` is better than `0.9.8` I think 😄

nicolai86:
Got it. Thank you for the details!

PR #20, opened 2017-10-07
add zero sized volumes to scaleway_server

### LINKED ISSUE #19: Secondry volume handling
Hi, I ran into an issue while trying to create a module to provision servers on Scaleway.

Provisioning a VC1S server works like a charm but when I try to provision a VC1M server I get the following error.

```
* scaleway_server.server: StatusCode: 400, Type: invalid_request_error, APIMessage: The total volume size of VC1M instances must be equal to 100GB
```
(side note: When I check the console the server has been created, with just the 50G volume)

Apparently there are some requirements when provisioning certain server types as described [here](https://github.com/scaleway/scaleway-cli/issues/361) and [here](https://github.com/scaleway/scaleway-cli/pull/462).

The VC1M needs a minimum of 100G of volumes attached. When creating a server via the console it automatically fills the gap by attaching the required volume.

The problem (for me at least)  is that I cannot use the volume parameter in the **scaleway_server** resource because that would prevent me from provisioning VC1S types as these do not allow extra volumes (50G is the maximum).

I tried to working around it by creating a **scaleway_volume** and **scaleway_volume_attachment** resource and only provision them based on the type (by using _count_ and mappings based on the type) but this doesn't prevent the error (maybe because the volume is attached after the server is create?).

Is there anyway for the provider to use the same logic as the console/scw? Ideally I would just specify the type and the provider attaches the necessary volume to meet the requirement. 

---
**Versions:**
Terraform v0.10.7
provider.scaleway ~> 0.1.1

nicolai86:
Add support for `volume` with `size_in_gb` of `0` in the `scaleway_server` resource - resolves #19 .

Now, the following terraform code is valid:

```
  name = "test"
  image = "some-guid"
  type = "VC1S"

  volume {
    size_in_gb = 0
    type = "l_ssd"
  }
}
```

previously, this would have errored because the volume size of 0 was disallowed. Now this is valid and it allows for the creation of terraform modules which span all available scaleway instance types.

The tests remain green, as this is only a small behaviour adjustment:

```
make testacc TESTARGS="-run='TestAccScalewayServer'"
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run='TestAccScalewayServer' -timeout 120m
?   	github.com/terraform-providers/terraform-provider-scaleway	[no test files]
=== RUN   TestAccScalewayServer_importBasic
--- PASS: TestAccScalewayServer_importBasic (128.65s)
=== RUN   TestAccScalewayServer_Basic
--- PASS: TestAccScalewayServer_Basic (123.43s)
=== RUN   TestAccScalewayServer_ExistingIP
--- PASS: TestAccScalewayServer_ExistingIP (105.07s)
=== RUN   TestAccScalewayServer_Volumes
--- PASS: TestAccScalewayServer_Volumes (143.62s)
=== RUN   TestAccScalewayServer_SecurityGroup
--- PASS: TestAccScalewayServer_SecurityGroup (133.63s)
PASS
ok  	github.com/terraform-providers/terraform-provider-scaleway/scaleway	634.434s
```

### DIFF
--- a/scaleway/resource_server.go
+++ b/scaleway/resource_server.go
@@ -61,9 +61,15 @@ func resourceScalewayServer() *schema.Resource {
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"size_in_gb": {
-							Type:         schema.TypeInt,
-							Required:     true,
-							ValidateFunc: validateVolumeSize,
+							Type:     schema.TypeInt,
+							Required: true,
+							ValidateFunc: func(v interface{}, k string) (ws []string, errors []error) {
+								value := v.(int)
+								if value > 150 {
+									errors = append(errors, fmt.Errorf("%q needs to be less than 150", k))
+								}
+								return
+							},
 						},
 						"type": {
 							Type:         schema.TypeString,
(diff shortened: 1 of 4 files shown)


nicolai86:
Optimistically pulling this in as tests are green.

scaleway/resource_server.go
```diff
@@ -61,9 +61,15 @@ func resourceScalewayServer() *schema.Resource {
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"size_in_gb": {
-							Type:         schema.TypeInt,
-							Required:     true,
-							ValidateFunc: validateVolumeSize,
+							Type:     schema.TypeInt,
+							Required: true,
```
tombuildsstuff: 🤔 another option would have been to make this field Optional instead?

nicolai86:
@tombuildsstuff would this work in the use case described in the linked ticket? Eg when using a map to set the value of an attribute - wouldn’t the attribute always be present, and I’d need to have a nil-check instead?

PR #30, opened 2017-12-13
retry on rate limit errors 

smola:
Scaleway API has a rate limit of 100 requests per second. When the limit is reached, the API replies with 429 HTTP status.

So, for every API request, we retry with exponential backoff whenever we find a 429 error.

PR #36, opened 2018-01-15
update scaleway-sdk dependency

nicolai86:
the SDK update allows custom HTTPClient implementations so we can implement
rate limited retry without touching all resources/ datasources.

related #30


nicolai86:
tests are green, pulling in the change to work on a cleaner rate limiting PR.

PR #39, opened 2018-02-18
detect hung servers by improving startup/ shutdown detection

### LINKED ISSUE #31: creating a server hangs up until 1h timeout if boot fails
### Terraform Version
`v0.10.0` (from provider tests)

### Affected Resource(s)
- scaleway_server

N/A: `make testacc`

### Expected Behavior
Tests pass.

### Actual Behavior
Tests creating a C1 server hang up if there is a boot problem.

### Steps to Reproduce
`make testacc`

### Important Factoids
I started experiencing this over the last weekend. Probably related to an incident with boot failures on C1 instances: https://status.online.net/index.php?do=details&task_id=1094

I'm having problems with boot again, with C1 instances stopping immediately after starting, `waitForServerState` never sees `running` state.

Maybe the provider should react and fail earlier if it sees `stopping` status after creating and powering on a server? Or maybe timeout should be lower than 1 hour?

nicolai86:
Explicitly declare valid pending states for startup/ shutdown. 
This will allow the proper detection of hung servers as `stopped` is no valid pending state for `startup`

closes #31

### DIFF
--- a/scaleway/helpers.go
+++ b/scaleway/helpers.go
@@ -59,7 +59,7 @@ func deleteRunningServer(scaleway *api.API, server *api.Server) error {
 		return err
 	}
 
-	return waitForServerState(scaleway, server.Identifier, "stopped")
+	return waitForServerShutdown(scaleway, server.Identifier)
 }
 
 // deleteStoppedServer needs to cleanup attached root volumes. this is not done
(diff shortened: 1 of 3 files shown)


nicolai86:
thank you for the review, @paultyng. 
From an UX perspective I'd rather fail within 15 seconds b/c Scaleway did not process the request yet than fail after 60 minutes because the server transition failed.

I've increased the delay to 15 seconds before checking to prevent false errors. Sadly Scaleway doesn't provide a server state history, so I can't differentiate between "state transition did not start yet" and "state transition failed and your stuck in initial state". Both things undesirable.

nicolai86:
Tests are green so moving this into master.

Issue #45, opened 2018-03-01
Feature request: allow adding ssh keys via the API


petems:
The provider currently has no way of adding SSH keys to your Scaleaway account.

Funnily enough, the official Scaleway [docs don't have instructions](https://developer.scaleway.com/#user-data), but there is this community post: 

https://community.online.net/t/api-account-ssh-keys/582/2

```
curl --header 'Content-Type: application/json' \
  --header 'X-Auth-Token: <YOUR TOKEN>' \
  https://account.cloud.online.net/users/<YOUR USER ID> \
  -X PATCH \
  -d '{"ssh_public_keys": [{"key": <YOUR KEY>}, {"key": <YOUR SECOND KEY>}]}'
```

petems:
As @paddycarver recommends, it would make sense to wait until this is officially documented to implement, as otherwise we'd end up in a situation supporting something that hasn't been officially "blessed" as a documented method.

nicolai86:
I think this would make a great addition to the provider!

petems:
I've reached out via a support ticket and their slack, hopefully I can get a response.

I might hack on trying to implement, then we can put a blocker on it until there's official docs.

nicolai86:
please add the SDK part as a PR to https://github.com/nicolai86/scaleway-sdk, as that's the Scaleway SDK in use here.

petems:
Oh cool, looks like it's already implemented:

https://github.com/nicolai86/scaleway-sdk/blob/master/user.go#L65

PR #46, opened 2018-03-15
update scaleway-sdk

nicolai86:
preparation to implement #45


nicolai86:
pulling this change in optimistically - as the modified functions are not used yet there is 0 risk.

PR #51, opened 2018-03-27
Fix example list

meyskens:
The example talks about the C2S which isn't an ARM server ;)


nicolai86:
Thank you for the correction, @meyskens !

PR #52, opened 2018-03-29
Fix disk size

meyskens:
Unfortunaly 20GB disk sizes ae not supported at Scaleway.


nicolai86:
Thanks again for another improvement to the documentation @meyskens !

PR #68, opened 2018-06-23
Update image.html.markdown

nicolaevladescu:
Fix typo


nicolai86:
Thank you for the correction!

PR #69, opened 2018-06-24
r/server: validate instance type availability

nicolai86:
Before creating a new `scaleway_server` instance, the availability is detected using the compute api specific `/products/servers/availability` API. 

Valid return values are (`available`, `scarce`, `shortage`), where `shortage` identifies out of stock.

In this case the provide will fail to create a new instance.

closes https://github.com/terraform-providers/terraform-provider-scaleway/issues/66 .


nicolai86:
tests are still green so pulling this change in.

PR #72, opened 2018-06-28
support concurrent server creation

### LINKED ISSUE #65: Server creation appears to happen sequentially
### Terraform Version
Terraform v0.11.7
+ provider.scaleway v1.4.1

### Affected Resource(s)
- scaleway_server

[tf.zip](https://github.com/terraform-providers/terraform-provider-scaleway/files/2062095/tf.zip)

### Expected Behavior
Servers are created in parallel

### Actual Behavior
Servers are created synchronously, making `tf apply` mega slow

### Steps to Reproduce
1. `terraform apply`

nicolai86:
instead of blocking for the entire resource creation duration, only
block for the actual call to create/ update the scaleway_server
resource; do not block while polling for the resource state.

this should speed up creating multiple servers considerably.

according to @sieben this should be safe now.

closes #65

### DIFF
--- a/scaleway/resource_server.go
+++ b/scaleway/resource_server.go
@@ -160,9 +160,6 @@ func attachIP(scaleway *api.API, serverID, IPAddress string) error {
 func resourceScalewayServerCreate(d *schema.ResourceData, m interface{}) error {
 	scaleway := m.(*Client).scaleway
 
-	mu.Lock()
-	defer mu.Unlock()
-
 	image := d.Get("image").(string)
 	var req = api.ServerDefinition{
 		Name:          d.Get("name").(string),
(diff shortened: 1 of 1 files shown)


nicolai86:
all tests are green so I'm pulling this in optimistically.

meyskens:
Not sure if related but since this was released i tried to create 3 servers and every time only 1/3 booted the rest set itself to an off state, went back to 1.4.1 and it worked fine. If you prefer i can make a seperate issue for this.

nicolai86:
@meyskens this is most definitively relates. I’m so sorry - can you create a new issue to track this?

nicolai86:
@meyskens if you open a bug please include a template which has this issue - I tried reproducing it locally with this template, but it worked without any issues:

```
provider "scaleway" {}

  name = "test"

  count = 5
  image = "5faef9cd-ea9b-4a63-9171-9e26bec03dbc"
  type  = "C1"
  tags  = ["terraform-test"]
}
```

nicolai86:
after some digging it seems to be related to `scaleway_volume_attachment` usage. investigating further...

Issue #67, opened 2018-06-21
SCALEWAY_ORGANIZATION is not the Token Access Key


deurk:
The documentation (https://www.terraform.io/docs/providers/scaleway/) states that to use the provider, one needs to set the environment variable `SCALEWAY_ORGANIZATION` to the Token Access Key. This is incorrect as it returns the following error when trying to use it:

`StatusCode: 400, Type: invalid_request_error, APIMessage: Validation Error, Details: map[organization:[<MY_SCALEWAY_TOKEN_ACCESS_KEY> is not a valid UUID.]]`

One actually needs to retrieve the UUID to use as the correct `SCALEWAY_ORGANIZATION` environment variable using the following command: 

`curl https://account.scaleway.com/organizations -H "X-Auth-Token: <TOKEN_SECRET_KEY>"`

The request will return a JSON result in which you will find the `id` within `organizations` to use.

nicolai86:
fixed in #75

PR #75, opened 2018-07-07
Clarify the credentials used for the provider

eallrich:
As captured in #67 the `organization` value for the provider is the organization ID of the account, not an access key. Since I had just experienced the same error when I followed the documentation, I've attempted to clarify how to get both the organization ID as well as the secret token.

Thank you to @deurk for providing the `curl` command to get the organization ID.


nicolai86:
looks good to me. thank you for tackling this, @eallrich !

PR #79, opened 2018-07-14
Docs: fixing the sidebar highlight

tombuildsstuff:
Noticed the sidebar didn't highlight correctly for the Security Groups / Volume attachment resources

screenshot 2018-07-14 at 09 17 35

This PR also adds the SSH Key Resource to the sidebar, since it's currently missing

### DIFF
--- a/website/docs/r/security_group.html.markdown
+++ b/website/docs/r/security_group.html.markdown
@@ -1,7 +1,7 @@
 ---
 layout: "scaleway"
 page_title: "Scaleway: security_group"
-sidebar_current: "docs-scaleway-resource-security_group"
+sidebar_current: "docs-scaleway-resource-security-group-x"
 description: |-
   Manages Scaleway security groups.
 ---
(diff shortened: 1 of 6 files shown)


website/docs/r/security_group.html.markdown
```diff
@@ -1,7 +1,7 @@
 ---
 layout: "scaleway"
 page_title: "Scaleway: security_group"
-sidebar_current: "docs-scaleway-resource-security_group"
+sidebar_current: "docs-scaleway-resource-security-group-x"
```
nicolai86: why the `-x` suffix?

tombuildsstuff: otherwise `...-security-group-rule` causes this highlight to be caught; we tend to use `-x` as a suffix to work around this (but any string will do)

nicolai86: Thank you for the explanation :)

PR #91, opened 2018-08-26
fix token import/ subsequent read

nicolai86:
the `secret_key` is not present in subsequent read operations, so we set
it after creation and fallback to the old semantics of using the ID if
it's not set.

we'll probably need to support a different way if important a token in
the future.


nicolai86:
all tests are green, merging

PR #94, opened 2018-09-13
provider/scaleway: add scaleway_bucket resource

nicolai86:
This PR adds a new resource to the Scaleway provider: `scaleway_bucket`.

This can be merged once Scaleway objectstorage is out of GA.

```
make testacc TESTARGS="-run='TestAccScalewayBucket'"
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run='TestAccScalewayBucket' -timeout 120m
?   	github.com/terraform-providers/terraform-provider-scaleway	[no test files]
=== RUN   TestAccScalewayBucket
--- PASS: TestAccScalewayBucket (13.35s)
PASS
ok  	github.com/terraform-providers/terraform-provider-scaleway/scaleway
```

The PR includes docs, tests as well as the resource itself.

### DIFF
--- a/scaleway/resource_bucket.go
+++ b/scaleway/resource_bucket.go
@@ -1,0 +1,69 @@
+package scaleway
+
+import (
+	"log"
+
+	"github.com/hashicorp/terraform/helper/schema"
+
+	api "github.com/nicolai86/scaleway-sdk"
+)
+
+func resourceScalewayBucket() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceScalewayBucketCreate,
+		Read:   resourceScalewayBucketRead,
+		Delete: resourceScalewayBucketDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+				Description: "The name of the bucket",
+			},
+		},
+	}
+}
+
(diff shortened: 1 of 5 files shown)


scaleway/resource_bucket.go
```diff
@@ -0,0 +1,61 @@
...
+				Description: "The name of the bucket",
+			},
+		},
+	}
+}
+
+func resourceScalewayBucketRead(d *schema.ResourceData, m interface{}) error {
+	scaleway := m.(*Client).scaleway
+
+	_, err := scaleway.ListObjects(d.Get("name").(string))
```
tombuildsstuff: so if this returns a not found error (e.g. a 404) we should clear the ID so this is detected and can be recreated e.g.

```
if err != nil {
  if resp.StatusCode == 404 {
    log.Printf("[DEBUG] Bucket %q was not found - removing from state!", name)
    d.SetId("")
    return nil
  }
}
```

scaleway/resource_bucket.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+func resourceScalewayBucketDelete(d *schema.ResourceData, m interface{}) error {
+	scaleway := m.(*Client).scaleway
+
+	err := scaleway.DeleteBucket(d.Id())
+	if err != nil {
+		return err
+	}
+	d.SetId("")
```
tombuildsstuff: we don't need to clear the ID here

nicolai86:
@tombuildsstuff thanks for taking a look. I've addressed the feedback - would you mind another look?

PR #96, opened 2018-10-24
introduce new scaleway_ip_reverse_dns resource

nicolai86:
some usecases like dynamic reverse dns result in circular dependencies
between `scaleway_ip` and `scaleway_server` resources.

To break this cycle the `reverse` attribute on `scaleway_ip` is going to
be deprecated, and instead a new `scaleway_ip_reverse_dns` resource is
made available, which allows the IP to be created before the reverse DNS
is persisted.

Tests for the new resource as well as old modified resource are green.

Once merged this will fix use cases like https://github.com/terraform-providers/terraform-provider-scaleway/issues/88

### DIFF
--- a/scaleway/resource_ip.go
+++ b/scaleway/resource_ip.go
@@ -32,6 +32,8 @@ func resourceScalewayIP() *schema.Resource {
 			"reverse": {
 				Type:        schema.TypeString,
 				Optional:    true,
+				Computed:    true,
+				Deprecated:  "use scaleway_ip_reverse_dns resource instead",
 				Description: "The ipv4 reverse dns",
 			},
 		},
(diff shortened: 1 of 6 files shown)


scaleway/resource_ip.go
```diff
@@ -32,6 +32,8 @@ func resourceScalewayIP() *schema.Resource {
 			"reverse": {
 				Type:        schema.TypeString,
 				Optional:    true,
+				Computed:    true,
+				Deprecated:  "use scaleway_ip_reverse_dns resource instead",
```
paultyng: You should modify the docs for `scaleway_ip` as well.

nicolai86: good point, I’ll add the deprecation notice there too.

nicolai86: updated! @paultyng PTAL

PR #107, opened 2019-01-14
Fix bullets and formatting in docs

alvin-huang:
These are just a few small changes in the docs to fix bullets so they render properly and a few formatting fixes.

Currently it shows the `+` instead of as a bulleted list. 
image


nicolai86:
Thanks for this improvement!

PR #110, opened 2019-02-14
Ignore leading and trailing whitespaces when adding public SSH key

azuenko:
Hello,

A small change to avoid using `trimspace()` Terraform function when adding keys generated by `ssh-keygen` and by Terraform TLS provider. Both generate keys with trailing newline character, which makes them to be rejected by Scaleway API. For details see https://community.scaleway.com/t/proposal-ssh-key-add-api-trim-spaces-on-server-side/7376


abitrolly:
Fixed in 1.9.0 https://github.com/terraform-providers/terraform-provider-scaleway/blob/master/CHANGELOG.md#190-march-15-2019

PR #113, opened 2019-03-15
Return the env org value instead of the key

Sh4d1:
Hey, instead of returning the `SCALEWAY_ORGANIZATION` env value, the key is returned.


nicolai86:
Thanks for the bugfix!

nicolai86:
v1.9.1 was released yesterday, so auth via env variables was only broken for a short period. Please upgrade ASAP :)

PR #118, opened 2019-03-24
sort images in a descending way

### LINKED ISSUE #117: scaleway_image: most_recent attribute its not working as expected
Hi there,

The sort behavior used by `scaleway_image` is [ascending](https://github.com/terraform-providers/terraform-provider-scaleway/blob/master/scaleway/data_source_image.go#L127). If we want to select the most recent image this sort has to be [descending](https://golang.org/pkg/time/#Time.After).

### Terraform Version
`Terraform v0.11.13` 

### Affected Resource(s)
- scaleway_image

```hcl
data "scaleway_image" "debian" {
  architecture = "x86_64"
  name         = "scw-node-v1-2019-03-23"
  most_recent  = "true"
}
```

### Expected Behavior
Return most recent image if multiple exist

### Actual Behavior
Return the first image created

bugbuilder:
Closes #117

### DIFF
--- a/scaleway/data_source_image.go
+++ b/scaleway/data_source_image.go
@@ -124,7 +124,7 @@ func dataSourceScalewayImageRead(d *schema.ResourceData, meta interface{}) error
 
 	mostRecent := d.Get("most_recent").(bool)
 	sort.Slice(images, func(i, j int) bool {
-		return images[i].ModificationDate.Before(*images[j].ModificationDate)
+		return images[i].ModificationDate.After(*images[j].ModificationDate)
 	})
 	log.Printf("[DEBUG] got %d images: %#v\n", len(images), images)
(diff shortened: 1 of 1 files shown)


nicolai86:
thanks for this bug fix, @bugbuilder .

Do you think there's a way we could add tests to prevent this from regressing?

bugbuilder:
One way if we add some kind of integration test using mock API or a CI environment hitting the real API.

Other way is just move the sort to an exported function and then add the function test to `data_source_image_test`.

IMHO the change is so minimal that no require to do any of that.

nicolai86:
Agreed on the size of the CR - I just remember this being changed before so I want to prevent toggle-CRs. 
Let’s change that eventually!

PR #132, opened 2019-05-29
refactor: update internal meta struct

### LINKED ISSUE #131: Refactor the configuration interface
## The Problem

Until now, the Scaleway terraform provider used the @nicolai86 [Scaleway SDK](https://github.com/nicolai86/scaleway-sdk). I thank the opportunity to thanks him for this great help! 👏

We have just released our new [Go SDK](https://github.com/scaleway/scaleway-sdk-go) and we want to use it in the implementation of future new resources (cf. #125 Step 1 - item 5 and 6).
Moreover, in order to avoid breaking changes during the [Step 1](https://github.com/terraform-providers/terraform-provider-scaleway/issues/125) we want to keep the first version of the SDK for all the deprecated resources.

## What that mean?

We need the two SDKs for the `>= v1.10.0` in a single shared structure. The old SDK will be considered deprecated and therefore will be removed - like any other deprecated elements - in the `v2.0.0`.

For the moment, here is our [configuration interface](https://github.com/terraform-providers/terraform-provider-scaleway/blob/master/vendor/github.com/hashicorp/terraform/helper/schema/provider.go#L71-L74):

```go
// Client contains scaleway api clients
type Client struct {
	scaleway *sdk.API
}
```

We have to rename it into something that is more abstracted - e.g. `Config` - that will contain the SDK clients:

```go
// Meta contains SDK clients used by resources.
//
// This meta value is passed into all resources.
type Meta struct {
        // In near futur: The new Scaleway SDK Client.
	scwClient        *scw.Client
        // In near futur: The AWS S3 Client that will be used for S3 resources.
        // Could be init by an `S3Client` method only when an S3 resource is used.
        s3Client         *s3.S3
        // The deprecated Scaleway SDK (will be removed in `v2.0.0`).
	deprecatedClient *api.API
}
```

QuentinBrosse:
Fix #131

### DIFF
--- a/scaleway/config.go
+++ b/scaleway/config.go
@@ -3,6 +3,8 @@ package scaleway
 import (
 	"bytes"
 	"context"
+	"encoding/json"
+	"fmt"
 	"io"
 	"io/ioutil"
 	"log"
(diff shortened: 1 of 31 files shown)


scaleway/config.go
```diff
@@ -23,16 +25,44 @@ type Config struct {
...
+	// Deprecated: The deprecated Scaleway SDK (will be removed in `v2.0.0`).
+	deprecatedClient *sdk.API
 }
 
-// client is a bridge between sdk.HTTPClient interface and retryablehttp.Client
+// Meta creates a meta instance from a client configuration.
+func (c *Config) Meta() (*Meta, error) {
+	meta := &Meta{}
+
+	deprecatedClient, err := c.GetDeprecatedClient()
```
nicolai86: what do you think about moving the code into a package and importing it as `deprecated.Meta()`? This way you don't need to rename everything and can call the deprecated methods in your new provider structs?

(recurs)

QuentinBrosse: It's a good idea but:
- Terraform recommends keeping all codebase in the `main` package
- This deprecated code will be replaced in a month at maximum

So we will keep a single package.

PR #160, opened 2019-06-12
feat: add helpers

jerome-quere:


jerome-quere:
helpers are now private

PR #162, opened 2019-06-12
feat: add getRegion helper

jerome-quere:


scaleway/helpers.go
```diff
@@ -281,7 +281,25 @@ func getZone(d terraformResourceData, meta *Meta) (utils.Zone, error) {
...
+	if exist {
+		return utils.ParseRegion(rawRegion.(string))
+	}
+
+	region, exist := meta.scwClient.GetDefaultRegion()
+	if exist {
+		return region, nil
+	}
+
+	return utils.Region(""), fmt.Errorf("could not detect region")
```
nicolai86: Nit: since the error does not have dynamic data consider declaring it:

ErrRegionNotFound 

Or similar

PR #172, opened 2019-06-20
feat: use instance helpers

kindermoumoute:


scaleway/compute_instance_helpers.go
```diff
@@ -0,0 +1,25 @@
...
+func getInstanceAPIWithZone(d *schema.ResourceData, m interface{}) (*instance.API, utils.Zone, error) {
+	meta := m.(*Meta)
+	instanceApi := instance.NewAPI(meta.scwClient)
+
+	zone, err := getZone(d, meta)
+	return instanceApi, zone, err
+}
+
+// getInstanceAPIWithZoneAndID returns an instance API with zone and ID extracted from the state
+func getInstanceAPIWithZoneAndID(m interface{}, id string) (*instance.API, utils.Zone, string, error) {
```
QuentinBrosse: If we need another data in the future the method will be called `getInstanceAPIWithZoneAndIDAndPlopAndPlip`? :/

What do you think about something like `getInstanceAPIAndGenerics`? It could return a `struct` if we need more data in the future.

DenBeke: Couldn't we store that in the meta object somewhere?

jerome-quere: This is a internal that we are almos certainly going to change.....
Let's pick a simple solution an update later if needed

PR #173, opened 2019-06-20
test: run tests in parallel

kindermoumoute:


scaleway/helpers_test.go
```diff
@@ -87,12 +87,6 @@ func TestParseZonedID(t *testing.T) {
 			id:         "my/id",
 			zone:       utils.ZoneFrPar1,
 		},
-		{
-			name:       "id with a region",
-			localityId: "fr-par/my-id",
-			id:         "my-id",
-			zone:       utils.Zone("fr-par"),
```
kindermoumoute: this test was provoking a warning (`"fr-par"` is not a valid zone)

QuentinBrosse: 👍

PR #164, opened 2019-06-12
feat: added resource scaleway_compute_instance_volume

DenBeke:

PR #174, opened 2019-06-20
feat: scaleway_compute_instance_server

kindermoumoute:


scaleway/resource_compute_instance_server.go
```diff
@@ -0,0 +1,297 @@
...
+			//				Computed: true,
+			//			},
+			//		},
+			//	},
+			//	Description: "Root volume attached to the server on creation",
+			//},
+			"enable_ipv6": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Default:     false,
```
jerome-quere: default true ?

scaleway/resource_compute_instance_server.go
```diff
@@ -0,0 +1,297 @@
...
+					"standby",
+				}, false),
+			},
+			"user_data": {
+				Type: schema.TypeMap,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+				Optional:    true,
+				Description: "The user data associated with the server", // TODO: document reserved keys (`cloud-init`)
```
jerome-quere: Maybe we should move cloud-init to a specific server property as this would trigger a forceNew

scaleway/resource_compute_instance_server.go
```diff
@@ -0,0 +1,297 @@
...
+	if err != nil && !is404Error(err) {
+		return err
+	}
+
+	// TODO: if root volume remove_on_terminate
+
+	return nil
+}
+
+func stateToAction(state string) instance.ServerAction {
```
jerome-quere: This function name should be more verbose as we have only one package

scaleway/resource_compute_instance_server.go
```diff
@@ -0,0 +1,297 @@
...
+	d.Set("image", response.Server.Image.ID)
+	d.Set("type", response.Server.CommercialType)
+	d.Set("tags", response.Server.Tags)
+	d.Set("security_group_id", response.Server.SecurityGroup.ID)
+
+	d.Set("enable_ipv6", response.Server.EnableIPv6)
+	d.Set("private_ip", *response.Server.PrivateIP)
+	d.Set("public_ip", response.Server.PublicIP.Address.String())
+
+	if response.Server.EnableIPv6 && response.Server.IPv6 != nil {
```
jerome-quere: What If ipv6 was enable and we disable it we should still Set public_ipv6 to ""

scaleway/resource_compute_instance_server.go
```diff
@@ -0,0 +1,297 @@
...
+	}
+
+	if d.HasChange("tags") {
+		updateRequest.Tags = utils.Strings(d.Get("tags").([]string))
+	}
+
+	if d.HasChange("security_group_id") {
+		updateRequest.SecurityGroup = &instance.SecurityGroupSummary{
+			ID:   d.Get("security_group_id").(string),
+			Name: getRandomName("sg"),
```
jerome-quere: Add a comment to specify that this value will be ignored by the API

scaleway/resource_compute_instance_server.go
```diff
@@ -0,0 +1,297 @@
...
+
+	d.Set("name", response.Server.Name)
+	d.Set("image", response.Server.Image.ID)
+	d.Set("type", response.Server.CommercialType)
+	d.Set("tags", response.Server.Tags)
+	d.Set("security_group_id", response.Server.SecurityGroup.ID)
+
+	d.Set("enable_ipv6", response.Server.EnableIPv6)
+	d.Set("private_ip", *response.Server.PrivateIP)
+	d.Set("public_ip", response.Server.PublicIP.Address.String())
```
jerome-quere: "state" should be set  or the `HasChange` in update may never be true

scaleway/resource_compute_instance_server_test.go
```diff
@@ -0,0 +1,181 @@
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+func testAccCheckScalewayComputeInstanceServerConfigMinimal1() string {
+	return `
+resource "scaleway_compute_instance_server" "webserver" {
```
jerome-quere: May be indented

scaleway/resource_compute_instance_server_test.go
```diff
@@ -0,0 +1,181 @@
...
+
+resource "scaleway_compute_instance_server" "base" {
+  name = "test"
+  image = "${data.scaleway_image.ubuntu.id}"
+  type = "%s"
+  tags = [ "terraform-test", "scaleway_compute_instance_server", "basic1" ]
+}`, architecture, serverType)
+}
+
+// todo: add tests with IP attachment
```
jerome-quere: No need This will be tested by IP

scaleway/resource_compute_instance_server_test.go
```diff
@@ -0,0 +1,181 @@
...
+
+// todo: add tests with additional volume attachement
```
jerome-quere: No need This will be tested by volume

scaleway/resource_compute_instance_server.go
kindermoumoute: True, it should ForceReboot instead of ForceNew though

scaleway/compute_instance_helpers.go
```diff
@@ -23,3 +23,18 @@ func getInstanceAPIWithZoneAndID(m interface{}, id string) (*instance.API, utils
 	zone, ID, err := parseZonedID(id)
 	return instanceApi, zone, ID, err
 }
+
+func flattenRootVolume(v interface{}) []map[string]interface{} {
```
DenBeke: Would be nice to have godoc here.

scaleway/compute_instance_helpers.go
```diff
@@ -23,3 +23,18 @@ func getInstanceAPIWithZoneAndID(m interface{}, id string) (*instance.API, utils
...
+	flattenVolume := []map[string]interface{}{{}}
+
+	vs, ok := v.([]map[string]interface{})
+	if ok {
+		flattenVolume = vs
+	}
+
+	if _, exist := flattenVolume[0]["delete_on_termination"]; !exist {
```
DenBeke: can` flattenVolume` be empty? This might segfault.

scaleway/resource_compute_instance_server.go
```diff
@@ -0,0 +1,360 @@
...
+		},
+		SchemaVersion: 0,
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Computed:    true,
+				Description: "The name of the server",
+			},
+			"image": {
```
DenBeke: You use `security_group_id` but here not `image_id`. Is this done on purpose?

scaleway/resource_compute_instance_server.go
```diff
@@ -0,0 +1,360 @@
...
+				Description: "The security group the server is attached to",
+			},
+			"root_volume": {
+				Type:     schema.TypeList,
+				MaxItems: 1,
+				Optional: true,
+				Computed: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"size_in_gb": {
```
DenBeke: You might use human readable size format here as I did in #164. But since you commented on my size in bytes we better discuss this together :)

scaleway/resource_compute_instance_server.go
```diff
@@ -0,0 +1,360 @@
...
+							Type:     schema.TypeInt,
+							Optional: true,
+							Computed: true,
+							ForceNew: true, // TODO: don't force new but stop server and create new volume instead
+						},
+						"delete_on_termination": {
+							Type:     schema.TypeBool,
+							Optional: true,
+							Default:  true,
```
DenBeke: Do you really want this to default to true? It's not the behaviour of the console. And it can make people lose data.

scaleway/resource_compute_instance_server.go
DenBeke: IPv6 support is not really native. Some default to `false` is better imho.

scaleway/resource_compute_instance_server.go
```diff
@@ -0,0 +1,360 @@
...
+	if err != nil {
+		if is404Error(err) {
+			d.SetId("")
+			return nil
+		}
+		return err
+	}
+	switch response.Server.State {
+	case instance.ServerStateStopped:
+		d.Set("state", "stopped")
```
DenBeke: Maybe we can create constants for the states? Because we use them on multiple places...

scaleway/resource_compute_instance_server.go
```diff
@@ -0,0 +1,360 @@
...
+		})
+	}
+
+	if response.Server.EnableIPv6 && response.Server.IPv6 != nil {
+		d.Set("public_ipv6", response.Server.IPv6.Address.String())
+	}
+
+	if vs, ok := response.Server.Volumes["0"]; ok {
+		rootVolume := flattenRootVolume(d.Get("root_volume"))
+		rootVolume[0]["volume_id"] = vs.ID
```
DenBeke: can this segfault?

scaleway/resource_compute_instance_server.go
```diff
@@ -0,0 +1,360 @@
...
+	}
+
+	if d.HasChange("state") {
+		previousState, nextState := d.GetChange("state")
+		for _, action := range stateToAction(previousState.(string), nextState.(string)) {
+			err = instanceApi.ServerActionAndWait(&instance.ServerActionAndWaitRequest{
+				Zone:     zone,
+				ServerID: ID,
+				Action:   action,
+				Timeout:  time.Minute * 10,
```
DenBeke: Can we put the timeout in a constant somewhere? Might be necessary some day to change it everywhere. Or even put it in the config. But a constant will do for now.

scaleway/resource_compute_instance_server_test.go
```diff
@@ -0,0 +1,266 @@
...
+		// Unexpected api error we return it
+		if !is404Error(err) {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func testAccCheckScalewayComputeInstanceServerConfigMinimal() string {
```
DenBeke: Might be more logica to put the test cases at the top of the file. And why create a function to return them of we can just put them in a var?

@jerome-quere and I put them at the top of the file, and created arrays for tests that have multiple steps. Look in `resource_compute_instance_ip_test.go` for an example.

scaleway/resource_compute_instance_server_test.go
DenBeke: That's not really a functionally we have on `scaleway_compute_instance_volume` resource. Or are you talking about creating a volume attachment resource?

scaleway/compute_instance_helpers.go
kindermoumoute: It's never empty (flattenVolume[0] is set line 28)

scaleway/resource_compute_instance_server.go
kindermoumoute: Good catch!

scaleway/resource_compute_instance_server.go
kindermoumoute: It is the behavior of the console.

scaleway/resource_compute_instance_server.go
kindermoumoute: Nope

scaleway/compute_instance_helpers.go
DenBeke: My bad, `{{}}` it easily missed when reading the code? :p

scaleway/resource_compute_instance_server.go
DenBeke: As discussed IRL we will stick to `size_in_gb`.

scaleway/compute_instance_helpers.go
kindermoumoute: Is it better this way?

DenBeke: Yeps!

scaleway/resource_compute_instance_server.go
kindermoumoute: let's keep default false for now

scaleway/resource_compute_instance_server.go
kindermoumoute: let's keep default false for now

PR #177, opened 2019-06-25
feat: prepare instance PR

kindermoumoute:


scaleway/config.go
```diff
@@ -114,7 +114,7 @@ func createRetryableHTTPClient(shouldLog bool) *client {
 	c.RetryMax = 3
 	c.RetryWaitMax = 2 * time.Minute
 	c.Logger = log.New(os.Stderr, "", 0)
-	c.RetryWaitMin = time.Minute
+	c.RetryWaitMin = time.Second * 2
```
QuentinBrosse: You don't want to revert it?

kindermoumoute: Let's keep the min low and increase it if we need (max remains 2 minutes)

PR #192, opened 2019-07-02
feat: placement group

kindermoumoute:


scaleway/resource_compute_instance_placement_group.go
```diff
@@ -0,0 +1,146 @@
...
+			"name": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Computed:    true,
+				Description: "The name of the placement group",
+			},
+			"policy_type": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Default:     instance.ComputeClusterPolicyTypeLowLatency.String(),
```
jerome-quere: I would not put any default value, it's an advance feature and none of the default value make sense.

jerome-quere: same for `policy_mode`

kindermoumoute: This is already the default on the API side and it's well documented https://developers.scaleway.com/en/products/instance/api/#post-2bb6da

PR #194, opened 2019-07-03
doc: scaleway_storage_object_bucket

DenBeke:


website/docs/r/storage_object_bucket.html.markdown
```diff
@@ -0,0 +1,47 @@
...
+    acl = "private"
+}
+```
+
+## Arguments Reference
+
+The following arguments are supported:
+
+* `name` - The name of the bucket.
+* `acl` - (Optional) The canned ACL you want to apply to the bucket.
```
QuentinBrosse: Add a `TODO` comment:
```md
[//]: # (TODO: Add a link to doc when it will exist.)
```

website/docs/r/storage_object_bucket.html.markdown
```diff
@@ -0,0 +1,47 @@
...
+    name = "some-unique-name"
+    acl = "private"
+}
+```
+
+## Arguments Reference
+
+The following arguments are supported:
+
+* `name` - The name of the bucket.
```
QuentinBrosse: (Required)

website/docs/r/storage_object_bucket.html.markdown
```diff
@@ -0,0 +1,47 @@
...
+* `acl` - (Optional) The canned ACL you want to apply to the bucket.
+* `region` - (Optional) The [region](https://developers.scaleway.com/en/quickstart/#region-definition) you want to attach this resource to.
```
QuentinBrosse: Like [here](https://github.com/terraform-providers/terraform-provider-scaleway/pull/193/files#diff-5c95b28312e3fd1bb991cb7cd9561937R153): `The [region](https://developers.scaleway.com/en/quickstart/#region-definition) in which the server should be created.`

website/docs/r/storage_object_bucket.html.markdown
```diff
@@ -0,0 +1,47 @@
...
+
+## Attributes Reference
+
+The following attributes are exported:
+
+* `id` - The ID of the bucket.
+* `name` - The name of the bucket.
```
QuentinBrosse: Remove `name`, `acl` and `region`

website/scaleway.erb
```diff
@@ -73,6 +73,16 @@
           </ul>
         </li>
 
+        <li<%= sidebar_current("docs-scaleway-resource-storage") %>>
```
jerome-quere: check with @QuentinBrosse but i think `sidebar_current` is deprecated

DenBeke: Yeps, that's correct. I must update this PR since we have merged Quentin's.

website/docs/r/storage_object_bucket.html.markdown
```diff
@@ -0,0 +1,45 @@
...
+* `name` - (Required) The name of the bucket.
+* `acl` - (Optional) The canned ACL you want to apply to the bucket.
+[//]: # (TODO: Add a link to doc when it will exist.)
+* `region` - (Optional) The [region](https://developers.scaleway.com/en/quickstart/#region-definition) in which the bucket should be created.
+
+## Attributes Reference
+
+In addition to all arguments above, the following attributes are exported:
+
+* `id` - The ID of the bucket.
```
jerome-quere: Should not be documented

QuentinBrosse: This regionalized it should be exported (like in any other resources). Il could be useful in the future to create resource dependencies.

PR #209, opened 2019-07-11
doc: compute_instance_placement_group

QuentinBrosse:


website/docs/r/compute_instance_placement_group.html.markdown
```diff
@@ -0,0 +1,41 @@
...
+- `project_id` - (Defaults to [provider](../index.html#project_id) `project_id`) The ID of the project the placement group is associated with.
+
+## Attributes Reference
+
+In addition to all arguments above, the following attributes are exported:
+
+- `id` - The ID of the placement group.
+- `policy_respected` - Is true when the policy is respected.
+
+## Import
```
jerome-quere: Did you try to look if there is an include like functionality in templates?

QuentinBrosse: It is not easily impossible...

PR #215, opened 2019-07-16
Try to parse the value security_group_id as a zoned ID 

mekahell:
The value of security_group_id can be a zoned ID instead of being a regular uuid


mekahell:
Hi Jerome, I've done the modifications

PR #218, opened 2019-07-17
fix: scaleway_compute_instance_ip read server_id

QuentinBrosse:


scaleway/resource_compute_instance_ip.go
```diff
@@ -106,6 +106,10 @@ func resourceScalewayComputeInstanceIPRead(d *schema.ResourceData, m interface{}
 	d.Set("zone", string(zone))
 	d.Set("project_id", res.IP.Organization)
 	d.Set("reverse", res.IP.Reverse)
+	
+	if res.IP.Server != nil {
+		d.Set("server_id", res.IP.Server.ID)
+	}
```
jerome-quere: Should be set to `""` when `res.IP.Server` is nill

PR #219, opened 2019-07-17
fix: reboot on placement group change

Sh4d1:
When assigning a running server to a placment group, it has to be rebooted to take effect

### DIFF
--- a/scaleway/resource_compute_instance_server.go
+++ b/scaleway/resource_compute_instance_server.go
@@ -381,6 +381,8 @@ func resourceScalewayComputeInstanceServerUpdate(d *schema.ResourceData, m inter
 		return err
 	}
 
+	var forceReboot bool
+
 	////
 	// Update the server
 	////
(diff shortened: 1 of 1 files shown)


scaleway/resource_compute_instance_server.go
```diff
@@ -427,6 +429,7 @@ func resourceScalewayComputeInstanceServerUpdate(d *schema.ResourceData, m inter
 		if placementGroupID == "" {
 			updateRequest.ComputeCluster = &instance.NullableStringValue{Null: true}
 		} else {
+			forceReboot = true
```
QuentinBrosse: I think we need to do it just after the `if d.HasChange("placement_group_id")`.

Sh4d1: We could, but if we detach the instance from the placement group, we have no rules of placement so the current placement doesn't really matter anymore right? The reboot will be useless

QuentinBrosse: Ok

PR #224, opened 2019-07-23
test: fix TestAccScalewayServer_BootType

QuentinBrosse:


scaleway/resource_server_test.go
```diff
@@ -344,7 +344,7 @@ data "scaleway_image" "ubuntu" {
 resource "scaleway_server" "base" {
   name = "test"
   image = "${data.scaleway_image.ubuntu.id}"
-  type = "DEV1-S"
+  type = "START1-S"
```
jerome-quere: Can you add a comment:

`This type is deprecated but as this resource will be removed in next major release we kept it`

PR #235, opened 2019-09-03
feat: enable bssd volumes

kindermoumoute:


scaleway/resource_instance_volume.go
```diff
@@ -34,7 +34,7 @@ func resourceScalewayInstanceVolume() *schema.Resource {
 				ForceNew:    true,
```
QuentinBrosse: We need to add `Optional:    true` and `Default: instance.VolumeTypeBSSD.String()`. Cf. [here](https://github.com/terraform-providers/terraform-provider-scaleway/pull/216/files#diff-136bae47c996c61549b8078d867e3cacR33)

PR #236, opened 2019-09-05
fix: set dynamic IP by default

kindermoumoute:
This will allow remote exec to work by default

### DIFF
--- a/scaleway/resource_instance_server.go
+++ b/scaleway/resource_instance_server.go
@@ -124,6 +124,12 @@ func resourceScalewayInstanceServer() *schema.Resource {
 				Computed:    true,
 				Description: "The public IPv4 address of the server",
 			},
+			"disable_public_ip": {
+				Type:        schema.TypeBool,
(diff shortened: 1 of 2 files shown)


scaleway/resource_instance_server.go
```diff
@@ -208,7 +214,7 @@ func resourceScalewayInstanceServerCreate(d *schema.ResourceData, m interface{})
 		CommercialType:    commercialType,
 		EnableIPv6:        d.Get("enable_ipv6").(bool),
 		SecurityGroup:     expandID(d.Get("security_group_id")),
-		DynamicIPRequired: Bool(false),
+		DynamicIPRequired: Bool(!d.Get("disable_public_ip").(bool)),
```
jerome-quere: why not using scw.BoolPtr ?

kindermoumoute: Helpers exist here as well, let's refacto this in another PR?

PR #254, opened 2019-09-24
doc: add API regions/zones names

bdronneau:
Hi Guys,

After using scaleway provider with alias for multiple regions I think list of region missing in current documentation: https://www.terraform.io/docs/providers/scaleway/index.html
Since there is a new version of this provider coming, I've added with new region format get from https://github.com/scaleway/scaleway-sdk-go/blob/master/scw/locality.go#L81. Maybe it's not the best place to put this informations, so tell me.

Best


bdronneau:
the rendering is not very good on terraform web page https://www.terraform.io/docs/providers/scaleway/guides/regions_and_zones.html#zones

QuentinBrosse:
Hi @bdronneau, thanks for your feedback, I will fix it in another PR. :)

PR #280, opened 2019-10-17
Update and rename account_ssk_key.html.markdown to account_ssh_key.html

marcw:
This is a quick typo fix for the documentation.


jerome-quere:
Thank you for your PR. Look like their is a broken link, probably in `scaleway.erb`

PR #333, opened 2019-11-21
doc(instance): use image name instead of static image id

remyleone:


website/docs/index.html.markdown
```diff
@@ -63,9 +63,14 @@ resource "scaleway_instance_security_group" "www" {
   }
 }
 
+data "scaleway_image" "ubuntu" {
```
jerome-quere: This data source will soon be deprecated

website/docs/index.html.markdown
```diff
@@ -63,9 +63,14 @@ resource "scaleway_instance_security_group" "www" {
...
+    architecture = "x86_64"
+    name = "Ubuntu Bionic"
+}
+
 resource "scaleway_instance_server" "web" {
   type = "DEV1-L"
-  image = "f974feac-abae-4365-b988-8ec7d1cec10d"
+  image = "${data.scaleway_image.ubuntu.id}"
```
jerome-quere: You can directly write `ubuntu` here

kindermoumoute: `ubuntu-bionic`

remyleone:
Fixed

PR #334, opened 2019-11-21
feat(rdb): add computed attribute and upgrade node type

kindermoumoute:


scaleway/resource_rdb_instance_beta.go
```diff
@@ -67,8 +68,53 @@ func resourceScalewayRdbInstanceBeta() *schema.Resource {
...
+						},
+					},
+				},
+			},
+			"certificate": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "Certificate of the database instance",
+			},
+			"backup_schedule_frequency": {
```
jerome-quere: I don't think this computed is usefull

scaleway/resource_rdb_instance_beta.go
```diff
@@ -140,9 +185,29 @@ func resourceScalewayRdbInstanceBetaRead(d *schema.ResourceData, m interface{})
 	d.Set("user_name", d.Get("user_name").(string)) // user name and
 	d.Set("password", d.Get("password").(string))   // password are immutable
 	d.Set("tags", res.Tags)
+	if res.Endpoint != nil {
```
jerome-quere: If res.Endpoint is null you should set the value to empty string.

website/docs/r/rdb_instance_beta.html.markdown
```diff
@@ -33,7 +33,7 @@ The following arguments are supported:
 
 - `node_type` - (Required) The type of database instance you want to create (e.g. `db-dev-s`).
 
-~> **Important:** Updates to `node_type` will recreate the Database Instance.
+~> **Important:** Updates to `node_type` will upgrade the Database Instance to the desired `node_type` without any interruption.
```
jerome-quere: Add a note that you cannot downgrade and instance

scaleway/resource_rdb_instance_beta.go
kindermoumoute: Should I set `endpoint_port` value to zero as well? 🤔

jerome-quere: yes

PR #336, opened 2019-11-21
chore: remove unnecessary get prefixes from getters

VoyTechnology:
The [Effective Go document](https://golang.org/doc/effective_go.html#Getters) states it is unnecessary to put a Get (or get in this case) prefix before the getter's name.

As Effective Go is the closest we have to an official Go styleguide,
and while officially it makes only "tips on writing clear, idiomatic Go code",
it makes reading Go consistent across projects. The get prefix is neither
idiomatic or needed.

### DIFF
--- a/scaleway/config.go
+++ b/scaleway/config.go
@@ -177,7 +177,7 @@ func (m *Meta) bootstrapS3Client() error {
 	config := &aws.Config{}
 	config.WithRegion(string(m.DefaultRegion))
 	config.WithCredentials(credentials.NewStaticCredentials(m.AccessKey, m.SecretKey, ""))
-	config.WithEndpoint(m.getS3Endpoint(m.DefaultRegion))
+	config.WithEndpoint(m.s3Endpoint(m.DefaultRegion))
 
 	s, err := session.NewSession(config)
 	if err != nil {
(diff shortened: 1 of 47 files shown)


VoyTechnology:
Yes, I understand it is a weird approach to resolving

> Rename the function (for consistency) and add a godoc:
> ```go
> // getContainerRegistryAPIWithRegion returns a new container registry API and the region for a Create request`
> func getContainerRegistryAPIWithRegion(d *schema.ResourceData, m interface{}) (*registry.API, scw.Region, error) {
> ```

_Originally posted by @QuentinBrosse in https://github.com/terraform-providers/terraform-provider-scaleway/pull/330/files_

I do think however it is better to fix the "bad" convention rather than enforcing it. This might be opinionated, but I think following the Effective Go is beneficial for the project. Loads of the standards are thrown out the door by Terraform (using `_` after `resource_{name}.go`, but we should follow the standards everywhere else where possible.

QuentinBrosse:
Hi @VoyTechnology and thanks for your contribution!

I'm not against it, can you just fix the pipeline please:
```
scaleway/resource_baremetal_server_beta.go:74:29: undefined: getBaremetalAPIWithZone
scaleway/resource_baremetal_server_beta.go:136:33: undefined: getBaremetalAPIWithZoneAndID
scaleway/resource_baremetal_server_beta.go:169:33: undefined: getBaremetalAPIWithZoneAndID
scaleway/resource_baremetal_server_beta.go:228:33: undefined: getBaremetalAPIWithZoneAndID
```

Note that I have edited the PR description to include the [Effective Go document link](https://golang.org/doc/effective_go.html#Getters).

QuentinBrosse:
Hi @VoyTechnology and happy new year!

There is no activity in your PR since its creation on Nov 22, 2019. Are you available to finish it?
If you are not, we will close this PR et reopen another one to continue the work.

Thanks

VoyTechnology:
Hi @QuentinBrosse, sorry I was a bit busy recently. I will try and get it finished this week.

VoyTechnology:
@QuentinBrosse PTAL

PR #346, opened 2019-12-02
feat: data source security group

kindermoumoute:


scaleway/data_source_instance_security_group.go
```diff
@@ -0,0 +1,76 @@
...
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/scaleway/scaleway-sdk-go/api/instance/v1"
+)
+
+func dataSourceScalewayInstanceSecurityGroup() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceScalewayInstanceSecurityGroupRead,
+
+		Schema: map[string]*schema.Schema{
```
jerome-quere: Should we read rules ?

kindermoumoute: I saw current security group data source has the 2 following computed: 
```go
			"description": {
				Type:     schema.TypeString,
				Computed: true,
			},
			"enable_default_security": {
				Type:     schema.TypeBool,
				Computed: true,
			},
```
My question is: how much data do we want compute on a data source? 🤔

scaleway/data_source_helpers.go
```diff
@@ -0,0 +1,71 @@
...
+
+import "github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+
+// datasourceSchemaFromResourceSchema is a recursive func that
+// converts an existing Resource schema to a Datasource schema.
+// All schema elements are copied, but certain attributes are ignored or changed:
+// - all attributes have Computed = true
+// - all attributes have ForceNew, Required = false
+// - Validation funcs and attributes (e.g. MaxItems) are not copied
+func datasourceSchemaFromResourceSchema(rs map[string]*schema.Schema) map[string]*schema.Schema {
```
jerome-quere: Should we add source in comment?

kindermoumoute: yes

PR #354, opened 2019-12-04
fix(helpers): remove extra quotes on flatten

Sh4d1:
Signed-off-by: Patrik Cyvoct <pcyvoct@scaleway.com>

PR #358, opened 2019-12-04
fix(instance): allow port-range in security groups

jerome-quere:


Sh4d1:
@jerome-quere #354

jerome-quere:
I add test + split the ip flatten bug in a speprate PR

scaleway/resource_instance_security_group.go
```diff
@@ -368,7 +368,7 @@ func securityGroupRuleSchema() *schema.Resource {
 			},
 			"port_range": {
 				Type:        schema.TypeString,
-				Computed:    true,
+				Optional:    true,
```
kindermoumoute: should this be conflit with port?

jerome-quere: ConflictsWith do no work on neasterd schema

QuentinBrosse: Why not `Optional` and `Computed`?

jerome-quere: It is never computed only the one you use will  be store in state either `port` or `port_range` but never both

PR #362, opened 2019-12-05
feat(marketplace): add image beta data source

kindermoumoute:


scaleway/data_source_marketplace_image_beta.go
```diff
@@ -0,0 +1,50 @@
...
+
+func dataSourceScalewayMarketplaceImageReadBeta(d *schema.ResourceData, m interface{}) error {
+	meta := m.(*Meta)
+	marketplaceAPI := marketplace.NewAPI(meta.scwClient)
+	zone, err := getZone(d, meta)
+	if err != nil {
+		return err
+	}
+
+	imageID, err := marketplaceAPI.GetLocalImageIDByLabel(&marketplace.GetLocalImageIDByLabelRequest{
```
QuentinBrosse: handle error

scaleway/data_source_marketplace_image_beta.go
```diff
@@ -0,0 +1,50 @@
...
+
+		Schema: map[string]*schema.Schema{
+			"label": {
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "Exact label of the desired image",
+			},
+			"instance_type": {
+				Type:        schema.TypeString,
+				Optional:    true,
```
QuentinBrosse: [`GetLocalImageIDByLabel`](https://github.com/terraform-providers/terraform-provider-scaleway/blob/712f0973ac1beafbd0104c11c206855b6f2d928b/vendor/github.com/scaleway/scaleway-sdk-go/api/marketplace/v1/marketplace_utils.go#L56:15) returns the first image that matches. To avoid flaky in unpredictable behaviors, maybe we should make it required?

website/docs/d/marketplace_image_beta.html.markdown
```diff
@@ -0,0 +1,34 @@
+---
+layout: "scaleway"
+page_title: "Scaleway: scaleway_marketplace_image_beta"
+description: |-
+  Gets image ID of an image from its label name.
```
QuentinBrosse: Gets local image ID?

kindermoumoute: for now we haven't introduce the concept of local image in terraform. Do we want that?

scaleway/data_source_marketplace_image_beta.go
kindermoumoute: The default value of `instance_type ` is `DEV1-S` shouldn't cover unpredictable behaviors?

website/docs/d/marketplace_image_beta.html.markdown
QuentinBrosse: It has been already introduced in issues... The idea is just to be clear on what do we exactly expose with this data source on the description.

PR #365, opened 2019-12-09
refactor!: change the way an ip is attached to a server

jerome-quere:


scaleway/resource_instance_server.go
```diff
@@ -373,11 +373,11 @@ func resourceScalewayInstanceServerRead(d *schema.ResourceData, m interface{}) e
 			"host": response.Server.PublicIP.Address.String(),
 		})
 		// Waiting for new breaking change version
```
QuentinBrosse: Remove the comment

website/docs/r/instance_ip.html.markdown
```diff
@@ -20,7 +20,6 @@ resource "scaleway_instance_ip" "server_ip" {}
 The following arguments are supported:
 
 - `reverse` - (Deprecated) Please use the `scaleway_instance_ip_reverse_dns` resource instead.
-- `server_id` - (Deprecated in favor of `[scaleway_instance_server.ip_id]`(instance_server.html.markdown#ip_id)) The ID of the server you want to attach this resource to.
```
QuentinBrosse: Add `deprecated` in the schema.

jerome-quere: I can't because server_id is still a valid attribute but a computed one. I tested with a real tf file. You get this error
```
Error: "server_id": this field cannot be set

  on scaleway.tf line 20, in resource "scaleway_instance_ip" "test":
  20: resource "scaleway_instance_ip" "test" {


```

scaleway/resource_instance_ip.go
```diff
@@ -70,19 +69,6 @@ func resourceScalewayInstanceIPCreate(d *schema.ResourceData, m interface{}) err
 	}
```
kindermoumoute: Remove the reverse block as well

PR #372, opened 2019-12-11
fix(k8s): fix default ingress

Sh4d1:


scaleway/resource_k8s_cluster_beta.go
```diff
@@ -53,7 +53,7 @@ func resourceScalewayK8SClusterBeta() *schema.Resource {
 			"ingress": {
 				Type:        schema.TypeString,
 				Optional:    true,
-				Default:     "no_ingress",
+				Default:     "none",
```
jerome-quere: Please use constant from sdk package. This will avoid similar issue in the future.

Sh4d1: There is none yet :sweat_smile:

PR #375, opened 2019-12-18
fix(k8s): only recreate default pool on its update

### LINKED ISSUE #374: Unable to delete Kapsule default pool
### Terraform Version
```
Terraform v0.12.18
+ provider.scaleway v1.13.0
```

### Affected Resource(s)
- scaleway_k8s_cluster_beta

```hcl
resource "scaleway_k8s_cluster_beta" "cluster" {
    cni           = "cilium"
    name          = "cluster"
    region        = "fr-par"
    tags          = [ "kube"]
    version       = "1.17.0"
}
```

### Debug Output
```
Error: "default_pool": required field is not set

  on scaleway.tf line 3, in resource "scaleway_k8s_cluster_beta" "cluster":
   3: resource "scaleway_k8s_cluster_beta" "cluster" {
```

### Expected Behavior
We should be able to delete the default pool, if we're using another one

### Actual Behavior
We can't delete the default pool, even though we can from the API/web interface

### Steps to Reproduce
1. `terraform apply`

Sh4d1:
Fixes #374 

### DIFF
--- a/scaleway/resource_k8s_cluster_beta.go
+++ b/scaleway/resource_k8s_cluster_beta.go
@@ -5,7 +5,6 @@ import (
 
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 	k8s "github.com/scaleway/scaleway-sdk-go/api/k8s/v1beta3"
-	"github.com/scaleway/scaleway-sdk-go/logger"
 	"github.com/scaleway/scaleway-sdk-go/scw"
 )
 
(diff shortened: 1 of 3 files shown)


scaleway/resource_k8s_cluster_beta.go
```diff
@@ -480,38 +477,49 @@ func resourceScalewayK8SClusterBetaDefaultPoolUpdate(d *schema.ResourceData, m i
...
 			}
-		}
 
-		_, err := k8sAPI.UpdatePool(updateRequest)
-		if err != nil {
-			if !is404Error(err) {
-				return err
+			_, err := k8sAPI.UpdatePool(updateRequest)
+			if err != nil {
+				if !is404Error(err) {
```
kindermoumoute: if this is a 404 something is wrong

kindermoumoute: either force recreate, either error

Sh4d1: Not it's desired behaviour. It was there before. If the default pool is deleted by hand, it should be recreated

PR #390, opened 2020-01-28
chore: linting with default set of golangci-lint rules

alekc:
Just some small tidying after linting with `golangci-lint run` (default rules out of box, version 1.23)

I left alone unused code 
```
scaleway/helpers.go:60:5: var `allStates` is unused (unused)
scaleway/helpers.go:445:6: func `isSDKError` is unused (unused)
scaleway/helpers.go:428:6: func `isResourceTimeoutError` is unused (unused)
scaleway/helpers.go:435:6: func `isSDKResponseError` is unused (unused)
scaleway/helpers_rdb.go:16:6: func `rdbAPI` is unused (unused)
scaleway/resource_server.go:14:5: var `sshHostFingerprints` is unused (unused)
scaleway/data_source_helpers.go:83:6: func `addRequiredFieldsToSchema` is unused (unused)
scaleway/resource_ip_test.go:121:6: func `testAccCheckScalewayIPAttributes` is unused (unused)
```

and for the rest it was simple stuff (errors not checked and had to be explicitely ignored in case of `d.Set`, usage of strings.equalFold, etc.


scaleway/resource_object_bucket.go
```diff
@@ -48,6 +48,9 @@ func resourceScalewayObjectBucketCreate(d *schema.ResourceData, m interface{}) e
 	acl := d.Get("acl").(string)
 
 	s3Client, region, err := s3ClientWithRegion(d, m)
+	if err != nil {
```
alekc: I added a return for this error, if it's something unwanted it has to be changed to `_` (err value is overwritten on the next row).

scaleway/resource_volume.go
```diff
@@ -110,7 +110,7 @@ func resourceScalewayVolumeUpdate(d *schema.ResourceData, m interface{}) error {
 		req.Size = &size
 	}
 
-	scaleway.UpdateVolume(d.Id(), req)
+	_, _ = scaleway.UpdateVolume(d.Id(), req)
```
alekc: this resource is deprecated, so it doesn't matter if we do not check for return

scaleway/resource_instance_server.go
```diff
@@ -421,17 +421,17 @@ func resourceScalewayInstanceServerRead(d *schema.ResourceData, m interface{}) e
...
 		}
 	}
-	d.Set("additional_volume_ids", additionalVolumesIDs)
+	_ = d.Set("additional_volume_ids", additionalVolumesIDs)
 
 	////
 	// Read server user data
 	////
-	allUserData, err := instanceAPI.GetAllServerUserData(&instance.GetAllServerUserDataRequest{
+	allUserData, _ := instanceAPI.GetAllServerUserData(&instance.GetAllServerUserDataRequest{
```
alekc: I suppressed the error here, but it might be a better idea to return an error in case it's not nil)

go.mod
```diff
@@ -2,6 +2,7 @@ module github.com/terraform-providers/terraform-provider-scaleway
 
 require (
 	github.com/aws/aws-sdk-go v1.22.0
+	github.com/google/go-cmp v0.3.0
```
QuentinBrosse: Is it useful?

alekc:
Sure, done. It would be also a good idea to integrate https://golangci.com/. It's free for open source and it's done automatically on every pull.

alekc:
@QuentinBrosse is there anything else you need for this MR?

PR #391, opened 2020-01-29
feat: add scaleway_instance_security_group_rules resource

### LINKED ISSUE #381: Don't remove SecurityGroup Rule flexibilty for v2
Hi Scaleway team,

I am currently using version 1.13 of the scaleway provider for terraform. I see that version 2 is in preparation but when I look at the methods exposed by this one I realize that some essential functionality has the configuration of SecurityGroup for me will not be supported.

I had opened an issue not long ago to ask if you were going to add the possibility of putting several SecurityGroups on a single instance, the idea was to create some by service and to add SecurityGroups on the machines that host Services. It was not in your roadmap I had to find another way to implement this functionality.
Suddenly I went through the `security_group_rule` which offers me the possibility of configuring rules on my securityGroup post creation, when I know the ips of the instances and their services to open or close the routes. My instances `drop` everything by default, I only open the necessary.

The problem is that I see that this method will not be supported for v2, on top of that we cannot add a securityGroup to an instance outside of the instance creation process. This citation makes it impossible for me to properly configure my network with default drop rules.
When I provision my instance, I do not know its ip and if it is linked with another instance I could not add the resulting rules, because the securityGroup is immutable.

Actualy v1.13:
 1. Create all SecurityGroup with empty rules (drop by default)
 2. Create all Instance (link with empty SecurtyGroup)
 3. Configure all rules in all SecurityGroup

v2.0:
 1. Create all SecurityGroup (cannot config with the final private IP)
 2. Create all Instance (link with SecurityGroup)
 3. Cannnot configure rules in SecurityGroup, because `scaleway_security_group_rule` is not availabled

Solution 1: Multi Security Group
 1.0 Create Instance (without securityGroup)
 1.1 Create SecurityGroup
 2.  Add securityGroups to instance (after creation)
Note: I don't need to add rules to the SecurityGroup, if I can put several SecurityGroups on a single instance

Solution 2: Conserve the add rule like actually
 1. Create all SecurityGroup with empty rules (drop by default)
 2. Create all Instance (link with empty SecurtyGroup)
 3. Configure all rules in all SecurityGroup

For the best:
 - Conserve the `scaleway_security_group_rule` feature in the v2 (the name/scope is not important, can change)
 - Implemented add SecurityGroups and can adding SecurityGroup(s) after instance creation

Happy New Year everyone,
Thank you

loicbourgois:
Fixes #381


scaleway/helpers.go
```diff
@@ -216,7 +216,7 @@ func withStoppedServer(scaleway *api.API, serverID string, run func(*api.Server)
 
 // parseLocalizedID parses a localizedID and extracts the resource locality and id.
 func parseLocalizedID(localizedID string) (locality string, ID string, err error) {
-	tab := strings.SplitN(localizedID, "/", 2)
```
kindermoumoute: this should remain 2

scaleway/helpers_instance.go
```diff
@@ -29,11 +29,11 @@ func instanceAPIWithZone(d *schema.ResourceData, m interface{}) (*instance.API,
 }
 
 // instanceAPIWithZoneAndID returns an instance API with zone and ID extracted from the state
-func instanceAPIWithZoneAndID(m interface{}, id string) (*instance.API, scw.Zone, string, error) {
+func instanceAPIWithZoneAndID(m interface{}, zoneID string) (*instance.API, scw.Zone, string, error) {
```
kindermoumoute: zonedID*

scaleway/resource_instance_security_group.go
```diff
@@ -126,16 +137,25 @@ func resourceScalewayInstanceSecurityGroupRead(d *schema.ResourceData, m interfa
 	_ = d.Set("inbound_default_policy", res.SecurityGroup.InboundDefaultPolicy.String())
 	_ = d.Set("outbound_default_policy", res.SecurityGroup.OutboundDefaultPolicy.String())
 
-	//
-	// Handle SecurityGroupRules
-	//
+	if d.Get("external_rules") != nil && d.Get("external_rules").(bool) {
```
kindermoumoute: do you need to check `!= nil` here? 🤔

scaleway/resource_instance_security_group.go
```diff
@@ -126,16 +137,25 @@ func resourceScalewayInstanceSecurityGroupRead(d *schema.ResourceData, m interfa
...
+		stateRules, err := getSecurityGroupRules(instanceApi, zone, ID, d)
```
kindermoumoute: should be `inboundRules, outboundRules, err := getSecurityGroupRules(instanceApi, zone, ID, d)`

scaleway/resource_instance_security_group.go
```diff
@@ -126,16 +137,25 @@ func resourceScalewayInstanceSecurityGroupRead(d *schema.ResourceData, m interfa
...
+	}
+	return nil
+}
+
+func getSecurityGroupRules(instanceApi *instance.API, zone scw.Zone, securityGroupID string, d *schema.ResourceData) (map[instance.SecurityGroupRuleDirection][]interface{}, error) {
 
 	resRules, err := instanceApi.ListSecurityGroupRules(&instance.ListSecurityGroupRulesRequest{
 		Zone:            zone,
-		SecurityGroupID: ID,
+		SecurityGroupID: securityGroupID,
```
kindermoumoute: Here, you should use `expandID(securityGroupID)`

scaleway/resource_instance_security_group.go
```diff
@@ -170,33 +191,41 @@ func resourceScalewayInstanceSecurityGroupRead(d *schema.ResourceData, m interfa
...
-	return nil
+	return stateRules, nil
 }
 
 func resourceScalewayInstanceSecurityGroupUpdate(d *schema.ResourceData, m interface{}) error {
 	meta := m.(*Meta)
 	instanceApi := instance.NewAPI(meta.scwClient)
 
-	zone, ID, err := parseZonedID(d.Id())
+	zone, securityGroupID, err := parseZonedID(d.Id())
```
kindermoumoute: we kept the variable name `ID` across all resources, let's keep consistency here

scaleway/resource_instance_security_group.go
```diff
@@ -170,33 +191,41 @@ func resourceScalewayInstanceSecurityGroupRead(d *schema.ResourceData, m interfa
...
+		SecurityGroupID:       securityGroupID,
+		Description:           scw.StringPtr(description),
 		InboundDefaultPolicy:  &inboundDefaultPolicy,
 		OutboundDefaultPolicy: &outboundDefaultPolicy,
 	}
 
-	// Only update name if on is provided in the state
-	if d.Get("name").(string) != "" {
+	// Only update name if one is provided in the state
+	if d.Get("name") != nil && d.Get("name").(string) != "" {
```
kindermoumoute: why do you check nil here?

scaleway/resource_instance_security_group.go
```diff
@@ -205,18 +234,28 @@ func resourceScalewayInstanceSecurityGroupUpdate(d *schema.ResourceData, m inter
...
-	//
-	// It works as followed:
-	//   1) Creates 2 map[direction][]rule: one for rules in state and one for rules in API
-	//   2) For each direction we:
-	//     A) Loop for each rule in state for this direction
-	//       a) Compare with api rule in this direction at the same index
-	//          if different update / if equals do nothing / if no more api rules to compare create new api rule
-	//     B) If there is more rule in the API we remove them
-	// *******
+	if d.Get("external_rules") != nil && d.Get("external_rules").(bool) {
```
kindermoumoute: why do you check nil?

scaleway/resource_instance_security_group.go
```diff
@@ -205,18 +234,28 @@ func resourceScalewayInstanceSecurityGroupUpdate(d *schema.ResourceData, m inter
...
+		// do nothing
```
kindermoumoute: reverse condition?

scaleway/resource_instance_security_group.go
```diff
@@ -227,9 +266,9 @@ func resourceScalewayInstanceSecurityGroupUpdate(d *schema.ResourceData, m inter
 	}
 
 	// Fill apiRules with data from API
-	resRules, err := instanceApi.ListSecurityGroupRules(&instance.ListSecurityGroupRulesRequest{
+	resRules, err := instanceAPI.ListSecurityGroupRules(&instance.ListSecurityGroupRulesRequest{
 		Zone:            zone,
-		SecurityGroupID: ID,
+		SecurityGroupID: securityGroupID,
```
kindermoumoute: expandID

scaleway/resource_instance_security_group_rules.go
```diff
@@ -0,0 +1,138 @@
...
+	d.SetId(securityGroupRulesIDFromSecurityGroupID(d))
+
+	// We call update instead of read as it will take care of creating rules.
+	return resourceScalewayInstanceSecurityGroupUpdate(d, m)
+}
+
+func resourceScalewayInstanceSecurityGroupRulesRead(d *schema.ResourceData, m interface{}) error {
+
+	securityGroupRulesZoneID := d.Id()
+	securityGroupZoneID := securityGroupZIDFromsecurityGroupRulesZID(securityGroupRulesZoneID)
```
kindermoumoute: keep variable naming consistent with other resources

scaleway/resource_instance_security_group.go
loicbourgois: Yes, I had to add it otherwise there was an error in a test. 

But maybe the right fix is to have a default value for `external_rules`. I don't know how that would work with `ConflictsWith`.

scaleway/helpers.go
loicbourgois: Are we sure ? 

This can lead to errors difficult to diagnose, in case we have something like `zone/zone/ID`.

Do we really want to support `/` in IDs ? Is there a real use case for this ?

scaleway/resource_instance_security_group.go
loicbourgois: ok, but it makes `updateSecurityGroupeRules(d, zone, ID, instanceApi)`, l.238, harder to understand

scaleway/resource_instance_security_group_rules.go
loicbourgois: I find using `ID`when it is in fact a `zonedID`confusing. 

Also, this is a special case where whe have 2 `IDs` because one is built from the other. Just using `ID` would be confusing.

scaleway/resource_instance_security_group.go
loicbourgois: tests were failing: `name` is optionnal

scaleway/resource_instance_security_group.go
loicbourgois: tests were failing: `external_rules` is optionnal

scaleway/resource_instance_security_group.go
```diff
@@ -184,19 +203,29 @@ func resourceScalewayInstanceSecurityGroupUpdate(d *schema.ResourceData, m inter
 		return err
 	}
 
-	inboundDefaultPolicy := instance.SecurityGroupPolicy(d.Get("inbound_default_policy").(string))
-	outboundDefaultPolicy := instance.SecurityGroupPolicy(d.Get("outbound_default_policy").(string))
+	inboundDefaultPolicy := instance.SecurityGroupPolicy("")
+	if d.Get("inbound_default_policy") != nil {
```
jerome-quere: How could this never happen before ? I don't think this could ever happen

scaleway/resource_instance_security_group_rules.go
```diff
@@ -0,0 +1,138 @@
...
+				Description: "Inbound rules for this set of security group rules",
+				Elem:        securityGroupRuleSchema(),
+			},
+			"outbound_rule": {
+				Type:        schema.TypeList,
+				Optional:    true,
+				Description: "Outbound rules for this set of security group rules",
+				Elem:        securityGroupRuleSchema(),
+			},
+			"zone":            zoneSchema(),
```
jerome-quere: Should not be there as the SG is already attached to a zone

scaleway/resource_instance_security_group_rules.go
```diff
@@ -0,0 +1,138 @@
...
+			"organization_id": organizationIDSchema(),
```
jerome-quere: Should not be there as the SG is already attached to an organization

scaleway/resource_instance_security_group_rules.go
```diff
@@ -0,0 +1,138 @@
...
+
+	res, err := instanceApi.GetSecurityGroup(&instance.GetSecurityGroupRequest{
+		SecurityGroupID: securityGroupID,
+		Zone:            zone,
+	})
+	if err != nil {
+		return err
+	}
+
+	d.Set("security_group_id", res.SecurityGroup.ID)
```
jerome-quere: This should be a zoned ID

scaleway/resource_instance_security_group_rules.go
```diff
@@ -0,0 +1,138 @@
...
+	d.Set("zone", zone)
```
jerome-quere: no zone

scaleway/resource_instance_security_group_rules.go
```diff
@@ -0,0 +1,138 @@
...
+	d.Set("organization_id", res.SecurityGroup.Organization)
```
jerome-quere: no organization_id

scaleway/resource_instance_security_group_rules.go
```diff
@@ -0,0 +1,138 @@
...
+func resourceScalewayInstanceSecurityGroupRulesDelete(d *schema.ResourceData, m interface{}) error {
+	meta := m.(*Meta)
+	instanceApi := instance.NewAPI(meta.scwClient)
+
+	zone, securityGroupID, err := parseZonedID(d.Id())
+	if err != nil {
+		return err
+	}
+
+	err = updateSecurityGroupeRules(d, zone, securityGroupID, instanceApi)
```
jerome-quere: need to reset both "in" and "out" boud_rules before calling update

scaleway/resource_instance_security_group_rules.go
```diff
@@ -0,0 +1,138 @@
...
+				Elem:        securityGroupRuleSchema(),
+			},
+			"zone":            zoneSchema(),
+			"organization_id": organizationIDSchema(),
+		},
+	}
+}
+
+func resourceScalewayInstanceSecurityGroupRulesCreate(d *schema.ResourceData, m interface{}) error {
+	d.SetId(securityGroupRulesIDFromSecurityGroupID(d))
```
jerome-quere: d.SetId(securityGroupRulesIDFromSecurityGroupID(d.Get("security_group_id").(string))

scaleway/resource_instance_security_group_rules.go
```diff
@@ -0,0 +1,138 @@
...
+
+	// We call update instead of read as it will take care of creating rules.
+	return resourceScalewayInstanceSecurityGroupUpdate(d, m)
+}
+
+func resourceScalewayInstanceSecurityGroupRulesRead(d *schema.ResourceData, m interface{}) error {
+
+	securityGroupRulesZonedID := d.Id()
+	securityGroupZonedID := securityGroupZIDFromsecurityGroupRulesZID(securityGroupRulesZonedID)
```
jerome-quere: ...From**S**...

scaleway/resource_instance_security_group_rules.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+
+	d.Set("inbound_rule", inboundRules)
+	d.Set("outbound_rule", outboundRules)
+	return nil
+}
+
+func resourceScalewayInstanceSecurityGroupRulesUpdate(d *schema.ResourceData, m interface{}) error {
+
+	instanceApi, zone, securityGroupID, err := instanceAPIWithZoneAndID(m, securityGroupZoneIDFromData(d))
```
jerome-quere: be consistent  with how you handle it in `Read`

scaleway/resource_instance_security_group_rules.go
```diff
@@ -0,0 +1,138 @@
...
+	if err != nil {
+		return err
+	}
+	d.SetId(newZonedId(zone, securityGroupRulesIDFromSecurityGroupID(d)))
```
jerome-quere: Remove this. If SG id change the `ForceNew` will trigger a delete / create

scaleway/resource_instance_security_group_rules.go
```diff
@@ -0,0 +1,138 @@
...
+)
+
+func resourceScalewayInstanceSecurityGroupRules() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceScalewayInstanceSecurityGroupRulesCreate,
+		Read:   resourceScalewayInstanceSecurityGroupRulesRead,
+		Update: resourceScalewayInstanceSecurityGroupRulesUpdate,
+		Delete: resourceScalewayInstanceSecurityGroupRulesDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
```
jerome-quere: A custom importer should:
  - Set the `security_group_id` field
  - Call create

scaleway/resource_instance_security_group_rules.go
```diff
@@ -0,0 +1,138 @@
...
+	if err != nil {
+		return err
+	}
+
+	return resourceScalewayInstanceSecurityGroupRulesRead(d, m)
+}
+
+func resourceScalewayInstanceSecurityGroupRulesDelete(d *schema.ResourceData, m interface{}) error {
+	meta := m.(*Meta)
+	instanceApi := instance.NewAPI(meta.scwClient)
```
jerome-quere: be consistent with Read and update

scaleway/resource_instance_security_group_rules.go
```diff
@@ -0,0 +1,138 @@
...
+// securityGroupRulesIDZFromSecurityGroupIDZ returns the
+//
+// SecurityGroupRules.ID should be the based on SecurityGroup.ID.
+// This is necessary to support Terraform import feature.
+// If we want to support multiple SGs(SecurityGroup) for 1 SGRS(SecurityGroupRules),
+// we could always use the first SecurityGroup's ID,
+// because from the API,
+// the data for a single SGRS is duplicated for all SGs using the same SGRS.
+func securityGroupRulesIDFromSecurityGroupID(d *schema.ResourceData) string {
+	// TODO: have different IDs for SecurityGroup and SecurityGroupRules
```
jerome-quere: This should not be an issue

scaleway/resource_instance_security_group_rules_test.go
```diff
@@ -0,0 +1,95 @@
...
+)
+
+func TestAccScalewayInstanceSecurityGroupRules(t *testing.T) {
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckScalewayInstanceSecurityGroupDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: conf1,
```
jerome-quere: Please inline config

scaleway/resource_instance_security_group_rules_test.go
```diff
@@ -0,0 +1,95 @@
...
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckScalewayInstanceSecurityGroupExists("scaleway_instance_security_group.sg01"),
+				),
+			},
+		},
+	})
+}
+
+var conf1 = `
+resource scaleway_instance_server s01 {
```
jerome-quere: No need to create a server

scaleway/resource_instance_security_group_rules_test.go
```diff
@@ -0,0 +1,95 @@
...
+						IPRange:      expandIPNet("0.0.0.0/0"),
+						DestPortFrom: scw.Uint32Ptr(80),
+						DestPortTo:   nil,
+						Protocol:     instance.SecurityGroupRuleProtocolTCP,
+						Action:       instance.SecurityGroupRuleActionAccept,
+					}),
+				),
+			},
+			{
+				Config: conf2,
```
jerome-quere: inline config

scaleway/resource_instance_security_group_rules_test.go
```diff
@@ -0,0 +1,95 @@
...
+	inbound_rule {
+		action = "accept"
+		port = 80
+		ip_range = "0.0.0.0/0"
+	}
+}
+`
+
+var conf2 = `
+resource scaleway_instance_server s01 {
```
jerome-quere: no need to create a server

scaleway/resource_instance_security_group.go
```diff
@@ -205,18 +234,26 @@ func resourceScalewayInstanceSecurityGroupUpdate(d *schema.ResourceData, m inter
...
-	//
-	// It works as followed:
-	//   1) Creates 2 map[direction][]rule: one for rules in state and one for rules in API
-	//   2) For each direction we:
-	//     A) Loop for each rule in state for this direction
-	//       a) Compare with api rule in this direction at the same index
-	//          if different update / if equals do nothing / if no more api rules to compare create new api rule
-	//     B) If there is more rule in the API we remove them
-	// *******
+	if d.Get("external_rules") == nil || (d.Get("external_rules") != nil && !d.Get("external_rules").(bool)) {
```
jerome-quere: check for nil  ?

scaleway/resource_instance_security_group_rules_test.go
```diff
@@ -0,0 +1,95 @@
...
+						DestPortFrom: scw.Uint32Ptr(80),
+						DestPortTo:   nil,
+						Protocol:     instance.SecurityGroupRuleProtocolTCP,
+						Action:       instance.SecurityGroupRuleActionAccept,
+					}),
+					resource.TestCheckResourceAttrPair("scaleway_instance_security_group.sg01", "id", "scaleway_instance_security_group_rules.sgrs01", "security_group_id"),
+				),
+			},
+			{
+				Config: conf3,
```
jerome-quere: inline config

scaleway/resource_instance_security_group_rules_test.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+resource scaleway_instance_security_group_rules sgrs01 {
+	security_group_id = scaleway_instance_security_group.sg01.id
+	inbound_rule {
+		action = "accept"
+		port = 80
+		ip_range = "0.0.0.0/0"
+	}
+}
```
jerome-quere: Try to add a rule so we can check change are correctly applied

scaleway/resource_instance_security_group.go
kindermoumoute: Exactly, `false` should be the default value and it won't trigger `ConflictsWith `

scaleway/resource_instance_security_group_rules.go
```diff
@@ -35,40 +36,27 @@ func resourceScalewayInstanceSecurityGroupRules() *schema.Resource {
 				Description: "Outbound rules for this set of security group rules",
 				Elem:        securityGroupRuleSchema(),
 			},
-			"zone":            zoneSchema(),
```
kindermoumoute: ok to remove org ID, but zone might be required

Use case:
```hcl
resource scaleway_instance_security_group_rules sgrs01 {
	security_group_id = "123456-789-abc-def" // handwritten UUID
	zone = nl-ams-1 // security group is in a different zone than default zone
} 
```

scaleway/helpers.go
kindermoumoute: we want to support `zone/ID` or `ID` and only these syntax

scaleway/resource_instance_security_group_rules.go
loicbourgois: `security_group_id` is a zonedID, computed from `scaleway_instance_security_group.id()`

scaleway/helpers.go
loicbourgois: Do we really want to allow `/` in ID ? 

Using `tab := strings.SplitN(localizedID, "/", 2)`, `my_zone/my_zone/my_ID` will succeed with `locality=my_zone, ID=my_zone/my_ID`: we want it to fail.

Using `-1`, and with a string containing 2 `/`, the function will fail because `len(tab)==3`.

scaleway/resource_instance_security_group_rules.go
kindermoumoute: There was a misunderstanding about the way to handle child resources. Child resource MUST rely on a parent ID that includes the locality `fr-par-1/xxxx-xx-xx-xx`, which means:
- let's discard this comment
- we must document this clearly (not the scope of this PR)
- we must review all child resources to make sure we do this consistently (not the scope of this PR)

scaleway/resource_instance_security_group_rules.go
```diff
@@ -0,0 +1,130 @@
...
+// securityGroupRulesZIDFromSecurityGroupZID returns the
+//
+// SecurityGroupRules.ID should be the based on SecurityGroup.ID.
+// This is necessary to support Terraform import feature. -> false -> why is this necessary ?
+// If we want to support multiple SGs(SecurityGroup) for 1 SGRS(SecurityGroupRules),
+// we could always use the first SecurityGroup's ID,
+// because from the API,
+// the data for a single SGRS is duplicated for all SGs using the same SGRS.
+func securityGroupRulesZIDFromSecurityGroupZID(SecurityGroupZID string) string {
+	return securityGroupZIDFromSecurityGroupRulesZID(SecurityGroupZID) + "-sgrs-id"
```
kindermoumoute: After few discussions, let's not add suffix for this resource. It'll create confusion for users who will import this resource, or are already familiarised with other resources.

PR #398, opened 2020-02-13
doc(instance): updated instance_security_group with port_range

gaeldb:
Add documentation and example for port_range usage in instance_security_group

### DIFF
--- a/website/docs/r/instance_security_group.html.markdown
+++ b/website/docs/r/instance_security_group.html.markdown
@@ -123,7 +123,7 @@ The `inbound_rule` and `outbound_rule` block supports:
 
 - `protocol`- (Defaults to `TCP`) The protocol this rule apply to. Possible values are: `TCP`, `UDP`, `ICMP` or `ANY`.
 
-- `port`- (Optional) The port this rule applies to. If no `port` nor `port_range` are specified, rule will apply to all port. Only one of `port` and `port_range` should be specified.
+- `port`- (Optional) The port this rule applies to. If no `port` nor `port_range` are specified, the rule will apply to all port. Only one of `port` and `port_range` should be specified.
 
 - `port_range`- Need terraform >= 0.13.0 (Optional) The port range (e.g `22-23`) this rule applies to. Port range MUST comply the Scaleway-notation: interval between ports must be a power of 2 `2^X-1` number (e.g 2^13-1=8191 in port_range = "10000-18191").If no `port` nor `port_range` are specified, rule will apply to all port. Only one of `port` and `port_range` should be specified. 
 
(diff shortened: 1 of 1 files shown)


gaeldb:
Added example and documentation following to https://github.com/terraform-providers/terraform-provider-scaleway/issues/399

website/docs/r/instance_security_group.html.markdown
```diff
@@ -117,7 +123,9 @@ The `inbound_rule` and `outbound_rule` block supports:
 
 - `protocol`- (Defaults to `TCP`) The protocol this rule apply to. Possible values are: `TCP`, `UDP`, `ICMP` or `ANY`.
 
-- `port`- (Optional) The port this rule apply to. If no port is specified, rule will apply to all port.
+- `port`- (Optional) The port this rule applies to. If no `port` nor `port_range` are specified, rule will apply to all port. Only one of `port` and `port_range` should be specified.
+
+- `port_range`- Need terraform >= 0.13.0 (Optional) The port range (e.g `22-23`) this rule applies to. Port range MUST comply the Scaleway-notation: interval between ports must be a power of 2 `2^X-1` number (e.g 2^13-1=8191 in port_range = "10000-18191").If no `port` nor `port_range` are specified, rule will apply to all port. Only one of `port` and `port_range` should be specified.
```
kindermoumoute: Remove `Need terraform >= 0.13.0`. This resource did not exist before v0.13.0 of this plugin

website/docs/r/instance_security_group.html.markdown
```diff
@@ -117,7 +123,9 @@ The `inbound_rule` and `outbound_rule` block supports:
 
 - `protocol`- (Defaults to `TCP`) The protocol this rule apply to. Possible values are: `TCP`, `UDP`, `ICMP` or `ANY`.
 
-- `port`- (Optional) The port this rule apply to. If no port is specified, rule will apply to all port.
+- `port`- (Optional) The port this rule applies to. If no `port` nor `port_range` are specified, rule will apply to all port. Only one of `port` and `port_range` should be specified.
```
remyleone:
```suggestion
- `port`- (Optional) The port this rule applies to. If no `port` nor `port_range` are specified, the rule will apply to all port. Only one of `port` and `port_range` should be specified.
```

website/docs/r/instance_security_group.html.markdown
remyleone:
```suggestion
- `port_range`- Need terraform >= 0.13.0 (Optional) The port range (e.g `22-23`) this rule applies to. Port range MUST comply with the Scaleway-notation: the interval between ports must be a power of 2 `2^X-1` number (e.g 2^13-1=8191 in port_range = "10000-18191"). If no `port` nor `port_range` are specified, the rule will apply to all port. Only one of the `port` and `port_range` option should be specified. 
```

QuentinBrosse: `Only one of `xxx` and `xxxx` option should be specified.` is a standard message that we use for all our oneof.

PR #408, opened 2020-02-21
feat(registry): add namespace data source

Sh4d1:


website/docs/d/registry_namespace_beta.html.markdown
```diff
@@ -0,0 +1,44 @@
+---
```
loicbourgois: add link in  `website/scaleway.erb`

scaleway/data_source_registry_namespace_beta.go
```diff
@@ -0,0 +1,88 @@
...
+			Region:      region,
+			NamespaceID: expandID(namespaceID),
+		})
+		if err != nil {
+			return err
+		}
+		namespace = res
+	}
+
+	d.SetId(datasourceNewRegionalID(namespace.ID, region))
```
loicbourgois: Shouldn't we call `SetId()` only in `create` ?

scaleway/data_source_registry_namespace_beta.go
```diff
@@ -0,0 +1,88 @@
...
+			"namespace_id": {
+				Type:          schema.TypeString,
+				Optional:      true,
+				Description:   "The ID of the Registry Namespace",
+				ConflictsWith: []string{"name"},
+				ValidateFunc:  validationUUIDorUUIDWithLocality(),
+			},
+			"region":          regionSchema(),
+			"organization_id": organizationIDSchema(),
+			"id": {
```
loicbourgois: do we need this ? 

it's missing from `resourceScalewayInstanceSecurityGroup`

scaleway/data_source_registry_namespace_beta.go
```diff
@@ -0,0 +1,88 @@
...
+		if err != nil {
+			return err
+		}
+		namespace = res
+	}
+
+	d.SetId(datasourceNewRegionalID(namespace.ID, region))
+	_ = d.Set("endpoint", namespace.Endpoint)
+	_ = d.Set("is_public", namespace.IsPublic)
+
```
loicbourgois: why not set `name`, `namespace_id`, `region`, `oarganization_id` ?

Sh4d1: `region` and `organization_id` are already set from provider default (or overrides).
`namespace_id` is the same as `id`
`name` could be set.
However since there is the `ConflictsWith` on the schema, I'm not sure about what the behaviour will be like :thinking:

scaleway/data_source_registry_namespace_beta.go
Sh4d1: Yeah I think `id` is set by default (but here we do not inherit from another schema). May need confirmatin though :sweat_smile:

scaleway/data_source_registry_namespace_beta.go
Sh4d1: Other data sources also calls the `SetId` :stuck_out_tongue:

website/docs/d/registry_namespace_beta.html.markdown
Sh4d1: :heavy_check_mark:

scaleway/data_source_registry_namespace_beta.go
loicbourgois: Maybe add a test

Try it without this field

If it fails, try to add back this field.

Sh4d1: Works without :)

scaleway/data_source_registry_namespace_beta.go
Sh4d1: Well tests are passing with name and namespace_id, adding them

scaleway/data_source_registry_namespace_beta.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/scaleway/scaleway-sdk-go/api/registry/v1"
+)
+
+func dataSourceScalewayRegistryNamespace() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceScalewayRegistryNamespaceRead,
+
+		Schema: map[string]*schema.Schema{
```
QuentinBrosse: Can you use the `datasourceSchemaFromResourceSchema` function?
[Example](https://github.com/terraform-providers/terraform-provider-scaleway/blob/master/scaleway/data_source_instance_server.go#L11-L24)

scaleway/data_source_registry_namespace_beta.go
```diff
@@ -0,0 +1,85 @@
...
+		namespace = res
+	}
+
+	d.SetId(datasourceNewRegionalID(namespace.ID, region))
+	_ = d.Set("namespace_id", namespace.ID)
+	_ = d.Set("name", namespace.Name)
+	_ = d.Set("endpoint", namespace.Endpoint)
+	_ = d.Set("is_public", namespace.IsPublic)
+
+	return nil
```
QuentinBrosse: With the `datasourceSchemaFromResourceSchema` method we can simply:

```suggestion
	namespaceID, ok := d.GetOk("namespace_id")
	if !ok {
		res, err := api.ListNamespaces(&registry.ListNamespacesRequest{
			Region: region,
			Name:   String(d.Get("name").(string)),
		})
		if err != nil {
			return err
		}
		if len(res.Namespaces) == 0 {
			return fmt.Errorf("no namespaces found with the name %s", d.Get("name"))
		}
		if len(res.Namespaces) > 1 {
			return fmt.Errorf("%d namespaces found with the same name %s", len(res.Namespaces), d.Get("name"))
		}
		namespaceID = res.Namespaces[0].ID
	}
        
	regionalID := datasourceNewRegionalID(namespace.ID, region)
	d.SetId(regionalID)
	_ = d.Set("namespace_id", regionalID)
	
	return resourceScalewayRegistryNamespaceBetaRead(d, m)
```

scaleway/data_source_registry_namespace_beta_test.go
```diff
@@ -0,0 +1,72 @@
...
+		CheckDestroy: testAccCheckScalewayRegistryNamespaceBetaDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: `
+					resource "scaleway_registry_namespace_beta" "test" {
+						name = "test-cr"
+					}
+					
+					data "scaleway_registry_namespace_beta" "test" {
+						name = "${scaleway_registry_namespace_beta.test.name}"
```
QuentinBrosse:
```suggestion
						name = scaleway_registry_namespace_beta.test.name
```

scaleway/data_source_registry_namespace_beta_test.go
```diff
@@ -0,0 +1,72 @@
...
+					}
+					
+					data "scaleway_registry_namespace_beta" "test2" {
+						namespace_id = "${scaleway_registry_namespace_beta.test.id}"
```
QuentinBrosse:
```suggestion
						namespace_id = scaleway_registry_namespace_beta.test.id
```

scaleway/data_source_registry_namespace_beta_test.go
```diff
@@ -0,0 +1,72 @@
...
+			},
+			{
+				Config: `
+					resource "scaleway_registry_namespace_beta" "test" {
+						name = "test-cr"
+						is_public = true
+					}
+					
+					data "scaleway_registry_namespace_beta" "test" {
+						name = "${scaleway_registry_namespace_beta.test.name}"
```
QuentinBrosse:
```suggestion
						name = scaleway_registry_namespace_beta.test.name
```

scaleway/data_source_registry_namespace_beta_test.go
```diff
@@ -0,0 +1,72 @@
...
+					}
+					
+					data "scaleway_registry_namespace_beta" "test2" {
+						namespace_id = "${scaleway_registry_namespace_beta.test.id}"
```
QuentinBrosse:
```suggestion
						namespace_id = scaleway_registry_namespace_beta.test.id
```

scaleway/data_source_registry_namespace_beta_test.go
```diff
@@ -0,0 +1,72 @@
...
+					}
+				`,
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckScalewayRegistryNamespaceBetaExists("scaleway_registry_namespace_beta.test"),
+					resource.TestCheckResourceAttr("scaleway_registry_namespace_beta.test", "name", "test-cr"),
+					resource.TestCheckResourceAttrSet("data.scaleway_registry_namespace_beta.test", "id"),
+					resource.TestCheckResourceAttr("data.scaleway_registry_namespace_beta.test", "is_public", "false"),
+					resource.TestCheckResourceAttrSet("data.scaleway_registry_namespace_beta.test", "endpoint"),
+					resource.TestCheckResourceAttrSet("data.scaleway_registry_namespace_beta.test", "namespace_id"),
```
QuentinBrosse: For `data.scaleway_registry_namespace_beta.test`, `name` attribute should be set too (same for step 2).

scaleway/data_source_registry_namespace_beta_test.go
```diff
@@ -0,0 +1,72 @@
...
+					resource.TestCheckResourceAttrSet("data.scaleway_registry_namespace_beta.test2", "id"),
+					resource.TestCheckResourceAttr("data.scaleway_registry_namespace_beta.test2", "is_public", "false"),
+					resource.TestCheckResourceAttrSet("data.scaleway_registry_namespace_beta.test2", "endpoint"),
+					resource.TestCheckResourceAttr("data.scaleway_registry_namespace_beta.test2", "name", "test-cr"),
```
QuentinBrosse: For `data.scaleway_registry_namespace_beta.test2`, `namespace_id` attribute should be set too (same for step 2).

PR #409, opened 2020-02-21
feat(registry): add image data source

Sh4d1:
Depends on #408 


loicbourgois:
Can you please add tests for this data source ? 

Especially since there are special cases with `name` and `namespace_id`.

PR #413, opened 2020-02-28
Fix argument disable_dynamic_ip

ChoiZ:
"disable_dynamic_ip": [REMOVED] Removed in favor of enable_dynamic_ip.

### DIFF
--- a/website/docs/r/instance_server.html.markdown
+++ b/website/docs/r/instance_server.html.markdown
@@ -57,7 +57,7 @@ resource "scaleway_instance_server" "web" {
   tags = [ "hello", "public" ]
 
   ip_id = scaleway_instance_ip.ip.id
-  disable_dynamic_ip = true # required for now when using `ip_id`
+  enable_dynamic_ip = false # required for now when using `ip_id`
 }
 ```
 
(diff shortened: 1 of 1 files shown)


website/docs/r/instance_server.html.markdown
```diff
@@ -57,7 +57,7 @@ resource "scaleway_instance_server" "web" {
   tags = [ "hello", "public" ]
 
   ip_id = scaleway_instance_ip.ip.id
-  disable_dynamic_ip = true # required for now when using `ip_id`
```
kindermoumoute: The whole line can be removed as `enable_dynamic_ip = false` is the default value for this attribute

website/docs/r/instance_server.html.markdown
```diff
@@ -57,7 +57,7 @@ resource "scaleway_instance_server" "web" {
...
+  enable_dynamic_ip = false # required for now when using `ip_id`
```
kindermoumoute:
```suggestion
```

website/docs/r/instance_server.html.markdown
ChoiZ: Oh yes of course :)

QuentinBrosse:
Fix #415

PR #422, opened 2020-03-05
fix(k8s): wait for default pool on pool replacement

Sh4d1:

### DIFF
--- a/scaleway/resource_k8s_cluster_beta.go
+++ b/scaleway/resource_k8s_cluster_beta.go
@@ -709,12 +709,19 @@ func resourceScalewayK8SClusterBetaDefaultPoolUpdate(d *schema.ResourceData, m i
 			if err != nil {
 				return err
 			}
+			defaultPoolID = newRegionalId(region, defaultPoolRes.ID)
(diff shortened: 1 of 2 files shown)


scaleway/resource_k8s_cluster_beta.go
```diff
@@ -709,12 +709,19 @@ func resourceScalewayK8SClusterBetaDefaultPoolUpdate(d *schema.ResourceData, m i
...
 
 			_ = d.Set("default_pool", []map[string]interface{}{defaultPool})
 
 			if oldPoolID != "" {
+				// wait for new pool to be ready before deleting old one
+				err = waitK8SPoolReady(k8sAPI, region, expandID(defaultPoolID))
+				if err != nil {
+					return err
+				}
+
```
foux: @Sh4d1 the problem doing it here, is that TF won't return until the new pool is ready. But I guess there's nothing around it :/

Sh4d1: Yep but at least it will crash an error is detected :)

PR #451, opened 2020-04-09
doc: Add a CODEOWNERS file

remyleone:
https://help.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,2 @@
+# Default owner for all pull requests
+*    @terraform-providers/scaleway-provider


.github/CODEOWNERS
```diff
@@ -0,0 +1,2 @@
+# Default owner for all pull requests
+*    @terraform-providers/scaleway-provider
```
jerome-quere, line 2: From what I understant in the doc https://help.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners#codeowners-syntax

You can only add User-Name or Orga Group 
`@terraform-providers/scaleway-provider` where did you get that from ?

remyleone: From here: https://github.com/terraform-providers/terraform-provider-aws/blob/master/.github/CODEOWNERS

jerome-quere: I don't think this apply to us has we are direclty added as developpers on the repo. No github team management in our case

PR #455, opened 2020-04-14
feat(instance): use RegionalID for server

Sh4d1:

### DIFF
--- a/scaleway/helpers.go
+++ b/scaleway/helpers.go
@@ -27,12 +27,29 @@ type RegionalID struct {
 	Region scw.Region
 }
 
-func expandRegionalID(id interface{}) RegionalID {
-	region, ID, _ := parseRegionalID(id.(string))
+func (z RegionalID) String() string {
+	return fmt.Sprintf("%s/%s", z.Region, z.ID)
+}
+
+func newRegionalID(region scw.Region, id string) RegionalID {
 	return RegionalID{
-		ID:     ID,
+		ID:     id,
 		Region: region,
 	}
+}
+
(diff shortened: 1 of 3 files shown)


scaleway/helpers.go
```diff
@@ -21,6 +21,34 @@ import (
...
 )
 
+// RegionalID represents an ID that is linked with a region, eg fr-par/11111111-1111-1111-1111-111111111111
+type RegionalID struct {
+	ID     string
+	Region scw.Region
+}
+
+func expandRegionalID(id interface{}) RegionalID {
+	region, ID, _ := parseRegionalID(id.(string))
```
jerome-quere, line 31: The idea Is to reverse dependence `expandRegionalID` should be the correct function to use.
`parseRegionalID` should rely on `expandRegionalID`

scaleway/resource_instance_server.go
```diff
@@ -231,38 +231,38 @@ func resourceScalewayInstanceServerCreate(d *schema.ResourceData, m interface{})
...
 			CommercialType: commercialType,
 			Zone:           zone,
-			ImageLabel:     image,
+			ImageLabel:     image.ID,
 		})
 		if err != nil {
 			return fmt.Errorf("could not get image '%s': %s", image, err)
 		}
-		image = imageUUID
+		image = expandZonedID(imageUUID)
```
jerome-quere, line 245: expand should be reserved for when you read from `state` . In this case we should have something like
```
image = newZonedId(imageUUID, zone)
```

jerome-quere:
You probably need to update read as well and make sure we are storink localized id in state

PR #502, opened 2020-06-19
chore(vendor): Update scaleway go sdk to latest version

didjcodt:
Signed-off-by: Clément Decoodt <cdecoodt@online.net>

There are some breaking changes since last time, these are fixed in the same commit to avoid a non-compilable commit.


didjcodt:
The CI test seems to be taking the old version while I already updated the go.mod/go.sum with the new version. Do you have any idea why?

jerome-quere:
For now, vendor are tracked. You should run a `go mod vendor` to update vendor folder

didjcodt:
Done, it seems to be working indeed :)

PR #503, opened 2020-06-19
feat(iot): add support for hub resource

didjcodt:
Based on #502 to use the latest Scaleway Go SDK

### DIFF
--- a/scaleway/helpers_iot.go
+++ b/scaleway/helpers_iot.go
@@ -1,0 +1,44 @@
+package scaleway
+
+import (
+	"fmt"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	iot "github.com/scaleway/scaleway-sdk-go/api/iot/v1"
+	"github.com/scaleway/scaleway-sdk-go/scw"
+)
+
+func iotAPIWithRegion(d *schema.ResourceData, m interface{}) (*iot.API, scw.Region, error) {
+	meta := m.(*Meta)
+	iotAPI := iot.NewAPI(meta.scwClient)
+
+	region, err := extractRegion(d, meta)
+
+	return iotAPI, region, err
+}
+
+func iotAPIWithRegionAndID(m interface{}, id string) (*iot.API, scw.Region, string, error) {
+	meta := m.(*Meta)
+	iotAPI := iot.NewAPI(meta.scwClient)
+
+	region, ID, err := parseRegionalID(id)
+	return iotAPI, region, ID, err
+}
+
+func waitIotHub(iotAPI *iot.API, region scw.Region, hubID string, desiredStates ...iot.HubStatus) error {
+	hub, err := iotAPI.WaitForHub(&iot.WaitForHubRequest{
+		HubID:  hubID,
+		Region: region,
+	})
+	if err != nil {
+		return err
+	}
+
+	for _, desiredState := range desiredStates {
+		if hub.Status == desiredState {
+			return nil
+		}
+	}
+
+	return fmt.Errorf("hub %s has state %s, wants one of %+q", hubID, hub.Status, desiredStates)
+}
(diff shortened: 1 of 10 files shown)


scaleway/provider.go
```diff
@@ -192,6 +192,7 @@ func Provider() terraform.ResourceProvider {
 			"scaleway_instance_security_group_rules": resourceScalewayInstanceSecurityGroupRules(),
 			"scaleway_instance_server":               resourceScalewayInstanceServer(),
 			"scaleway_instance_placement_group":      resourceScalewayInstancePlacementGroup(),
+			"scaleway_iot_hub_beta":                  resourceScalewayIotHubBeta(),
```
remyleone, line 195: Should we avoid having beta in the name of the resource?

scaleway/resource_iot_hub_beta.go
```diff
@@ -0,0 +1,249 @@
...
+	err = waitIotHub(iotAPI, region, res.ID, iot.HubStatusReady)
+	if err != nil {
+		return err
+	}
+
+	// Enable/Disable hub if needed. The only case we need to check is to eventually
+	// disable the hub (enabled by default)
+	// We need to ensure the hub was fully enabled before because the hub cannot
+	// be updated while enabling.
+	if enabled, ok := d.GetOk("enabled"); ok && !enabled.(bool) {
```
jerome-quere, line 111: To avoid code duplication you could consider. Not sure about the `HasChange` method comportement thought
```
d.SetId(newRegionalId(region, res.ID))
resourceScalewayIotHubBetaRead(d, m)
return resourceScalewayIotHubBetaUpdate(d, m)
```

scaleway/provider.go
didjcodt: I have done this to stick to the convention with k8s and lb, should the naming be related to the API or the state of the product?

scaleway/resource_iot_hub_beta.go
```diff
@@ -0,0 +1,249 @@
...
+			err = waitIotHub(iotAPI, region, hubID, iot.HubStatusReady)
+			if err != nil {
+				return err
+			}
+		} else {
+			_, err = iotAPI.DisableHub(&iot.DisableHubRequest{
+				Region: region,
+				HubID:  hubID,
+			})
+			if err != nil {
```
jerome-quere, line 217: You don't wait here ?

scaleway/provider.go
jerome-quere: We add an internal debate about this as we foud out that migration from `beta` resource to non-beta is a HUGE pain for users. We would prefer and mandatory env variable to activate beta feature but it would not be coherent with other resources.

didjcodt: I can do as you prefer ;)

scaleway/resource_iot_hub_beta.go
didjcodt: Calling the update method might probably trigger an enable on an already enabled hub

scaleway/resource_iot_hub_beta.go
didjcodt: There can be an specific path (even if it would be hard to produce in terraform) where this could lead to a "Resource is busy", indeed. I changed this to wait in every case. This was not done before because the devices are disconnected right during the call :)

scaleway/provider.go
didjcodt: Changed it to remove the beta for now, if you want to keep it just ask ;)

scaleway/helpers_iot.go
```diff
@@ -0,0 +1,50 @@
...
+	"fmt"
+	"time"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	iot "github.com/scaleway/scaleway-sdk-go/api/iot/v1beta1"
+	"github.com/scaleway/scaleway-sdk-go/scw"
+)
+
+const (
+	IotHubWaitTimeout = 10 * time.Minute
```
remyleone, line 13: Should be in the SDK https://github.com/scaleway/scaleway-sdk-go/blob/master/api/lb/v1/lb_utils.go#L11

scaleway/provider.go
```diff
@@ -199,6 +199,7 @@ func Provider() func() *schema.Provider {
 				"scaleway_instance_security_group_rules": resourceScalewayInstanceSecurityGroupRules(),
 				"scaleway_instance_server":               resourceScalewayInstanceServer(),
 				"scaleway_instance_placement_group":      resourceScalewayInstancePlacementGroup(),
+				"scaleway_iot_hub":                       resourceScalewayIotHub(),
```
remyleone, line 202: Should be consistant with scaleway_iot_hub_beta

remyleone (approved):

PR #520, opened 2020-07-09
feat(rdb): add scaleway_rdb_user_resource_beta

paulez:
This is to add a resource to manage RDB database users.

This allows to manage users admin permissions and passwords on RDB instances.

I've run the acceptance tests I've added:

```
17:12 paul@sca1 ~/github/terraform-provider-scaleway/scaleway% time TF_ACC=1 go test -v . -run "TestAccScalewayRdbUser*" -timeout=10m -count=1
=== RUN   TestAccScalewayRdbUserBeta
=== PAUSE TestAccScalewayRdbUserBeta
=== CONT  TestAccScalewayRdbUserBeta
--- PASS: TestAccScalewayRdbUserBeta (214.60s)
PASS
ok  	github.com/terraform-providers/terraform-provider-scaleway/scaleway	214.609s
TF_ACC=1 go test -v . -run "TestAccScalewayRdbUser*" -timeout=10m -count=1  6,05s user 0,81s system 3% cpu 3:37,17 total
```

Next step will be to allow creating instances without setting a user and password, so users can be fully managed using this type of resource.

### DIFF
--- a/scaleway/resource_rdb_user_beta.go
+++ b/scaleway/resource_rdb_user_beta.go
@@ -88,7 +88,7 @@ func resourceScalewayRdbUserBetaRead(d *schema.ResourceData, m interface{}) erro
 
 	res, err := rdbAPI.ListUsers(&rdb.ListUsersRequest{
 		Region:     region,
-		InstanceID: expandID(instanceId),
+		InstanceID: instanceId,
 		Name:       &userName,
 	})
 
@@ -97,6 +97,7 @@ func resourceScalewayRdbUserBetaRead(d *schema.ResourceData, m interface{}) erro
 	}
 
 	var user = res.Users[0]
+	_ = d.Set("instance_id", fmt.Sprintf("%s/%s", region, instanceId))
 	_ = d.Set("name", user.Name)
 	_ = d.Set("password", d.Get("password").(string)) // password are immutable
 	_ = d.Set("is_admin", user.IsAdmin)
--- a/scaleway/resource_rdb_user_beta.go
+++ b/scaleway/resource_rdb_user_beta.go
@@ -55,10 +55,10 @@ func resourceScalewayRdbUserBetaCreate(d *schema.ResourceData, m interface{}) er
 	if err != nil {
 		return err
 	}
-	instanceId := d.Get("instance_id").(string)
+	instanceID := d.Get("instance_id").(string)
 	createReq := &rdb.CreateUserRequest{
 		Region:     region,
-		InstanceID: expandID(instanceId),
+		InstanceID: expandID(instanceID),
 		Name:       d.Get("name").(string),
 		Password:   d.Get("password").(string),
 		IsAdmin:    d.Get("is_admin").(bool),
@@ -69,7 +69,7 @@ func resourceScalewayRdbUserBetaCreate(d *schema.ResourceData, m interface{}) er
 		return err
 	}
 
-	d.SetId(resourceScalewayRdbUserBetaID(region, expandID(instanceId), res.Name))
+	d.SetId(resourceScalewayRdbUserBetaID(region, expandID(instanceID), res.Name))
 
 	return resourceScalewayRdbUserBetaRead(d, m)
 }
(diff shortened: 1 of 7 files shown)


kindermoumoute (changes requested):

scaleway/resource_rdb_user_beta.go
```diff
@@ -0,0 +1,176 @@
...
+	})
+
+	if err != nil {
+		return err
+	}
+
+	var user = res.Users[0]
+	_ = d.Set("instance_id", fmt.Sprintf("%s/%s", region, instanceId))
+	_ = d.Set("name", user.Name)
+	_ = d.Set("password", d.Get("password").(string)) // password are immutable
```
remyleone, line 102: Does that mean that password are stored in plain test ?

remyleone: Would it be better to hash this to avoid storing sensitive information in plain text? https://registry.terraform.io/providers/hashicorp/random/latest/docs/resources/password

scaleway/resource_rdb_user_beta.go
```diff
@@ -0,0 +1,176 @@
...
+	_ = d.Set("is_admin", user.IsAdmin)
+
+	d.SetId(fmt.Sprintf("%s/%s/%s", regionName, instanceId, user.Name))
```
remyleone, line 105: Could we have an helper for this instead of having several different sprintf that might get out of sync

website/docs/r/rdb_user_beta.html.markdown
```diff
@@ -0,0 +1,48 @@
...
+
+```hcl
+resource scaleway_rdb_user_beta db_admin {
+  	instance_id = scaleway_rdb_instance_beta.main.id
+  	name = "titi"
+  	password = "R34lP4sSw#Rd"
+  	is_admin = true
+}
+```
+
```
kindermoumoute, line 26: can you add an example with the[ provider password](https://registry.terraform.io/providers/hashicorp/random/latest/docs/resources/password)?

website/docs/r/rdb_user_beta.html.markdown
```diff
@@ -0,0 +1,48 @@
+---
```
remyleone, line 1: Could you add this file in https://github.com/scaleway/terraform-provider-scaleway/blob/master/website/scaleway.erb

scaleway/resource_rdb_user_beta.go
```diff
@@ -0,0 +1,176 @@
...
+	rdbAPI, region, err := rdbAPIWithRegion(d, m)
+	if err != nil {
+		return err
+	}
+
+	regionName, instanceId, userName, err := resourceScalewayRdbUserBetaParseId(d.Id())
+
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
paulez: `rdbAPIWithRegionAndID` doesn't support more than one / in the ID. Here I am storing Region/Instance/Username but this is not supported by the helpers it seems. Do you think I should modify the helpers to support this? This could have some side effects.

scaleway/resource_rdb_user_beta.go
paulez: I am following recommendations from https://www.terraform.io/docs/state/sensitive-data.html
The password is stored as is in local state. However this is the standard behaviour for providers, user are encouraged to use an encrypted state store when using secrets.

website/docs/r/rdb_user_beta.html.markdown
```diff
@@ -0,0 +1,53 @@
...
+
+~> **Important:** Updates to `instance_id` will recreate the Database User.
+
+- `name` - (Required) Database User name.
+
+~> **Important:** Updates to `name` will recreate the Database User.
+
+- `password` - (Required) Database User password.
+
+- `is_admin` - (Optional) Grand admin permissions to the Database User.
```
remyleone, line 46:
```suggestion
- `is_admin` - (Optional) Grant admin permissions to the Database User.
```

paulez: Oh that's grand!

scaleway/resource_rdb_user_beta.go
```diff
@@ -0,0 +1,185 @@
...
+	if err != nil {
+		if is404Error(err) {
+			d.SetId("")
+			return nil
+		}
+		return err
+	}
+
+	var user = res.Users[0]
+	d.Set("instance_id", fmt.Sprintf("%s/%s", region, instanceID))
```
kindermoumoute, line 104:
```suggestion
	d.Set("instance_id", newRegionalID(region, instanceID))
```

kindermoumoute (approved):
LGTM

PR #530, opened 2020-08-20
chore(release): use goreleaser

kindermoumoute:
Notes:
- Secrets `GPG_PRIVATE_KEY`, `PASSPHRASE` and `GITHUB_TOKEN` must be set
- Changelog does not use conventional commit (this is not a feature of goreleaser)
- Version is set through a ldflag `github.com/scaleway/terraform-provider-scaleway/scaleway.version={{.Version}}`, either we choose not to commit version anymore, either we add some kind of post-release workflow that open a PR


remyleone:
Should we use a script similar to the one in https://github.com/scaleway/scaleway-cli to have the environment variables set the credentials available?

kindermoumoute:
> Should we use a script similar to the one in https://github.com/scaleway/scaleway-cli to have the environment variables set the credentials available?

There are too many difference between a CLI release and a terraform plugin release to use the same script ultimately:
- Terraform GPG keys process
- CLI release container, distribution-specific package, ...

remyleone (approved):

PR #547, opened 2020-09-16
chore: add support for golangci-lint

remyleone:


scaleway/resource_object_bucket.go
```diff
@@ -162,6 +162,10 @@ func resourceScalewayObjectBucketUpdate(d *schema.ResourceData, m interface{}) e
 				TagSet: tagsSet,
 			},
 		})
+		if err != nil {
+			l.Errorf("Couldn't update bucket tags: %s", err)
```
jerome-quere, line 166: returning the error should be enough as it will be logged upper in the stack.

jerome-quere (changes requested):

kindermoumoute (approved):
LGTM.

jerome-quere (approved):

PR #566, opened 2020-09-22
Increase resource size limit (volume+server)

sefam-mchatelain:
from 150 to 250, because it is supported by the API, and `fr-sefam-devops` has a disk of 200Gb.

### DIFF
--- a/scaleway/resource_server.go
+++ b/scaleway/resource_server.go
@@ -83,8 +83,8 @@ func resourceScalewayServer() *schema.Resource {
 							Required: true,
 							ValidateFunc: func(v interface{}, k string) (ws []string, errors []error) {
 								value := v.(int)
-								if value > 150 {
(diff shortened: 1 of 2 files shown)


Vespira (commented):
I saw this into the file you edited

> DeprecationMessage: `This resource is deprecated and will be removed in the next major version.
>  Please use scaleway_instance_server instead.`

Is this something we should care about or is it not important from now ?

Sh4d1 (dismissed):
The `scaleway_server` is indeed a deprecated resource. You should use `scaleway_instance_server` instead. Though I see no problem in the MR anyway. @jerome-quere WDYT?

Sh4d1:
@sefam-mchatelain the old resources were deleted in https://github.com/scaleway/terraform-provider-scaleway/commit/231fc91aa8ec8ae7ba6a12f4372cd36dc51d622b so I'll close your PR since it won't be possible to merge it. I recommend to use the new `scaleway_instance)server`.

PR #569, opened 2020-09-23
feat(rdb): add support for data source rdb instance

remyleone:


scaleway/provider.go
```diff
@@ -241,6 +241,7 @@ func Provider() terraform.ResourceProvider {
 			"scaleway_instance_volume":         dataSourceScalewayInstanceVolume(),
 			"scaleway_baremetal_offer":         dataSourceScalewayBaremetalOffer(),
 			"scaleway_lb_ip_beta":              dataSourceScalewayLbIPBeta(),
+			"scaleway_rdb_instance_beta":       dataSourceScalewayRDBInstanceBeta(),
```
kindermoumoute, line 244: should we start removing the `beta`?

PR #570, opened 2020-09-23
refactor: provider configuration

jerome-quere:


scaleway/resource_lb_backend_beta_test.go
```diff
@@ -165,6 +168,7 @@ func TestAccScalewayLbBackendBeta_HealthCheck(t *testing.T) {
 						name = "bkd01"
 						forward_protocol = "tcp"
 						forward_port = 80
+						proxy_protocol = "none"
```
jerome-quere, line 171: Should not be required. Maybe default value in ressource is not correct

PR #595, opened 2020-10-06
chore: use expandStringPtr instead of scw.StringPtr

remyleone:


scaleway/data_source_account_ssh_key.go
```diff
@@ -49,7 +48,7 @@ func dataSourceScalewayAccountSSHKeyRead(d *schema.ResourceData, m interface{})
 		sshKey = res
 	} else {
 		res, err := accountAPI.ListSSHKeys(&account.ListSSHKeysRequest{
-			Name: scw.StringPtr(d.Get("name").(string)),
+			Name: expandStringPtr(d.Get("name").(string)),
```
jerome-quere, line 51: I think the `.(string)` cast is unnecessary

jerome-quere (approved):

PR #603, opened 2020-10-06
feat(account): add support for project in SSH key

remyleone:


QuentinBrosse (changes requested):
We should add a test for the `project_id` field.

Sh4d1 (dismissed):
LGTM

QuentinBrosse (approved):
Sorry for that, we don't have acc tests with a custom organization, so no new tests are needed.

PR #617, opened 2020-10-09
test: add cassette system

jerome-quere:


scaleway/provider_test.go
```diff
@@ -1,16 +1,34 @@
...
-var testAccProviders map[string]*schema.Provider
-var testAccProvider *schema.Provider
+var (
+	// Deprecated
+	testAccProviders map[string]*schema.Provider
+	// Deprecated
+	testAccProvider *schema.Provider
+
+	// UpdateCassettes will update all cassettes of a given test
+	UpdateCassettes = flag.Bool("cassettes", os.Getenv("CLI_UPDATE_CASSETTES") == "true", "Record Cassettes")
```
remyleone, line 27: `TF_UPDATE_CASSETTES` ?

scaleway/provider_test.go
```diff
@@ -19,3 +37,80 @@ func init() {
...
+
+	// Replace special characters.
+	fileName = specialChars.ReplaceAllLiteralString(fileName, "") + suffix
+
+	return filepath.Join(".", "testdata", fileName)
+}
+
+// getHTTPRecoder creates a new httpClient that records all HTTP requests in a cassette.
+// This cassette is then replayed whenever tests are executed again. This means that once the
+// requests are recorded in the cassette, no more real HTTP request must be made to run the tests.
```
remyleone, line 58: ``
```suggestion
// requests are recorded in the cassette, no more real HTTP requests must be made to run the tests.
```

scaleway/provider_test.go
```diff
@@ -19,3 +37,80 @@ func init() {
...
+//
+// It is important to call add a `defer cleanup()` so the given cassette files are correctly
```
remyleone, line 60:
```suggestion
// It is important to add a `defer cleanup()` so the given cassette files are correctly
```

scaleway/version.go
```diff
@@ -1,4 +1,4 @@
 package scaleway
 
-// version not a const because we modify it in provider_test.go
-var version = "v1.16.0"
+// This variable is set by go releaser on release
+var version = "develop"
```
remyleone, line 4: Does the releaser script need a fix? Should we create a ticket for it?

scaleway/provider_test.go
```diff
@@ -19,3 +37,80 @@ func init() {
 }
 
 func testAccPreCheck(_ *testing.T) {}
+
+// getTestFilePath returns a valid filename path based on the go test name and suffix. (Take care of non fs friendly char)
+func getTestFilePath(t *testing.T, suffix string) string {
```
remyleone, line 42: All the testdata golden and cassettes will ALL be in the same folder. Plus all the file name will start with `test-acc-scaleway` that will not bring any information and might hurt the visibility in IDE pans.
Should we rename this pattern to something more practical? I think we can also rename the convention of TestAccScaleway because I don't see the point of it.

scaleway/provider_test.go
remyleone, line 27:
```suggestion
	UpdateCassettes = flag.Bool("cassettes", os.Getenv("TF_UPDATE_CASSETTES") == "true", "Record Cassettes")
```

remyleone (approved):

PR #645, opened 2020-10-16
chore(instance): add support for cassettes based testing

remyleone:


scaleway/data_source_instance_volume_test.go
```diff
@@ -1,21 +1,33 @@
...
+	"fmt"
 	"testing"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
 )
 
 func TestAccScalewayDataSourceInstanceVolume_Basic(t *testing.T) {
+	tt := NewTestTools(t)
+	defer tt.Cleanup()
+	volumeName := newRandomName("volume")
```
jerome-quere, line 13: This type of randomness could cause problems with cassette. We should consider the generation a name based on the test name. Or remove uniqness if not needed

scaleway/resource_instance_server.go
```diff
@@ -381,7 +381,7 @@ func resourceScalewayInstanceServerRead(ctx context.Context, d *schema.ResourceD
 	}
 
 	if response.Server.PrivateIP != nil {
```
jerome-quere, line 383: We should use the `flatenStrinPtr` function for this use case

jerome-quere (changes requested):

jerome-quere (approved):

PR #646, opened 2020-10-16
chore(baremetal): add cassettes testing

remyleone:


scaleway/resource_baremetal_server_test.go
```diff
@@ -40,25 +40,28 @@ func testSweepBaremetalServer(_ string) error {
 	})
 }
 
-func TestAccScalewayBaremetalServerMinimal1(t *testing.T) {
+func TestAccScalewayBaremetalServer_Basic(t *testing.T) {
+	tt := NewTestTools(t)
+	defer tt.Cleanup()
+
 	SSHKeyName := newRandomName("ssh-key")
```
jerome-quere, line 47: We should be careful with randomness and cassettes

jerome-quere (changes requested):

jerome-quere (approved):

PR #652, opened 2020-10-21
feat(instance): add support for bootscript

remyleone:


scaleway/resource_instance_server.go
```diff
@@ -174,9 +174,14 @@ func resourceScalewayInstanceServer() *schema.Resource {
 			},
 			"boot_type": {
 				Type:        schema.TypeString,
-				Computed:    true,
+				Optional:    true,
 				Description: "The boot type of the server",
 			},
+			"bootscript": {
```
kindermoumoute, line 180: bootscript_id

scaleway/resource_instance_server.go
```diff
@@ -174,9 +174,14 @@ func resourceScalewayInstanceServer() *schema.Resource {
 			},
 			"boot_type": {
 				Type:        schema.TypeString,
-				Computed:    true,
+				Optional:    true,
 				Description: "The boot type of the server",
```
kindermoumoute, line 178: add default value

scaleway/resource_instance_server.go
```diff
@@ -259,6 +264,12 @@ func resourceScalewayInstanceServerCreate(ctx context.Context, d *schema.Resourc
 		Tags:              expandStrings(d.Get("tags")),
 	}
 
+	if bootType, ok := d.GetOk("boot_type"); ok && bootType == "bootscript" {
```
kindermoumoute, line 267: [use bootType const](https://github.com/scaleway/scaleway-sdk-go/blob/3c87d958dd1172db3151d8fdbbb51eb309b1bd32/api/instance/v1/instance_sdk.go#L92)

kindermoumoute (changes requested):

scaleway/resource_instance_server.go
Sh4d1: maybe add the allowed values too ?

Sh4d1:
What happens when a user specifies a bootscript without changing the boot type ? We could either throw an error, or change the boot type.

docs/resources/instance_server.md
```diff
@@ -177,7 +177,7 @@ attached to the server. Updates to this field will trigger a stop/start of the s
 
 - `boot_type` - The boot Type of the server. Possible values are: `local`, `bootscript` or `rescue`.
 
-- `bootscript` - The ID of the bootscript to use
+- `bootscript` - The ID of the bootscript to use  (set boot_type to `bootscript`).
```
kindermoumoute, line 180: bootscript_id

kindermoumoute (changes requested):

Sh4d1 (changes requested):
Need to set the bootype in create req event when no bootscript

Some questions:
- what about the update?
- what if the user sets the boot_type to local and a bootscript ID ? The states will mismatch, I think throwing an error would be better

NoSuchCommand:
> Need to set the bootype in create req event when no bootscript
> 
> Some questions:
> 
>     * what about the update?
> 
>     * what if the user sets the boot_type to local and a bootscript ID ? The states will mismatch, I think throwing an error would be better

scaleway/resource_instance_server.go
```diff
@@ -259,6 +271,12 @@ func resourceScalewayInstanceServerCreate(ctx context.Context, d *schema.Resourc
 		Tags:              expandStrings(d.Get("tags")),
 	}
 
+	if bootScriptID, ok := d.GetOk("bootscript_id"); ok {
+		b := instance.BootTypeBootscript
```
kindermoumoute, line 275: IIRC a bootscript can be run in rescue mode too, that's why you should read `boot_type` attribute too

kindermoumoute (changes requested):

scaleway/resource_instance_server.go
```diff
@@ -272,11 +272,16 @@ func resourceScalewayInstanceServerCreate(ctx context.Context, d *schema.Resourc
 	}
 
 	if bootScriptID, ok := d.GetOk("bootscript_id"); ok {
-		b := instance.BootTypeBootscript
-		req.BootType = &b
 		req.Bootscript = expandStringPtr(bootScriptID)
 	}
 
+	if bootType, ok := d.GetOk("boot_type"); ok {
+		bootType, ok := bootType.(instance.BootType)
```
kindermoumoute, line 279: I think you mean `bootType := instance.BootType(bootType.(string))`

kindermoumoute: https://play.golang.org/p/80ZOT8_UdkG

kindermoumoute (dismissed):
LGTM

scaleway/resource_instance_server.go
```diff
@@ -259,6 +271,15 @@ func resourceScalewayInstanceServerCreate(ctx context.Context, d *schema.Resourc
 		Tags:              expandStrings(d.Get("tags")),
 	}
 
+	if bootScriptID, ok := d.GetOk("bootscript_id"); ok {
+		req.Bootscript = expandStringPtr(bootScriptID)
+	}
+
+	if bootType, ok := d.GetOk("boot_type"); ok {
+		bootType := instance.BootType(bootType.(string))
```
Sh4d1, line 279: remove this line

Sh4d1 (changes requested):

Sh4d1 (approved):
LGTM

PR #653, opened 2020-10-21
feat(object): add support for exporting endpoint url

remyleone:


kindermoumoute (approved):
LGTM

scaleway/resource_object_bucket.go
```diff
@@ -131,6 +136,8 @@ func resourceScalewayObjectBucketRead(d *schema.ResourceData, m interface{}) err
 
 	_ = d.Set("tags", flattenObjectBucketTags(tagsSet))
 
+	_ = d.Set("endpoint", fmt.Sprintf("https://%s.s3.%s.scw.cloud", bucketName, region))
```
jerome-quere, line 139: I think this string is already present somewhere else. Maybe it should be an helper function

remyleone: https://github.com/scaleway/terraform-provider-scaleway/pull/658

PR #658, opened 2020-10-23
chore(object): create an helper function for endpoint url

remyleone:


scaleway/helpers_object.go
```diff
@@ -92,3 +94,7 @@ func expandObjectBucketTags(tags interface{}) []*s3.Tag {
 
 	return tagsSet
 }
+
+func endpointURL(bucketName string, region scw.Region) string {
```
QuentinBrosse, line 98: Maybe rename it `objectEndpointURL` as it is shared in the whole `scaleway` package?

QuentinBrosse (changes requested):

QuentinBrosse (approved):
LGTM

PR #666, opened 2020-10-30
feat(instance): add support for placement_group_id in Read function

remyleone:


scaleway/resource_instance_server.go
```diff
@@ -399,6 +399,7 @@ func resourceScalewayInstanceServerRead(ctx context.Context, d *schema.ResourceD
 	}
 
 	if response.Server.PlacementGroup != nil {
+		_ = d.Set("placement_group_id", response.Server.PlacementGroup.ID)
```
jerome-quere, line 402: You should store a localized id

jerome-quere (changes requested):

jerome-quere (approved):

PR #672, opened 2020-11-02
fix(instance): ensure that tags are removed when deleted in tf

remyleone:


remyleone:
https://play.golang.org/p/uA8bWCNiTpR

scaleway/resource_instance_server.go
```diff
@@ -508,6 +508,9 @@ func resourceScalewayInstanceServerUpdate(ctx context.Context, d *schema.Resourc
 
 	if d.HasChange("tags") {
 		tags := expandStrings(d.Get("tags"))
+		if tags == nil {
```
jerome-quere, line 511: func expandStrings(data interface{}) []string {
	if data == nil {
		return []string{}
	}
	stringSlice := []string{}
	for _, s := range data.([]interface{}) {
		stringSlice = append(stringSlice, s.(string))
	}
	return stringSlice
}

jerome-quere (changes requested):

jerome-quere (approved):

PR #680, opened 2020-11-04
chore(account): refactor data ressource workflow

remyleone:


scaleway/data_source_account_ssh_key.go
```diff
@@ -11,35 +11,26 @@ import (
...
 
 func dataSourceScalewayAccountSSHKey() *schema.Resource {
+	dsSchema := datasourceSchemaFromResourceSchema(resourceScalewayAccountSSKKey().Schema)
+	addOptionalFieldsToSchema(dsSchema, "name")
+
+	dsSchema["name"].ConflictsWith = []string{"ssh_key_id"}
+	dsSchema["ssh_key_id"] = &schema.Schema{
+		Type:         schema.TypeString,
+		Optional:     true,
+		Computed:     true,
```
jerome-quere, line 21: Should not be marked as computed

scaleway/data_source_account_ssh_key.go
```diff
@@ -11,35 +11,26 @@ import (
...
+		Description:  "The ID of the SSH key",
+		ValidateFunc: validationUUIDorUUIDWithLocality(),
```
jerome-quere, line 23: ssh-key are never localized

jerome-quere (changes requested):
view comments

remyleone:
@jerome-quere do you see other resources that should be normalized?

jerome-quere (approved):

PR #681, opened 2020-11-04
fix(instance): handle root volume with no size as default

remyleone:


jerome-quere (commented):
You can have a look at the create server in the CLI IIRC we have the same kind of fix

scaleway/helpers_instance.go
```diff
@@ -159,3 +160,72 @@ func reachState(ctx context.Context, instanceAPI *instance.API, zone scw.Zone, s
...
+
+		max := humanize.Bytes(uint64(volumeConstraint.MaxSize))
+		return fmt.Errorf("%s total local volume size must be between %s and %s", commercialType, min, max)
+	}
+
+	return nil
+}
+
+// sanitizeVolumeMap removes extra data for API validation.
+func sanitizeVolumeMap(serverName string, volumes map[string]*instance.VolumeTemplate) map[string]*instance.VolumeTemplate {
```
jerome-quere, line 214: Why do we need this ? 
Can we construct the struct correclty in the first place ?

scaleway/resource_instance_server.go
```diff
@@ -284,16 +284,43 @@ func resourceScalewayInstanceServerCreate(ctx context.Context, d *schema.Resourc
...
+	if serverType != nil {
+		// We had a local root volume if it is not already present
+		if rootVolume := req.Volumes["0"]; rootVolume == nil {
+			req.Volumes["0"] = &instance.VolumeTemplate{
+				Name:       newRandomName("vol"),
+				VolumeType: instance.VolumeVolumeTypeLSSD,
+				Size:       serverType.VolumesConstraint.MinSize,
+			}
+		}
+	}
```
jerome-quere, line 297: Should this be in an else of the if just above ?

scaleway/resource_instance_server_test.go
```diff
@@ -863,3 +863,35 @@ func TestAccScalewayInstanceServer_AlterTags(t *testing.T) {
...
+		Steps: []resource.TestStep{
+			{
+				Config: `
+					resource "scaleway_instance_volume" "data" {
+						size_in_gb = 100
+						type = "b_ssd"
+					}
+
+					resource "scaleway_instance_server" "main" {
+						type = "DEV1-L"
```
jerome-quere, line 883: DEV1-L Are longer to shutdown for test maybe prefer a DEV1-S

jerome-quere (changes requested):

scaleway/helpers_instance.go
remyleone: I added documentation to this function. I want to
1. Remain consistent with what is done in the CLI 
2. It allows us to manage multiple cases of entries and multiple validations in a single place. At least the advantage is that if their constraints change we have a specific place or apply the cleaning functions that depend on several different constraints (image specified or not, volume already existing or not).

scaleway/resource_instance_server.go
remyleone: `serverType` should be defined outside anyway because of the `validateLocalVolumeSizes` that come later that requires it.

jerome-quere (approved):

PR #691, opened 2020-11-12
fix(baremetal): fix baremetal test

remyleone:
Tries to fix this: https://github.com/scaleway/terraform-provider-scaleway/runs/1387987910?check_suite_focus=true#step:4:1776


scaleway/resource_baremetal_server_test.go
```diff
@@ -53,6 +53,14 @@ func TestAccScalewayBaremetalServer_Basic(t *testing.T) {
 		ProviderFactories: tt.ProviderFactories,
 		CheckDestroy:      testAccCheckScalewayBaremetalServerDestroy(tt),
 		Steps: []resource.TestStep{
+			{
```
jerome-quere, line 56: Why do we need to add a step ?

PR #696, opened 2020-11-13
chore: Enable explicit default timeout on all resources

remyleone:


scaleway/resource_baremetal_server.go
```diff
@@ -22,8 +22,8 @@ func resourceScalewayBaremetalServer() *schema.Resource {
 		},
 		SchemaVersion: 0,
 		Timeouts: &schema.ResourceTimeout{
-			Create: &baremetalServerResourceTimeout,
-			Delete: &baremetalServerResourceTimeout,
+			Create: schema.DefaultTimeout(baremetalServerResourceTimeout),
```
jerome-quere, line 25: Why not use default ?

scaleway/resource_instance_ip_reverse_dns.go
```diff
@@ -18,6 +18,9 @@ func resourceScalewayInstanceIPReverseDNS() *schema.Resource {
 		Importer: &schema.ResourceImporter{
 			StateContext: schema.ImportStatePassthroughContext,
 		},
+		Timeouts: &schema.ResourceTimeout{
+			Default: schema.DefaultTimeout(instanceIPTimeout),
```
jerome-quere, line 22: I wounder if we would not be easier to read if time was directly specified without using a varibale

jerome-quere (changes requested):

scaleway/resource_instance_ip_reverse_dns.go
remyleone: I would prefer to have all timeout for a given set of resources be defined in a single place that allows for easy modification and reading once when we want to change the default for a given set of resources.

jerome-quere (changes requested):
We should standardize the name of all timeout vars
```
default<ProductName><RessourceName>Timeout = XX;
```

jerome-quere (approved):

PR #697, opened 2020-11-13
chore: add support for sentry crash reports

remyleone:


scaleway/provider.go
```diff
@@ -26,6 +33,19 @@ func DefaultProviderConfig() *ProviderConfig {
...
+		Release: version,
+	})
+	if err != nil {
+		log.Fatalf("sentry.Init: %s", err)
+	}
+	// Flush buffered events before the program terminates.
+	// Set the timeout to the maximum duration the program can afford to wait.
+	defer sentry.Flush(2 * time.Second)
+
+	sentry.CaptureMessage("It works!")
```
kindermoumoute, line 47: remove this line

scaleway/provider.go
```diff
@@ -26,6 +33,19 @@ func DefaultProviderConfig() *ProviderConfig {
...
+	err := sentry.Init(sentry.ClientOptions{
+		Dsn:     dsn,
+		Release: version,
+	})
+	if err != nil {
+		log.Fatalf("sentry.Init: %s", err)
+	}
+	// Flush buffered events before the program terminates.
+	// Set the timeout to the maximum duration the program can afford to wait.
+	defer sentry.Flush(2 * time.Second)
```
kindermoumoute, line 45: Do we want to recover from panic and send a sentry the same way we do in the CLI?

kindermoumoute (changes requested):

scaleway/sentry.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+func logAndSentry(sentryClient *raven.Client, err error) {
+	log.Printf("%s", err)
+	if sentryClient != nil {
+		log.Printf("sending sentry report: %s", sentryClient.CaptureErrorAndWait(err, nil))
+	}
+}
+
+// newSentryClient create a sentry client with build info tags.
```
QuentinBrosse, line 44:
```suggestion
// newSentryClient creates a sentry client with build info tags.
```

QuentinBrosse (commented):
Just a NIT

scaleway/provider.go
```diff
@@ -26,6 +26,11 @@ func DefaultProviderConfig() *ProviderConfig {
 
 // Provider returns a terraform.ResourceProvider.
 func Provider(config *ProviderConfig) plugin.ProviderFunc {
+	// Catch every panic after this line. This will send an anonymous report on Scaleway's sentry.
```
jerome-quere, line 29: This should be done in the main as this function immediately exist so no error will be panic in this context

scaleway/sentry.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+
+	sentryClient, err := newSentryClient()
+	if err != nil {
+		log.Printf("cannot create sentry client: %s", err)
+	}
+
+	err, isError := e.(error)
+	if isError {
+		logAndSentry(sentryClient, err)
```
jerome-quere, line 29: I think error logging is handled by the terraform binary directly as we did not handle it before.

scaleway/sentry.go
```diff
@@ -0,0 +1,58 @@
...
+		logAndSentry(sentryClient, fmt.Errorf("unknownw error: %v", e))
+	}
+
+	panic(e) // lintignore:R009
+}
+
+func logAndSentry(sentryClient *raven.Client, err error) {
+	log.Printf("%s", err)
+	if sentryClient != nil {
+		log.Printf("sending sentry report: %s", sentryClient.CaptureErrorAndWait(err, nil))
```
jerome-quere, line 40: IMHO This should be a debug log

jerome-quere (changes requested):

scaleway/sentry.go
```diff
@@ -0,0 +1,58 @@
...
+		logAndSentry(sentryClient, err)
+	} else {
+		logAndSentry(sentryClient, fmt.Errorf("unknownw error: %v", e))
+	}
+
+	panic(e) // lintignore:R009
+}
+
+func logAndSentry(sentryClient *raven.Client, err error) {
+	log.Printf("%s", err)
```
jerome-quere, line 38: Remove this line

scaleway/sentry.go
jerome-quere, line 29: You count inline this function call

scaleway/sentry.go
```diff
@@ -0,0 +1,58 @@
...
+	client, err := raven.New(dsn)
+	if err != nil {
+		return nil, err
+	}
+
+	tagsContext := map[string]string{
+		"version": Version,
+	}
+
+	client.SetTagsContext(tagsContext)
```
jerome-quere, line 55: client.SetTagsContext(map[string]string{
		"version": Version,
	}

jerome-quere (changes requested):

remyleone:
One solution would be to have a handler on each Create/Read/Update/Delete function that would catch each panic and send it. At the moment, the panics are sent to terraform client directly and are not caught by our sentry function.

PR #702, opened 2020-11-16
fix(instance): handle unauthorized on placement group and ip

remyleone:


scaleway/resource_instance_placement_group.go
```diff
@@ -92,7 +92,7 @@ func resourceScalewayInstancePlacementGroupRead(ctx context.Context, d *schema.R
 	}, scw.WithContext(ctx))
 
 	if err != nil {
-		if is404Error(err) {
+		if is404Error(err) || is403Error(err) {
```
jerome-quere, line 95: This only happend when importing the default security group do we want to handle it ?

jerome-quere (changes requested):

jerome-quere (approved):

PR #714, opened 2020-11-23
fix(account): fix conflict when renaming a resource

remyleone:


remyleone:
Blocked by internal ticket AUTH-392

kindermoumoute:
I suggest we close this PR since this behavior is on the SSH key API side. (we can reopen it the day we change the API)

Monitob (approved):

PR #743, opened 2021-01-08
feat(instance): handle non-default case for delete on termination

remyleone:


kindermoumoute:
Closing this in favor of #749

PR #776, opened 2021-02-05
feat(as): add support for apple silicon M1-M

remyleone:


scaleway/helpers_apple_silicon.go
```diff
@@ -0,0 +1,37 @@
...
+import (
+	"time"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	applesilicon "github.com/scaleway/scaleway-sdk-go/api/applesilicon/v1alpha1"
+	"github.com/scaleway/scaleway-sdk-go/scw"
+)
+
+const (
+	defaultAppleSiliconServerTimeout = 60 * time.Minute
```
kindermoumoute, line 12: 2 minutes is enough 👍

scaleway/resource_apple_silicon_server.go
```diff
@@ -0,0 +1,178 @@
...
+		},
+		Importer: &schema.ResourceImporter{
+			StateContext: schema.ImportStatePassthroughContext,
+		},
+		SchemaVersion: 0,
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:        schema.TypeString,
+				Description: "Name of the server",
+				Optional:    true,
```
kindermoumoute, line 30: Can you allow computed name here? (same way we do for other products)

scaleway/resource_apple_silicon_server.go
```diff
@@ -0,0 +1,178 @@
...
+
+	return resourceScalewayAppleSiliconServerRead(ctx, d, m)
+}
+
+func resourceScalewayAppleSiliconServerDelete(ctx context.Context, d *schema.ResourceData, m interface{}) diag.Diagnostics {
+	asAPI, zone, ID, err := asAPIWithZoneAndID(m, d.Id())
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
```
kindermoumoute, line 167: nit: we could look at the `deletable_at` value from the state and throw an error if it's not deletable.. but it might introduce bugs if the client doesn't have a good clock 🤔

kindermoumoute (changes requested):

scaleway/resource_apple_silicon_server.go
remyleone: Yes I hesitate about this feature because it relies on an accurate clock that could be wrong. I think the safest approach is to code a complete WaitForPossibleDeletion function that would behave like a Wait and try periodically.

remyleone: https://github.com/scaleway/scaleway-sdk-go/issues/722

kindermoumoute (commented):
Can we disable nigtly tests on this?

remyleone:
@kindermoumoute Yes :-) it is currently the case. Nightly tests are enabled per product: https://github.com/scaleway/terraform-provider-scaleway/blob/master/.github/workflows/nightly.yml#L13

kindermoumoute (approved):
LGTM

PR #777, opened 2021-02-05
feat(rdb): add support for settings

remyleone:


scaleway/helpers_rdb.go
```diff
@@ -47,3 +47,25 @@ func flattenRdbInstanceReadReplicas(readReplicas []*rdb.Endpoint) interface{} {
 	}
 	return replicasI
 }
+
+func flattenInstanceSettings(settings []*rdb.InstanceSetting) interface{} {
+	res := map[string]string{}
```
kindermoumoute, line 52: nit

```suggestion
	res := make(map[string]string)
```

kindermoumoute (approved):
LGTM

PR #789, opened 2021-02-26
fix(instance): Fix inconsistant block volume detach

remyleone:


scaleway/resource_instance_volume.go
```diff
@@ -238,18 +238,20 @@ func resourceScalewayInstanceVolumeDelete(ctx context.Context, d *schema.Resourc
 	}
 
 	err = resource.Retry(d.Timeout(schema.TimeoutDelete), func() *resource.RetryError {
-		volumeResp, err := instanceAPI.GetVolume(&instance.GetVolumeRequest{
```
Monitob, line 241: Can you add the new way to [wait](https://github.com/scaleway/terraform-provider-scaleway/blob/master/scaleway/helpers_baremetal.go#L121) with the default timeouts

Monitob (approved):

PR #797, opened 2021-03-17
fix: fix creation requests locality when resource specified locality differs from the provider

remyleone:


scaleway/resource_instance_volume_test.go
```diff
@@ -258,3 +258,42 @@ func testAccCheckScalewayInstanceVolumeDestroy(tt *TestTools) resource.TestCheck
...
+					provider "scaleway" {
+						zone = "fr-par-1"
+						region = "fr-par"
+					}
+
+					resource "scaleway_instance_volume" "extended_disk" {
+						type             = "b_ssd"
+						name             = "test-instance-disk-1"
+						size_in_gb       = 1
+						zone             = "pl-waw-1"
```
jerome-quere, line 281: Are we sure sweeper are run in pl-waw-1 ?

scaleway/resource_apple_silicon_server.go
```diff
@@ -80,11 +80,17 @@ func resourceScalewayAppleSiliconServerCreate(ctx context.Context, d *schema.Res
 	}
 
 	createReq := &applesilicon.CreateServerRequest{
+		Zone:      zone,
 		Name:      expandOrGenerateString(d.Get("name"), "m1"),
 		Type:      d.Get("type").(string),
 		ProjectID: d.Get("project_id").(string),
 	}
 
+	if definedZone, ok := d.GetOk("zone"); ok {
```
jerome-quere, line 89: Stange this is already done in asAPIWithZone and all other functions

jerome-quere (changes requested):

PR #798, opened 2021-03-22
feat(k8s): remove dashboard and ingress fields

Sh4d1:

### DIFF
--- a/docs/data-sources/k8s_cluster.md
+++ b/docs/data-sources/k8s_cluster.md
@@ -64,10 +64,6 @@ In addition to all above arguments, the following attributes are exported:
 
 - `cni` - The Container Network Interface (CNI) for the Kubernetes cluster.
 
-- `enable_dashboard` - True if the [Kubernetes dashboard](https://github.com/kubernetes/dashboard) is enabled for the Kubernetes cluster.
-
-- `ingress` - The [ingress controller](https://kubernetes.io/docs/concepts/services-networking/ingress-controllers/) deployed on the Kubernetes cluster.
-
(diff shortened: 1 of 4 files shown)


remyleone (approved):

TommyTheTribe:
What is the reason for deleting this please ?

Sh4d1:
Hey @TommyTheTribe !
- the dashboard field is now deprecated since you can access it via the Scaleway console without it being deployed on the cluster
- the ingresses will soon be deprecated, and it was easier to remove them from the 2.0 release before, rather than depreciating them after the 2.0 relase was cut. A replacement for the ingresses is on its way, not sure il will be available in Terraform though 🤔 in either case, best way is still to deploy the ingress on your own so you're in full control of the configuration :)

TommyTheTribe:
Ok thanks for response ;)

chinochao:
The dashboard does seems deprecated in the API for Scaleway. 

Is the ingress also getting deprecated from the API? If you guys are not using the addon ingress which does so much more like the ability to deploy LB from Kuberentes then how do you guys accomplish deploying a LB in Scaleway from Kuberentes? I am referring to this. https://www.scaleway.com/en/docs/using-a-load-balancer-to-expose-your-kubernetes-kapsule-ingress-controller-service/

Sh4d1:
@chinochao the ingress is also going to be deprecated soon. There will be a replacement (maybe not available in Terraform though). 
You can either deploy the plain manifest (https://kubernetes.github.io/ingress-nginx/deploy/#scaleway for instance)
Or use Helm on the ingress of your choice :)

PR #816, opened 2021-04-23
feat: only return warnings for non public locality

remyleone:


scaleway/helpers.go
```diff
@@ -275,15 +274,28 @@ func regionSchema() *schema.Schema {
...
+		Description:      "The region you want to attach the resource to",
+		Optional:         true,
+		ForceNew:         true,
+		Computed:         true,
+		ValidateDiagFunc: validateWithWarnings(allRegions, "region"),
+	}
+}
+
+// validateWithWarnings helps to only returns warnings in case we got a non public locality passed
+func validateWithWarnings(correctValues []string, field string) func(i interface{}, path cty.Path) diag.Diagnostics {
```
jerome-quere, line 287: `validateLocality` ?

remyleone: The code is not specific to Locality so it was unclear whether we might need it for something else.

jerome-quere: `validateStringInSliceWithWarning` ?

jerome-quere (changes requested):

jerome-quere (approved):

PR #835, opened 2021-05-17
test(rdb/Volume): Block storage for RDB is in beta only in WaW & AMS

jeansebastienh:


scaleway/resource_rdb_instance_test.go
```diff
@@ -168,6 +168,7 @@ func TestAccScalewayRdbInstance_Volume(t *testing.T) {
 						disable_backup = true
 						user_name = "my_initial_user"
 						password = "thiZ_is_v&ry_s3cret"
+						region= "nl-ams"
```
remyleone, line 171: Why do we need to change it here?

remyleone (approved):

PR #836, opened 2021-05-17
fix(instance/volume): Wait for the volume after resizing it

jeansebastienh:
The test was failing although the volume was properly resized.

The provider was waiting **before** resizing the volume.

```bash
--- PASS: TestAccScalewayInstanceVolume_ResizeBlock (18.26s)                                                           
PASS                                                                                                                   
ok      github.com/scaleway/terraform-provider-scaleway/scaleway        18.271s  
```

### DIFF
--- a/scaleway/resource_instance_volume.go
+++ b/scaleway/resource_instance_volume.go
@@ -171,6 +171,14 @@ func resourceScalewayInstanceVolumeUpdate(ctx context.Context, d *schema.Resourc
 		if oldSize, newSize := d.GetChange("size_in_gb"); oldSize.(int) > newSize.(int) {
 			return diag.FromErr(fmt.Errorf("block volumes cannot be resized down"))
 		}
+		_, err := instanceAPI.WaitForVolume(&instance.WaitForVolumeRequest{
+			VolumeID:      id,
+			Zone:          zone,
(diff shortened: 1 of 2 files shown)


scaleway/resource_instance_volume.go
```diff
@@ -171,15 +171,6 @@ func resourceScalewayInstanceVolumeUpdate(ctx context.Context, d *schema.Resourc
 		if oldSize, newSize := d.GetChange("size_in_gb"); oldSize.(int) > newSize.(int) {
 			return diag.FromErr(fmt.Errorf("block volumes cannot be resized down"))
 		}
-		_, err := instanceAPI.WaitForVolume(&instance.WaitForVolumeRequest{
```
remyleone, line 174: I think we need both. We cannot update if it is not in a stable state

jeansebastienh: Hum, we are in an update context so the volume should be in a stable state. If not the API should reject the call IMO.

remyleone: Maybe another operation coming from the console or the API through the CLI altered the state of the volume. Putting a wait is not a problem, at worst it will do a request, see it is ok and move on.

remyleone (approved):

PR #842, opened 2021-05-20
test(Lb/Backend): Removing unused ip declaration from the first step

jeansebastienh:
In the first step, `scaleway_instance_ip.ip02` is not used so let's remove
it. In fact, cassette is not able to distingish the HTTP calls for
`resource scaleway_instance_ip ip01 {}` and `resource
scaleway_instance_ip ip02 {}` as they are identicals making the
acceptance tests failing


scaleway/resource_lb_backend_test.go
```diff
@@ -27,7 +27,6 @@ func TestAccScalewayLbBackend_Basic(t *testing.T) {
 					}
 
 					resource scaleway_instance_ip ip01 {}
-					resource scaleway_instance_ip ip02 {}
```
remyleone, line 30: I would be in favor of adding it to the backend and test that adding several IP works reliably :)

jeansebastienh: I think it won't work with cassette for the reason I described above.

I will try to found a workaround

jeansebastienh: In fact, it's working https://github.com/scaleway/terraform-provider-scaleway/pull/842/commits/67b07f79828b2d06be00a1edf6b0335f2fa8464e

PR #859, opened 2021-07-16
feat(iot): Add ability to set user device certificate

grom-42:


scaleway/resource_iot_device.go
```diff
@@ -141,18 +130,31 @@ func resourceScalewayIotDevice() *schema.Resource {
 					Schema: map[string]*schema.Schema{
 						"crt": {
 							Type:        schema.TypeString,
+							Optional:    true,
 							Computed:    true,
```
Monitob, line 134:
```suggestion
							Optional:    true,
							ForceNew:      true,
							Computed:    true,
```
 Please consider add ForceNew

scaleway/resource_iot_device.go
```diff
@@ -141,18 +130,31 @@ func resourceScalewayIotDevice() *schema.Resource {
 					Schema: map[string]*schema.Schema{
 						"crt": {
 							Type:        schema.TypeString,
```
Monitob, line 132:
```suggestion
							Type:        schema.TypeString,
							Required:    true,
```
This should be required

scaleway/resource_iot_device.go
```diff
@@ -319,11 +335,32 @@ func resourceScalewayIotDeviceRead(ctx context.Context, d *schema.ResourceData,
 		_ = d.Set("message_filters", []map[string]interface{}{mf})
 	}
 
+	////
+	// Read Device certificate
+	////
+
+	// As we cannot read the key, we get back it from cache and do not change it.
```
Monitob, line 342: @grom-42 I can't find any code related with your comment. Can you give an inside?

scaleway/resource_iot_device.go
```diff
@@ -242,12 +244,26 @@ func resourceScalewayIotDeviceCreate(ctx context.Context, d *schema.ResourceData
...
-		"crt": res.Certificate.Crt,
-		"key": res.Certificate.Key,
+	// If user certifcate is provided.
+	if devCrt, ok := d.GetOk("certificate.0.crt"); ok {
+		// Set user certificate to device.
+		// It cannot currently be added in the create device request.
+		_, err := iotAPI.SetDeviceCertificate(&iot.SetDeviceCertificateRequest{
+			Region:         region,
+			DeviceID:       res.Device.ID,
+			CertificatePem: devCrt.(string),
```
Monitob, line 254: Please add a safer method to cast this. e.g.
`s, ok := devCrt.(string)
if !ok {
    //  manage your error
}`

scaleway/resource_iot_device.go
```diff
@@ -319,11 +335,32 @@ func resourceScalewayIotDeviceRead(ctx context.Context, d *schema.ResourceData,
...
+			Region:   region,
+			DeviceID: deviceID,
+		}, scw.WithContext(ctx))
+		if err != nil {
+			return diag.FromErr(err)
+		}
+		// Set device certificate.
+		cert := map[string]interface{}{
+			"crt": devCrt.CertificatePem,
+			"key": devCrtKey.(string),
```
Monitob, line 354: idem

scaleway/resource_iot_device.go
grom-42: As discussed we do not want to trigger the certificate re-creation if the user updates its certificate.

scaleway/resource_iot_device.go
grom-42: As discussed, do not provide a custom certificate is a legit use case, so it is not mandatory.

scaleway/resource_iot_device.go
grom-42: As discussed, for coherency with the basecode we will let as is for now and take it in consideration for future improvements.

scaleway/resource_iot_device.go
grom-42: idem :)

Monitob (approved):

PR #860, opened 2021-07-22
Update registry_namespace.md

juli3nk:
Fix typo in the name argument example.

I have a non-obvious error in the return `Namespace already exist`. While no namespace exists and the problem seems to be that the underscore is not allowed in the name. Replacing it with a dash works.


remyleone (approved):

remyleone:
@juli3nk Thanks a lot :) indeed only alphanumerical characters and `-` are allowed

PR #867, opened 2021-08-27
Add support for LB Private Networks and LB Routes

Tuetuopay:
As the title says, this PR adds the resource `scaleway_lb_private_network` and `scaleway_lb_route` to the provider.

### DIFF
--- a/scaleway/testdata/lb-private-network-basic.cassette.yaml
+++ b/scaleway/testdata/lb-private-network-basic.cassette.yaml
@@ -13,7 +13,7 @@ interactions:
     url: https://api.scaleway.com/vpc/v1/zones/fr-par-1/private-networks
     method: POST
   response:
-    body: '{"id":"afbabef1-f79e-4350-81be-98e67341dcb2","name":"test-lb-pn","tags":[],"organization_id":"63a66ec9-a385-4194-bc15-04aa6921274a","created_at":"2021-10-22T09:38:53.011189Z","updated_at":"2021-10-22T09:38:53.011189Z","project_id":"63a66ec9-a385-4194-bc15-04aa6921274a","zone":"fr-par-1"}'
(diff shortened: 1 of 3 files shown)


scaleway/helpers_lb.go
```diff
@@ -293,3 +293,33 @@ func lbUpgradeV1RegionalToZonedID(element string) (string, error) {
...
+		return nil
+	}
+	return cfg.IPAddress
+}
+
+func expandLbPrivateNetworkDHCPConfig(raw interface{}) *lb.PrivateNetworkDHCPConfig {
+	if raw == nil || len(raw.([]interface{})) < 1 {
+		return nil
+	}
+	return &lb.PrivateNetworkDHCPConfig{}
```
remyleone, line 317: Is it normal that the returned value does not depend on raw?

scaleway/resource_lb_private_network.go
```diff
@@ -0,0 +1,186 @@
...
+		CreateContext: resourceScalewayLbPrivateNetworkCreate,
+		ReadContext:   resourceScalewayLbPrivateNetworkRead,
+		DeleteContext: resourceScalewayLbPrivateNetworkDelete,
+		Importer: &schema.ResourceImporter{
+			StateContext: schema.ImportStatePassthroughContext,
+		},
+		Timeouts: &schema.ResourceTimeout{
+			Default: schema.DefaultTimeout(defaultLbLbTimeout),
+		},
+		SchemaVersion: 1,
```
remyleone, line 24:
```suggestion
		SchemaVersion: 0,
```

scaleway/helpers_lb.go
Tuetuopay: Yes, the information is just `PrivateNetworkDHCPConfig{}` to tell to use DHCP, no more data is present in this struct.

PR #917, opened 2021-11-03
feat(lb): add support for private network in lb resource

Monitob:
- Support private network on a load balancer


scaleway/resource_lb.go
```diff
@@ -71,6 +72,48 @@ func resourceScalewayLb() *schema.Resource {
...
+					Schema: map[string]*schema.Schema{
+						"private_network_id": {
+							Type:         schema.TypeString,
+							Required:     true,
+							ForceNew:     true,
+							ValidateFunc: validationUUIDorUUIDWithLocality(),
+							Description:  "The Private Network ID",
+						},
+						"static_config": {
+							Description: "Define two local IP addresses of your choice for each load balancer instance",
```
remyleone, line 90:
```suggestion
							Description: "Define two IP addresses in the subnet of your private network that will be assigned for the principal and standby node of your load balancer.",
```

remyleone: How can I know the one used for the principal and the one used by the standby?

scaleway/resource_lb.go
```diff
@@ -71,6 +72,48 @@ func resourceScalewayLb() *schema.Resource {
...
+							Optional:    true,
+							Computed:    true,
+							ForceNew:    true,
+							Elem: &schema.Schema{
+								Type:         schema.TypeString,
+								ValidateFunc: validation.IsIPAddress,
+							},
+						},
+						"dhcp_config": {
+							Description: "Set to true if you want to let DHCP assign IP addresses",
```
remyleone, line 101: IP addresses to the two underlying nodes?

scaleway/resource_lb_test.go
```diff
@@ -80,9 +89,39 @@ func TestAccScalewayLbLb_WithIP(t *testing.T) {
...
+					resource scaleway_vpc_private_network pnLB01 {
+						name = "pn-with-lb-static"
+					}
+
+					resource scaleway_lb lb01 {
+					    ip_id = scaleway_lb_ip.ip01.id
+						name = "test-lb-with-static"
+						type = "LB-S"
+						release_ip = true
+						private_network {
```
remyleone, line 102: Could you try to have a test where you have two private networks? One where you have 2 private networks and then you delete one?

scaleway/resource_lb.go
```diff
@@ -204,17 +282,50 @@ func resourceScalewayLbDelete(ctx context.Context, d *schema.ResourceData, meta
...
 		Timeout:       scw.TimeDurationPtr(LbWaitForTimeout),
-		RetryInterval: DefaultWaitRetryInterval,
+		RetryInterval: &retryInterval,
 	}, scw.WithContext(ctx))
-
 	if err != nil && !is404Error(err) {
 		return diag.FromErr(err)
 	}
 
+	if currentLB.PrivateNetworkCount != 0 {
```
remyleone, line 297: Where do we attach a new private network in case I've added one in a second step? I don't see any attach in this function.

scaleway/resource_lb_test.go
remyleone: And one where you have one private network initially and you add a second one in a second step.

Monitob: Only one private network is allowed for now.

scaleway/resource_lb.go
```diff
@@ -190,10 +268,72 @@ func resourceScalewayLbUpdate(ctx context.Context, d *schema.ResourceData, meta
...
+			Zone:          zone,
+			LBID:          ID,
+			Timeout:       scw.TimeDurationPtr(defaultInstanceServerWaitTimeout),
+			RetryInterval: &retryInterval},
+			scw.WithContext(ctx))
+		if err != nil && !is404Error(err) {
+			return diag.FromErr(err)
+		}
+		// detach private networks
+		for _, pn := range pns {
```
remyleone, line 291: If we detach all private networks by default it means that ongoing connections will be lost if there is one between the load-balancer and another host in the private network. We will need to ensure that no connection is dropped before doing a release.

scaleway/helpers_lb.go
```diff
@@ -83,6 +84,122 @@ func flattenLbACLAction(action *lb.ACLAction) interface{} {
...
+	return res, nil
+}
+
+func isPrivateNetworkEqual(A, B interface{}) bool {
+	// Find out the diff Private Network or not
+	if _, ok := A.(*lb.PrivateNetwork); ok {
+		if _, ok := B.(*lb.PrivateNetwork); ok {
+			if A.(*lb.PrivateNetwork).PrivateNetworkID == B.(*lb.PrivateNetwork).PrivateNetworkID {
+				// if both has dhcp config should not update
+				if A.(*lb.PrivateNetwork).DHCPConfig != nil && B.(*lb.PrivateNetwork).DHCPConfig != nil {
```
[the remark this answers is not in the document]
Monitob: There is not data exported for DHCP via this api  to compare. As result from a GET you are gonna obtain always an empty json. `{}`

scaleway/helpers_lb_test.go
```diff
@@ -0,0 +1,53 @@
+package scaleway
+
+import (
+	"testing"
+
+	"github.com/scaleway/scaleway-sdk-go/api/lb/v1"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestIsEqualPrivateNetwork(t *testing.T) {
```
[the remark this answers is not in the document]
Monitob: There is not variables exported on this [type](https://github.com/scaleway/scaleway-sdk-go/blob/29e472beb5cd81eceb91164ac3ee4236493f1739/api/lb/v1/lb_sdk.go#L1498)

remyleone (approved):

PR #937, opened 2021-11-23
fix(rdb): populate waitinstance on related resources

Monitob:
- fix(rdb): populate waiting and retry
- fix(rdb): get region from id and retry on database creation


scaleway/resource_rdb_user.go
```diff
@@ -70,7 +70,7 @@ func resourceScalewayRdbUserCreate(ctx context.Context, d *schema.ResourceData,
 
 	ins, err := waitInstance(ctx, rdbAPI, region, instanceID)
 	if err != nil {
-		return diag.FromErr(err)
+		diag.FromErr(err)
```
remyleone, line 73: Why is this change needed?

scaleway/resource_rdb_instance.go
```diff
@@ -321,7 +321,7 @@ func resourceScalewayRdbInstanceCreate(ctx context.Context, d *schema.ResourceDa
 		}
 
 		_, err = waitInstance(ctx, rdbAPI, region, res.ID)
-		if err != nil && !is404Error(err) {
+		if err != nil {
```
remyleone, line 324: Why are those changes needed? Should this kind of change be present on all other resources that check for 404 errors?

scaleway/resource_rdb_database.go
```diff
@@ -58,52 +60,91 @@ func resourceScalewayRdbDatabase() *schema.Resource {
...
 		return diag.FromErr(err)
 	}
 
-	d.SetId(resourceScalewayRdbDatabaseID(region, expandID(instanceID), res.Name))
+	d.SetId(resourceScalewayRdbDatabaseID(region, instanceID, db.Name))
 
 	return resourceScalewayRdbDatabaseRead(ctx, d, meta)
 }
 
+func getDatabase(ctx context.Context, api *rdb.API,
```
remyleone, line 108: Could the signature be on a single line to be diff friendly?

scaleway/resource_rdb_database.go
```diff
@@ -58,52 +60,91 @@ func resourceScalewayRdbDatabase() *schema.Resource {
...
 	createReq := &rdb.CreateDatabaseRequest{
 		Region:     region,
 		InstanceID: instanceID,
 		Name:       d.Get("name").(string),
 	}
 
-	res, err := rdbAPI.CreateDatabase(createReq, scw.WithContext(ctx))
+	var db *rdb.Database
+	//  wrapper around StateChangeConf that will just retry the database creation
+	err = resource.RetryContext(context.Background(), 2*time.Minute, func() *resource.RetryError {
```
remyleone, line 81: Could this 2*time.Minute be configurable?

scaleway/resource_rdb_user.go
Monitob: oops copy/paste error. I've juste added the return

scaleway/resource_rdb_instance.go
Monitob: After a review of code, I remove is404Error where I consider where I think is necessary that the instance must exist. Which protect other request with an InstanceID not null and in a stable mode.

remyleone (approved):

PR #945, opened 2021-11-30
fix(iot): cassettes

Monitob:


remyleone:
What does this PR fixes? The nightly test for iot seems to be passing

Monitob:
@remyleone Yes the nightly test are fixed but the cassettes seems not updated that provoques all pipelines.

PR #953, opened 2021-12-02
fix(lb): rename loadbalancer

Monitob:
- fix(lb): rename test data source load balancer [error](https://github.com/scaleway/terraform-provider-scaleway/runs/4389230988?check_suite_focus=true#step:4:11862)


scaleway/data_source_lb_test.go
```diff
@@ -21,7 +21,7 @@ func TestAccScalewayDataSourceLb_Basic(t *testing.T) {
 
 					resource scaleway_lb main {
 					    ip_id = scaleway_lb_ip.main.id
-						name = "test-lb"
+						name = "data-test-lb"
```
remyleone, line 24: What does this test fix?

Monitob: It's a  terraform error: `Error: 2 lbs found with the same name test-lb` when launching in terraform apply in parallel with resources. There is no error from our apis.

remyleone: It is really strange. Could you add a test that it is legit to have two lb with the same name? Normally they are made unique by their id, not their name.

remyleone: I think we can keep the name it removes a line from the diff :)

remyleone (approved):

PR #957, opened 2021-12-03
fix cloud-init

Monitob:


scaleway/resource_instance_server_test.go
```diff
@@ -360,14 +356,12 @@ func TestAccScalewayInstanceServer_UserData_WithoutCloudInitAtStart(t *testing.T
...
 				// Without cloud-init
 				Config: `
-					data "scaleway_instance_image" "ubuntu" {
-						architecture = "x86_64"
-						name         = "Ubuntu 20.04 Focal Fossa"
-					}
-
 					resource "scaleway_instance_server" "base" {
-						image = "${data.scaleway_instance_image.ubuntu.id}"
+						image = "ubuntu-focal"
```
remyleone, line 360: is it normal that it is `ubuntu-focal` here and `ubuntu_focal` down there ?

scaleway/resource_instance_server_test.go
```diff
@@ -378,16 +372,13 @@ func TestAccScalewayInstanceServer_UserData_WithoutCloudInitAtStart(t *testing.T
...
-						name         = "Ubuntu 20.04 Focal Fossa"
-					}
-
 					resource "scaleway_instance_server" "base" {
-						image = "${data.scaleway_instance_image.ubuntu.id}"
+						image = "ubuntu_focal"
 						type  = "DEV1-S"
 						tags  = [ "terraform-test", "scaleway_instance_server", "user_data" ]
-
+						root_volume {
```
remyleone, line 379: Why is this change required?

scaleway/resource_instance_server_test.go
Monitob: The label on instance server supports the use of '-' and '_'

remyleone: I suggest using only one everywhere. By doing so, it is easier to migrate all occurrences to the next Ubuntu LTS once it is made available.

remyleone (approved):

PR #965, opened 2021-12-13
fix(lb): error check

Monitob:
* handle check 404 and 403 error on creation / read


scaleway/resource_lb.go
```diff
@@ -153,7 +153,7 @@ func resourceScalewayLbCreate(ctx context.Context, d *schema.ResourceData, meta
 		RetryInterval: &retryInterval,
 	}, scw.WithContext(ctx))
 	// check err waiting process
-	if err != nil && !is404Error(err) {
+	if err != nil {
```
remyleone, line 156: Why is this change required? When can we have a 404 in this case that is wanted?

Monitob: On creation process, we should wait an stable state and confirm that the resource was created.

remyleone (approved):

PR #970, opened 2021-12-15
fix(instance):  fix ip detach by adding wait resources calls

lde:
Fix public ip switch on instance_server


scaleway/resource_instance_server.go
```diff
@@ -678,6 +679,10 @@ func resourceScalewayInstanceServerUpdate(ctx context.Context, d *schema.Resourc
 			if err != nil {
 				return diag.FromErr(err)
 			}
+			//we wait to ensure to not detach the new ip.
+			if newIPID != "" {
```
remyleone, line 683: @Monitob could you amend this PR to have a wait based workflow? We should be able to see whether the IP got a status or if the server got a status updated with the IP. Checking with the instance team could be helpful.

Monitob: This resource has been updated to wait a stable state.

remyleone (dismissed):

remyleone (approved):

Monitob (approved):

PR #983, opened 2021-12-29
feat(lb): add wait lb & instances attached

Monitob:


scaleway/helpers_lb.go
```diff
@@ -17,6 +17,7 @@ import (
 const (
 	LbWaitForTimeout   = 10 * time.Minute
 	defaultLbLbTimeout = 10 * time.Minute
+	retryLbIPInterval  = 1 * time.Minute
```
remyleone, line 20: it seems very long to wait a minute per attempt. Could we make it smaller like 5 seconds or something like that?

Monitob: Yes let me try this and I keep you up

Monitob: The test were successful

remyleone (approved):

PR #984, opened 2021-12-30
fix(k8s): return error code from k8sGetLatestVersionFromMinor call

detached:
The `k8sGetLatestVersionFromMinor` call is the first interaction with the scaleway API when auto_update is enabled and the version is set to a x.y minor version. 
But a possible error message from the API call is never logged. So if you have a typo in your credentials or forgot to add the project_id to the provider, you are confronted with the error message "minor version x.y must be used with auto upgrade enable".

This PR just reports the error from `k8sGetLatestVersionFromMinor` to the user.


remyleone (approved):

remyleone:
Thanks a lot :)

PR #1020, opened 2022-01-06
Fix(lb): release_ip deprecated

Monitob:


docs/resources/lb.md
```diff
@@ -55,6 +55,8 @@ In addition to all arguments above, the following attributes are exported:
 - `ip_address` -  The load-balance public IP Address
 - `organization_id` - The organization ID the load-balancer is associated with.
 
+~> **Important:** `release_ip` will not be supported. This prevents the destruction of the IP from releasing a LBs.
+The `resource_lb_ip` will the only to handle those IPs.
```
remyleone, line 59:
```suggestion
The `resource_lb_ip` will be the only resource that handles those IPs.
```

scaleway/resource_lb.go
```diff
@@ -71,6 +71,7 @@ func resourceScalewayLb() *schema.Resource {
 				Optional:    true,
 				Default:     false,
 				Description: "Release the IPs related to this load-balancer",
+				Deprecated:  "The resource ip will be destroyed by it's own resource. Please set this to `false`",
```
remyleone, line 74: Could you include a migration guide in case this resource is deprecated? For instance by showing an example with a scaleway_lb that got it to true and show how to migrate with the now two distinct resources.

Monitob: Thank you,  I've just added.

remyleone (approved):

Issue #1023, opened 2022-01-06
[Domain record] dns_zone is not set in tfstate when using terraform import


paulden:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform Version

* Terraform version : `1.0.7`
* Scaleway provider: `2.2.0`

### Affected Resource(s)

* scaleway_domain_record

### Terraform Configuration Files

```hcl
resource "scaleway_domain_record" "test" {
  dns_zone = "domain.tld"
  name     = "test"
  data     = "1.2.3.4"
  ttl      = 600
  type     = "A"
}
```

### Expected Behavior

When importing resource to state (using the [documentation](https://registry.terraform.io/providers/scaleway/scaleway/latest/docs/resources/domain_record#import) provided), it should import the `dns_zone`: 

```
$ terraform import scaleway_domain_record.test domain.tld/11111111-1111-1111-1111-111111111111
$ terraform state show scaleway_domain_record.test
# scaleway_domain_record.records_test:
resource "scaleway_domain_record" "records_test" {
    data       = "1.2.3.4"
    dns_zone   = "domain.tld"
    id         = "11111111-1111-1111-1111-111111111111"
    name       = "test"
    priority   = 0
    project_id = "11111111-1111-1111-1111-111111111111"
    ttl        = 600
    type       = "A"

    timeouts {}
}
```

### Actual Behavior

`dns_zone` is not present in tfstate after import : 

```
$ terraform import scaleway_domain_record.test domain.tld/11111111-1111-1111-1111-111111111111
$ terraform state show scaleway_domain_record.test
# scaleway_domain_record.records_test:
resource "scaleway_domain_record" "records_test" {
    data       = "1.2.3.4"
    id         = "11111111-1111-1111-1111-111111111111"
    name       = "test"
    priority   = 0
    project_id = "11111111-1111-1111-1111-111111111111"
    ttl        = 600
    type       = "A"

    timeouts {}
}
```

It causes errors when trying to plan / apply after that since the scaleway-sdk-go cannot read a domain record when not specifying the zone.

### Steps to Reproduce

[...]

PR #1024, opened 2022-01-06
fix(domain-record): Add dns_zone to tfstate when using terraform import

paulden:
Related to https://github.com/scaleway/terraform-provider-scaleway/issues/1023

When importing a domain record, it goes into the first `if` : https://github.com/scaleway/terraform-provider-scaleway/blob/master/scaleway/resource_domain_record.go#L267

By the end, the `dnsZone` that is parsed is never set : https://github.com/scaleway/terraform-provider-scaleway/blob/master/scaleway/resource_domain_record.go#L340

I suggest here setting the `dns_zone` at the end, feedback welcome if you see other ways to fix this issue.

Tested locally on Terraform 1.0.7.


jerjako (approved):
Look good to me :+1:

PR #1027, opened 2022-01-07
feat(domain-zone): Add resource and datasource for domain zone

LucasBoisserie:


Monitob:
Thanks for the contribution @LucasBoisserie

Issue #1040, opened 2022-01-13
Can't create instances in fr-par-2 ?


vdmkenny:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform Version

1.1.3

### Affected Resource(s)

* scaleway_instance_server

### Terraform Configuration Files

```hcl
resource "scaleway_instance_server" "haproxy_1" {
  name = "haproxy-1"
  type = "DEV1-S"
  image = "rockylinux_8"
  project_id = var.project_id
  zone = "fr-par-1"

  tags = [var.environment]

  private_network {
    pn_id = scaleway_vpc_private_network.internal.id
  }
}

resource "scaleway_instance_server" "haproxy_2" {
  name = "haproxy-2"
  type = "DEV1-S"
  image = "rockylinux_8"
  project_id = var.project_id
  zone = "fr-par-2"

  tags = [var.environment]

  private_network {
    pn_id = scaleway_vpc_private_network.internal.id
  }
}

```
### Panic Output

```
scaleway_instance_server.haproxy_2: Creating...
scaleway_instance_server.haproxy_1: Modifying... [id=fr-par-1/96f1e380-3488-4e37-920f-8865643ef9f7]
scaleway_instance_server.haproxy_1: Modifications complete after 2s [id=fr-par-1/96f1e380-3488-4e37-920f-8865643ef9f7]
scaleway_instance_server.haproxy_2: Still creating... [10s elapsed]
...
scaleway_instance_server.haproxy_2: Still creating... [1m40s elapsed]
╷
│ Error: scaleway-sdk-go: resource unknown with ID  is not found
│ 
│   with scaleway_instance_server.haproxy_2,
│   on instances.tf line 15, in resource "scaleway_instance_server" "haproxy_2":
│   15: resource "scaleway_instance_server" "haproxy_2" {
│ 
╵
ERROR: Job failed: exit code 1
```

### Expected Behavior

With this configuration, it should have created an instance in fr-par-1 and fr-par-2 and exited gracefully

### Actual Behavior

[...]

PR #1043, opened 2022-01-14
Fix(instance): check private network and add test case

Monitob:
Related to #1040


scaleway/resource_instance_server_test.go
```diff
@@ -935,6 +962,7 @@ func TestAccScalewayInstanceServer_PrivateNetwork(t *testing.T) {
 					resource "scaleway_instance_server" "base" {
 					  image = "ubuntu_focal"
 					  type  = "DEV1-S"
+                      zone  = "fr-par-1"
```
remyleone, line 965: indentation error here

remyleone (approved):

PR #1045, opened 2022-01-17
Fix(k8s): set waiting pool ready as mandatory

Monitob:


scaleway/resource_k8s_pool.go
```diff
@@ -277,15 +277,15 @@ func resourceScalewayK8SPoolCreate(ctx context.Context, d *schema.ResourceData,
 
 	d.SetId(newRegionalIDString(region, res.ID))
 
-	if waitForCluster {
-		_, err = waitK8SCluster(ctx, k8sAPI, region, cluster.ID)
+	if d.Get("wait_for_pool_ready").(bool) { // wait for the pool to be ready if specified (including all its nodes)
+		err = waitK8SPoolReady(ctx, k8sAPI, region, res.ID)
```
remyleone, line 281: Why are they swapped? before we wait first for cluster and then for pool and now it is the opposite.

Monitob: Create a cluster without a pool, it must go to `pool_required`, it is the creation of a pool that triggers the creation of the cluster. Verify the pool's state enhance the cluster creation correctly.  if the pool is deleted immediately, then the cluster will remain in creating endlessly.

remyleone (approved):

Issue #1054, opened 2022-01-22
Panic: index out of range with rdb_privilege


bLuka:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform Version

`Terraform v1.1.2`

### Affected Resource(s)

* `scaleway_rdb_privilege`

### Terraform Configuration Files

I don't think this would be meaningful. It looks to be tied to an unexpected or not properly handled API response.

### Panic Output

[See here](https://gist.github.com/bLuka/10882f2e3263185bc987e1fa46c3c966).

### Expected Behavior

I expected to see the Terraform plan, as it always shown to me.

### Actual Behavior

Terraform panics on the Scaleway provider, even though I did not modify anything related to scaleway_rdb_privilege after the last working plan.

### Steps to Reproduce

1. `terraform plan`

### Important Factoids

I have 2 rdb instances created on that project. Both have 2 users. Instance A has 1 admin, and 1 user **with no privilege**, while instance B has 1 admin, and 1 user with 2 privileges.

### References

It is related to this line: [scaleway/resource_rdb_privilege.go#L120](https://github.com/scaleway/terraform-provider-scaleway/blob/master/scaleway/resource_rdb_privilege.go#L120).

A quick fix would be to check the array's length before dereferencing, but what to do if array is empty? Error handling?

[API documentation here](https://developers.scaleway.com/en/products/rdb/api/#privileges-13f20d) don't mention 404 if no privileges are found ([as it is checked in the code here](https://github.com/scaleway/terraform-provider-scaleway/blob/0b4d02388f8dd161684428533125cd01b3f6d506/scaleway/resource_rdb_privilege.go#L113)). So I'm wondering if this would do the trick:

```go
	if err != nil {
		if is404Error(err) { // Can API really throw 404 on this?
			d.SetId("")
			return nil
		}
		return diag.FromErr(err)
	}
	// Additional check here:
	if len(res.Privileges) == 0 { // (Or res.TotalCount I think?)
		d.SetId("")
		return nil // Are we ok with that too?
	}

	var privilege = res.Privileges[0]
	// …
```

[...]

PR #1057, opened 2022-01-25
Fix(rdb-user- privileges): add check user and retry process

Monitob:


bLuka:
Hello,

I wanted to understand properly, and looked at your changes. I don't see any change from the place where the panic from #1054 happened. Do you mean the issue happened because of a failure during resource creation?

Thank you very much in advance for your answers!

Monitob:
> Hello,
> 
> I wanted to understand properly and look at your changes. I don't see any change from the place where the panic from #1054 happened. Do you mean the issue happened because of a failure during resource creation?
> 
> Thank you very much in advance for your answers!

Yes, I've noticed that the error from the API was not the right one.

Important points to understand the bug.

- The database name in the query was the instance name. (failure during the resource creation and the tfstate was not good)
- The code HTTP was 409 with not a clear message. Only state `unavailable`
- The privilege is a resource related to a user and a database. (this both resource doesn’t have an ID to work with)

Conclusion:

- The 409 means `the user does not exist` or `the instance is really doing something` so wait
- We have to check if the user really exist, otherwise, we skip
- To remove a user we need to put their privileges to none.
- The user and privilege could happen really fast so we check every time and it could be async.

The terraform provider allows us to manage the state of the resources and there are some cases that the state checks are not made.

Thank you for your @bLuka question and allow me to identify this. Please feel free to create a new issue for further fixes or contact me by our community slack.

scaleway/resource_rdb_privilege_test.go
```diff
@@ -27,18 +90,18 @@ func TestAccScalewayRdbPrivilege_Basic(t *testing.T) {
 						is_ha_cluster = false
 						tags = [ "terraform-test", "scaleway_rdb_user", "minimal" ]
 					}
-
+
```
remyleone, line 93: Could you remove those lines with whitespaces?

scaleway/resource_rdb_user.go
```diff
@@ -188,11 +192,26 @@ func resourceScalewayRdbUserDelete(ctx context.Context, d *schema.ResourceData,
 		return diag.FromErr(err)
 	}
 
-	err = rdbAPI.DeleteUser(&rdb.DeleteUserRequest{
-		Region:     region,
-		InstanceID: instanceID,
-		Name:       userName,
-	}, scw.WithContext(ctx))
+	err = resource.RetryContext(context.Background(), readWriteDataBaseTimeOut, func() *resource.RetryError {
```
remyleone, line 195: Don't we already have a ctx passed with resourceScalewayRdbPrivilegeCreate ?

scaleway/resource_rdb_user.go
```diff
@@ -82,15 +82,19 @@ func resourceScalewayRdbUserCreate(ctx context.Context, d *schema.ResourceData,
 	}
 
 	var user *rdb.User
-	//  wrapper around StateChangeConf that will just retry the write on database
+	//  wrapper around StateChangeConf that will just retry write on database
 	err = resource.RetryContext(context.Background(), readWriteDataBaseTimeOut, func() *resource.RetryError {
```
remyleone, line 86: Don't we already have a ctx passed with resourceScalewayRdbPrivilegeCreate ?

scaleway/resource_rdb_privilege.go
```diff
@@ -138,15 +178,49 @@ func resourceScalewayRdbPrivilegeUpdate(ctx context.Context, d *schema.ResourceD
...
 		DatabaseName: d.Get("database_name").(string),
-		UserName:     d.Get("user_name").(string),
+		UserName:     userName,
 		Permission:   rdb.Permission(d.Get("permission").(string)),
 	}
-	_, err = rdbAPI.SetPrivilege(updateReq, scw.WithContext(ctx))
-	if err != nil && !is404Error(err) {
+
+	//  wrapper around StateChangeConf that will just retry the database creation
+	err = resource.RetryContext(context.Background(), defaultRdbInstanceTimeout, func() *resource.RetryError {
```
remyleone, line 209: Don't we already have a ctx passed with resourceScalewayRdbPrivilegeCreate ?

scaleway/resource_rdb_privilege_test.go
```diff
@@ -18,6 +18,69 @@ func TestAccScalewayRdbPrivilege_Basic(t *testing.T) {
 		ProviderFactories: tt.ProviderFactories,
 		CheckDestroy:      testAccCheckScalewayRdbInstanceDestroy(tt),
 		Steps: []resource.TestStep{
+			{
+				Config: `
+					resource "scaleway_rdb_instance" "main" {
```
remyleone, line 23: What do you test with those resources? I don't see any assertion changing.

Monitob: I've just added a check for this  resource.

remyleone (approved):

PR #1058, opened 2022-01-26
fix(vpc_gw_dhcp): Properly set fields that should default to true

Tuetuopay:
Terraform will make sure that fields not present in the file will be set, so we never send nil to the api and get its defaults. Set the defaults ourselves to be consistent with the docs.

Intended to fix https://github.com/scaleway/terraform-provider-scaleway/issues/1059


scaleway/resource_vpc_public_gateway_dhcp_test.go
```diff
@@ -53,6 +53,17 @@ func testAccCheckScalewayVPCPublicGatewayDHCPExists(tt *TestTools, n string) res
 			return err
 		}
 
+		// Test default values
+		if !dhcp.EnableDynamic {
+			return fmt.Errorf("enable_dynamic is false, should defautl to true")
```
remyleone, line 58:
```suggestion
			return fmt.Errorf("enable_dynamic is false, should default to true")
```

scaleway/resource_vpc_public_gateway_dhcp_test.go
```diff
@@ -53,6 +53,17 @@ func testAccCheckScalewayVPCPublicGatewayDHCPExists(tt *TestTools, n string) res
...
+		}
+		if !dhcp.PushDefaultRoute {
+			return fmt.Errorf("push_default_route is false, should defautl to true")
```
remyleone, line 61:
```suggestion
			return fmt.Errorf("push_default_route is false, should default to true")
```

scaleway/resource_vpc_public_gateway_dhcp_test.go
```diff
@@ -53,6 +53,17 @@ func testAccCheckScalewayVPCPublicGatewayDHCPExists(tt *TestTools, n string) res
...
+		}
+		if !dhcp.PushDNSServer {
+			return fmt.Errorf("push_dns_server is false, should defautl to true")
```
remyleone, line 64:
```suggestion
			return fmt.Errorf("push_dns_server is false, should default to true")
```

scaleway/resource_vpc_public_gateway_dhcp_test.go
Tuetuopay: fixed

scaleway/resource_vpc_public_gateway_dhcp_test.go
Tuetuopay: fixed

scaleway/resource_vpc_public_gateway_dhcp_test.go
Tuetuopay: fixed

Monitob (dismissed):

scaleway/resource_vpc_public_gateway_dhcp.go
```diff
@@ -53,8 +53,8 @@ func resourceScalewayVPCPublicGatewayDHCP() *schema.Resource {
 			},
 			"enable_dynamic": {
 				Type:        schema.TypeBool,
-				Computed:    true,
 				Optional:    true,
+				Default:     true,
```
Monitob, line 57: This has an impact on the pat-rules test. You should just rerun the test to update the cassettes

Monitob (changes requested):

PR #1071, opened 2022-02-04
fix(lb): fix assignation of ip zone and make region computed

Monitob:


scaleway/helpers.go
```diff
@@ -310,8 +315,8 @@ func regionComputedSchema() *schema.Schema {
 	}
 }
 
-// validateStringInSliceWithWarning helps to only returns warnings in case we got a non public locality passed
-func validateStringInSliceWithWarning(correctValues []string, field string) func(i interface{}, path cty.Path) diag.Diagnostics {
+// ValidateStringInSliceWithWarning helps to only returns warnings in case we got a non public locality passed
```
remyleone, line 318: Why is this change required?

Monitob: Not necessary anymore. I'll rollback this to private

scaleway/resource_lb_ip.go
```diff
@@ -33,11 +33,17 @@ func resourceScalewayLbIP() *schema.Resource {
...
 			},
-			"region":          regionComputedSchema(),
-			"zone":            zoneSchema(),
+			"zone": {
+				Type:             schema.TypeString,
+				Description:      "The zone you want to attach the resource to",
+				Optional:         true,
+				ForceNew:         true,
+				ValidateDiagFunc: validateStringInSliceWithWarning(AllZones(), "zone"),
+				Default:          scw.ZoneFrPar1,
```
remyleone, line 42: It should be the zone of the provider.

Monitob: Yes, well seen. we should set zoneSchema back.

remyleone (approved):

PR #1088, opened 2022-02-14
feat(apple-silicon): add documentation and missing types

Monitob:


scaleway/resource_apple_silicon_server.go
```diff
@@ -48,7 +50,24 @@ func resourceScalewayAppleSiliconServer() *schema.Resource {
 				Description: "VNC url use to connect remotely to the desktop GUI",
 				Computed:    true,
 			},
-
+			"state": {
```
remyleone, line 53: I don't think you can voluntarly put the M1 in an error state

scaleway/resource_apple_silicon_server.go
```diff
@@ -48,7 +50,24 @@ func resourceScalewayAppleSiliconServer() *schema.Resource {
...
+				Type:        schema.TypeString,
+				Optional:    true,
+				Default:     applesilicon.ServerStatusStarting.String(),
+				Description: "The state of the server",
+				ValidateFunc: validation.StringInSlice([]string{
```
remyleone, line 58: I don't think it makes sense to validate that a user can put a M1 in an error state.

scaleway/resource_apple_silicon_server.go
```diff
@@ -48,7 +50,24 @@ func resourceScalewayAppleSiliconServer() *schema.Resource {
 				Description: "VNC url use to connect remotely to the desktop GUI",
 				Computed:    true,
 			},
-
+			"state": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Default:     applesilicon.ServerStatusStarting.String(),
```
remyleone, line 56: Should it be a computed state?

docs/resources/apple_silicon.md
```diff
@@ -0,0 +1,58 @@
...
+
+Creates and manages Scaleway Apple silicon M1. For more information,
+see [the documentation](https://www.scaleway.com/en/docs/compute/apple-silicon/concepts).
+
+## Examples
+
+### Basic
+
+```hcl
+    resource scaleway_apple_silicon_server server {
```
remyleone, line 16: Remove indentation

scaleway/resource_apple_silicon_server.go
Monitob: In fact, it must be only computed to be read only

scaleway/resource_apple_silicon_server.go
Monitob: The attribute will be read-only. we should keep the error in that case.

docs/resources/apple_silicon.md
Monitob: Removed in last commit

scaleway/resource_apple_silicon_server.go
remyleone: No. If it is read only terraform does not have any value here. Let's suppose that apple silicon add a new state tomorrow, it will break with this code. Better to remove it all together.

PR #1092, opened 2022-02-15
fix(domain): avoid root zone deletion attempt

jerjako:


Monitob:
@jerjako looks good this PR, but you need to remake the cassettes.

Monitob (approved):

PR #1110, opened 2022-02-24
fix(lb): parse load balander id in frontend

Monitob:


scaleway/resource_lb_frontend.go
```diff
@@ -153,14 +153,14 @@ func resourceScalewayLbFrontend() *schema.Resource {
...
+	lbAPI, zone, err := lbAPIWithZone(d, meta)
 	if err != nil {
 		return diag.FromErr(err)
 	}
 
-	zone, lbID, err := parseZonedID(d.Get("lb_id").(string))
-	if err != nil {
-		return diag.FromErr(err)
+	lbID := expandID(d.Get("lb_id"))
+	if len(lbID) == 0 {
```
remyleone, line 162: Why should it be a length list? why not compare it to an empty string?

scaleway/resource_lb_frontend.go
```diff
@@ -211,9 +211,9 @@ func resourceScalewayLbFrontendRead(ctx context.Context, d *schema.ResourceData,
 		return diag.FromErr(err)
 	}
 
-	_, lbID, err := parseZonedID(d.Get("lb_id").(string))
-	if err != nil {
-		return diag.FromErr(err)
+	lbID := expandID(d.Get("lb_id"))
+	if len(lbID) == 0 {
```
remyleone, line 215: Why should it be a length list? why not compare it to an empty string?

remyleone (changes requested):

scaleway/resource_lb_frontend.go
Monitob: Just changed

scaleway/resource_lb_frontend.go
Monitob: Just changed

remyleone (approved):

scaleway/resource_lb_frontend.go
```diff
@@ -153,14 +153,14 @@ func resourceScalewayLbFrontend() *schema.Resource {
...
-	lbAPI, _, err := lbAPIWithZone(d, meta)
+	lbAPI, zone, err := lbAPIWithZone(d, meta)
 	if err != nil {
 		return diag.FromErr(err)
 	}
 
-	zone, lbID, err := parseZonedID(d.Get("lb_id").(string))
-	if err != nil {
-		return diag.FromErr(err)
+	lbID := expandID(d.Get("lb_id"))
```
remyleone, line 161: What happens if the provider and the zone of id are different ?

Monitob: The lb should be at the same zone

remyleone (approved):

PR #1124, opened 2022-03-04
feat(instance): add support for tags for security groups

remyleone:


Monitob:
@remyleone can you add this to the doc, please. Otherwise LGTM.

Monitob (approved):

PR #1125, opened 2022-03-04
Add volume_type param in the root_volume of a server

clement-gilbert:


scaleway/resource_instance_server.go
```diff
@@ -312,31 +323,34 @@ func resourceScalewayInstanceServerCreate(ctx context.Context, d *schema.Resourc
...
-			}
+			volume_type = instance.VolumeVolumeTypeLSSD.String()
 		}
 	}
 
+	var size scw.Size
+	if size_input == 0 && volume_type == instance.VolumeVolumeTypeLSSD.String() {
+		// Compute the size so it will be valid against the local volume constraints
+		// It wouldn't be valid if another local volume is added but it's not the main use
+		// however in this case the user would be informed that it not fulfil the local volume constraints
```
remyleone, line 342:
```suggestion
		// however in this case the user would be informed that it does not fulfill the local volume constraints
```

remyleone (approved):

PR #1132, opened 2022-03-08
feat(instance): add support for tags in placement group

remyleone:


scaleway/resource_instance_placement_group.go
```diff
@@ -55,6 +55,14 @@ func resourceScalewayInstancePlacementGroup() *schema.Resource {
 				Computed:    true,
 				Description: "Is true when the policy is respected.",
 			},
+			"tags": {
```
Monitob, line 58: Can you add a test for this attribute, please?

Monitob (approved):

PR #1206, opened 2022-04-12
fix(instance): add support for handling empty security-groups tags

Monitob:


scaleway/resource_instance_security_group_test.go
```diff
@@ -354,7 +350,6 @@ func TestAccScalewayInstanceSecurityGroup_RemovePort(t *testing.T) {
 			{
 				Config: `
 					resource "scaleway_instance_security_group" "base" {
-						tags = [ "test-terraform" ]
```
remyleone, line 357: Why are you changing this test? This is precisely what we want to test. Maybe you can add a test with:
- no tags
- adding tags
- modifying tags
- deleting tags

Monitob: Yes, I've just added a dedicated test only for tags on `TestAccScalewayInstanceSecurityGroup_Tags` and also for the other resource_instance_XXX with tags

Monitob: I've just added new test for this

scaleway/resource_instance_security_group_test.go
```diff
@@ -38,7 +38,6 @@ func TestAccScalewayInstanceSecurityGroup_Basic(t *testing.T) {
 					resource "scaleway_instance_security_group" "base" {
 						name = "sg-name"
 						inbound_default_policy = "drop"
-						tags = [ "test-terraform" ]
```
Monitob, line 41: @remyleone  here I've removed to test the update of tags.

remyleone (approved):

remyleone (changes requested):
Fix ci

remyleone (approved):

PR #1222, opened 2022-04-20
fix(DHCP-reservation) update static ip and add test for update case

Monitob:


scaleway/resource_vpc_public_gateway_dhcp_reservation.go
```diff
@@ -150,11 +150,21 @@ func resourceScalewayVPCPublicGatewayDHCPReservationUpdate(ctx context.Context,
...
 
-	if d.HasChangesExcept("ip_address") {
+	if d.HasChanges("ip_address") {
 		ip := net.ParseIP(d.Get("ip_address").(string))
 		if ip == nil {
 			return diag.FromErr(fmt.Errorf("could not parse ip_address"))
 		}
+
+		gatewayNetworkID := expandID(d.Get("gateway_network_id"))
+		_, err = vpcgwAPI.WaitForGatewayNetwork(&vpcgw.WaitForGatewayNetworkRequest{
```
remyleone, line 160: Can we have a wait function that enable fast retries?

remyleone (approved):

PR #1229, opened 2022-04-21
feat(lb): add support for http_filter_option

remyleone:


scaleway/resource_lb_frontend.go
```diff
@@ -146,6 +147,11 @@ func resourceScalewayLbFrontend() *schema.Resource {
 											Type: schema.TypeString,
 										},
 									},
+									"http_filter_option": {
```
Monitob, line 150: Can we add this to the docs with the example please? This [link](https://developers.scaleway.com/en/products/lb/zoned_api/#http-filter-option-9d5c12) can be useful

scaleway/resource_lb_frontend.go
```diff
@@ -146,6 +147,11 @@ func resourceScalewayLbFrontend() *schema.Resource {
...
+										Type:        schema.TypeString,
```
Monitob, line 151: Should we consider any limit for this attribute. Like the length?. Also can this attribute be computed by any other tool like console or API call?

remyleone: It cannot be computed by the console/API call as it is a filter fully defined by the user. As for the length, it depends on each web server but those limits are usually in the order of 4kb to 48kb so I guess it is not worthwhile to add a constraint on the terraform side because we don't control the limit of valid header size accepted on the backend side.

Monitob (approved):

Issue #1209, opened 2022-04-13
broken Load balancer - private network binding reported as healthy


Timost:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform Version
1.1.8

### Affected Resource(s)

- scaleway_lb -> private_network

### Terraform Configuration Files

```hcl
data "scaleway_vpc_private_network" "pn01" {
  name = "pn01"
}

data "scaleway_vpc_private_network" "pn02" {
  name = "pn02"
}

data "scaleway_lb_ip" "lb01_ip" {
  ip_id = var.lb01_ip_id
}

resource "scaleway_lb" "main_lb" {
  name  = "my_main_lb"
  ip_id = data.scaleway_lb_ip.lb01_ip.id
  zone  = var.zone_id
  type  = "LB-S"

  project_id = var.project_id
  tags       = ["network", "stuff"]

  private_network {
    private_network_id = data.scaleway_vpc_private_network.pn01.id
    dhcp_config        = true
    zone               = var.zone_id
  }

  private_network {
    private_network_id = data.scaleway_vpc_private_network.pn02.id
    dhcp_config        = true
    zone               = var.zone_id
  }
}
```

### Expected Behavior

Terraform shouldn't exit successfully if the private network binding wasn't successful.

### Actual Behavior

Terrform exists successfully but the symptoms
- health checks on the load balancer for resources inside the affected private network fail even if the endpoints are responding
- SSL certificates linked to the affected resources might fail in the browser

Luckily I cliked on the private network tab '^^

### Steps to Reproduce

1. `terraform apply`

### References

* #0000

PR #1259, opened 2022-05-04
fix(lb): check private network status

Monitob:
- Add check on load balancer private network attachement status
- Add context on missing queries
- Update Test
- Update Cassettes
- Update LB documentation
Related to #1209

### DIFF
--- a/scaleway/resource_lb.go
+++ b/scaleway/resource_lb.go
@@ -199,9 +199,7 @@ func resourceScalewayLbRead(ctx context.Context, d *schema.ResourceData, meta in
 	_ = d.Set("region", region.String())
 	_ = d.Set("organization_id", lb.OrganizationID)
 	_ = d.Set("project_id", lb.ProjectID)
-	if len(lb.Tags) > 0 {
-		_ = d.Set("tags", lb.Tags)
-	}
+	_ = d.Set("tags", lb.Tags)
 	// For now API return lowercase lb type. This should be fixed in a near future on the API side
 	_ = d.Set("type", strings.ToUpper(lb.Type))
 	_ = d.Set("ip_id", newZonedIDString(zone, lb.IP[0].ID))
--- a/scaleway/resource_lb.go
+++ b/scaleway/resource_lb.go
@@ -2,9 +2,11 @@ package scaleway
 
 import (
 	"context"
+	"fmt"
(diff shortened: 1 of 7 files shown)


scaleway/resource_lb_test.go
```diff
@@ -69,37 +69,6 @@ func TestAccScalewayLbLb_WithIP(t *testing.T) {
 		ProviderFactories: tt.ProviderFactories,
 		CheckDestroy:      testAccCheckScalewayLbDestroy(tt),
 		Steps: []resource.TestStep{
-			{
```
remyleone, line 72: Why is this step removed?

scaleway/resource_lb_test.go
```diff
@@ -295,6 +265,106 @@ func TestAccScalewayLbLb_WithIP(t *testing.T) {
 	})
 }
 
+func TestAccScalewayLbLb_WithMultiConfigs(t *testing.T) {
```
remyleone, line 268: WithSeveralPrivateNetworks ?

scaleway/resource_lb.go
```diff
@@ -201,24 +199,23 @@ func resourceScalewayLbRead(ctx context.Context, d *schema.ResourceData, meta in
 	_ = d.Set("region", region.String())
 	_ = d.Set("organization_id", lb.OrganizationID)
 	_ = d.Set("project_id", lb.ProjectID)
-	_ = d.Set("tags", lb.Tags)
+	if len(lb.Tags) > 0 {
```
remyleone, line 202: Why does this line need to be changed?

scaleway/resource_lb_test.go
Monitob: This test is now considered as an error. This configuration will always have an private network attachement state as an error.

scaleway/resource_lb.go
Monitob: This line is causing a diff after a resource creation because the attribute is `Optional` and the API is returning `[]`. As result the sdk is asking an extra `apply`.

remyleone: What happens in the case where the tags are deleted and now there are 0 tags in the list but the state is not yet updated?

remyleone (approved):

PR #1266, opened 2022-05-09
fix(rdb): add support for concurrent rdb bulk delete

remyleone:


Monitob:
PR blocked for issue on [sdk](https://github.com/hashicorp/terraform-plugin-sdk/issues/536)

PR #1270, opened 2022-05-11
fix(rdb-acl): set information on schema

Monitob:


scaleway/resource_rdb_acl.go
```diff
@@ -218,6 +222,31 @@ func rdbACLExpand(data []interface{}) ([]*rdb.ACLRuleRequest, error) {
 	return res, nil
 }
 
+func rdbACLRulesFlattenFromSchema(rules []*rdb.ACLRule, dataFromSchema []interface{}) []map[string]interface{} {
```
remyleone, line 225: This looks a lot like a merge operation. Should it be renamed to convey this idea?

scaleway/resource_rdb_acl.go
```diff
@@ -218,6 +222,31 @@ func rdbACLExpand(data []interface{}) ([]*rdb.ACLRuleRequest, error) {
...
+	}
+
+	for _, ruleFromSchema := range dataFromSchema {
+		currentRule := ruleFromSchema.(map[string]interface{})
+		ip, err := expandIPNet(currentRule["ip"].(string))
+		if err != nil {
+			return nil // append errors
+		}
+
+		aclRule := ruleMap[ip.String()]
```
remyleone, line 239: What happens in the case `aclRule` is nil here? (Zero value of the type of the values of `ruleMap`

scaleway/resource_rdb_acl.go
Monitob: Good to me

scaleway/resource_rdb_acl.go
Monitob: Does it mean that It can be added by the other channel like the CLI and then try to read the ACL rules? I should definitely test that

remyleone (approved):

PR #1275, opened 2022-05-16
fix(lb): add ForceNew to the schema of the certificate

remyleone:


scaleway/resource_lb_certificate.go
```diff
@@ -49,6 +49,7 @@ func resourceScalewayLbCertificate() *schema.Resource {
 						"common_name": {
 							Type:        schema.TypeString,
 							Required:    true,
+							ForceNew:    true,
```
Monitob, line 52: Please add an explicit note in the documentation.

Monitob (approved):

PR #1298, opened 2022-05-20
chore: fix staticcheck linter

remyleone:


remyleone (commented):
@Monitob it seems that rdb is using deprecated API regarding the endpoint

PR #1312, opened 2022-06-02
fix: PN delete operation and wait before consider a nic deleted

abarbare:
It seems, we could have a private network nic flaged as deleted but still present on VM. 
It's a problem in case of unplug/plug of private NIC with the same private network id as private nic is not correctly replaced in the VM


Monitob:
Why this is closed. It's going to be fixed on the API?

PR #1324, opened 2022-06-07
fix(k8s): skip oidc test

Codelax:


Monitob (approved):

scaleway/resource_k8s_cluster_test.go
```diff
@@ -241,6 +241,7 @@ func TestAccScalewayK8SCluster_Autoscaling(t *testing.T) {
 }
 
 func TestAccScalewayK8SCluster_OIDC(t *testing.T) {
+	t.Skip("Skip OIDC test, we need to set-up an OIDC server to test it")
```
remyleone, line 244: We can reuse one that is existing outside:

- https://developers.google.com/identity/protocols/oauth2/openid-connect

Codelax: We need an account to use an existing one

remyleone: We have plenty :) We can use opensource@scaleway.com for instance.

PR #1325, opened 2022-06-08
fix(k8s): fix waiting for multicloud cluster

remyleone:


scaleway/resource_k8s_cluster.go
```diff
@@ -385,8 +385,13 @@ func resourceScalewayK8SClusterCreate(ctx context.Context, d *schema.ResourceDat
 	}
 
 	d.SetId(newRegionalIDString(region, res.ID))
-
-	_, err = waitK8SClusterPool(ctx, k8sAPI, region, res.ID, d.Timeout(schema.TimeoutCreate))
+	if clusterType.(string) == "multicloud" {
+		// In case of multi-cloud, we do not have the guarantee that a pool will be created in Scaleway.
+		_, err = waitK8SCluster(ctx, k8sAPI, region, res.ID, d.Timeout(schema.TimeoutCreate))
```
Codelax, line 390: `waitK8SCluster` do the same thing as `waitK8SClusterPool` but allows more terminal status, why not using `waitK8SCluster` for standard and multicloud cluster

Codelax, line 390: `waitK8SCluster` do the same thing as `waitK8SClusterPool` but allows more terminal status, why not using `waitK8SCluster` for standard and multicloud cluster

Monitob: `waitK8SClusterPool`  will wait that the pool is created but when we threat `multicloud` the pools can be elsewhere.

scaleway/resource_k8s_cluster.go
```diff
@@ -385,8 +385,13 @@ func resourceScalewayK8SClusterCreate(ctx context.Context, d *schema.ResourceDat
 	}
 
 	d.SetId(newRegionalIDString(region, res.ID))
-
-	_, err = waitK8SClusterPool(ctx, k8sAPI, region, res.ID, d.Timeout(schema.TimeoutCreate))
+	if clusterType.(string) == "multicloud" {
```
Monitob, line 388: Can we set the types on the SDK? or as const values

scaleway/resource_k8s_cluster.go
Monitob: How can we handle the transient states on multicloud?

Codelax:
```go
func (s *API) WaitForClusterPool(req *WaitForClusterRequest, opts ...scw.RequestOption) (*Cluster, error) {
	cluster, err := async.WaitSync(&async.WaitSyncConfig{
		Get: func() (interface{}, bool, error) {
			cluster, err := s.GetCluster(&GetClusterRequest{
				ClusterID: req.ClusterID,
				Region:    req.Region,
			}, opts...)
			if err != nil {
				return nil, false, err
			}

			_, isTerminal := terminalStatus[cluster.Status]
			return cluster, isTerminal, nil
		},
		Timeout:          timeout,
		IntervalStrategy: async.LinearIntervalStrategy(retryInterval),
	})
}
```

`waitK8SClusterPool` will not wait for pool, it is the same as `waitK8SCluster` but with different terminal state

For `waitK8SClusterPool`
```go
	terminalStatus := map[ClusterStatus]struct{}{
		ClusterStatusPoolRequired: {},
	}
```

For `waitK8SCluster`
```go
	terminalStatus := map[ClusterStatus]struct{}{
		ClusterStatusReady:        {},
		ClusterStatusLocked:       {},
		ClusterStatusDeleted:      {},
		ClusterStatusPoolRequired: {},
	}
```

Monitob: Ok for the creation was necessary only check the pool creation. It's a special case. I will not be the case for multicloud

scaleway/resource_k8s_cluster.go
remyleone: Nope they are like DEV1-S. We don't enum them because they can change more often.

Monitob (approved):

PR #1326, opened 2022-06-09
fix(k8s): fix k8s oidc lists update and fix test

Codelax:


scaleway/helpers.go
```diff
@@ -394,6 +394,19 @@ func expandStringsPtr(data interface{}) *[]string {
 	return &stringSlice
 }
 
+// expandUpdatedStringsPtr expands a string slice but will default to an empty list
```
remyleone, line 397:
```suggestion
// expandUpdatedStringsPtr expands a string slice but will default to an empty list.
```

scaleway/helpers.go
```diff
@@ -394,6 +394,19 @@ func expandStringsPtr(data interface{}) *[]string {
...
+// should be used on schema update so emptying a list will update resource.
```
remyleone, line 398:
```suggestion
// Should be used on schema update so emptying a list will update resource.
```

remyleone (approved):

PR #1343, opened 2022-06-27
feat(iam): add iam api key resource

yfodil:


scaleway/resource_iam_api_key.go
```diff
@@ -0,0 +1,173 @@
...
+				Description: "The date and time of the creation of the iam api key",
+			},
+			"updated_at": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "The date and time of the last update of the iam api key",
+			},
+			"expires_at": {
+				Type:        schema.TypeString,
+				Computed:    true,
```
remyleone, line 40: Can it be set? What is the default value for the lifetime of an API key?

scaleway/resource_iam_api_key.go
```diff
@@ -0,0 +1,173 @@
...
+			},
+			"editable": {
+				Type:        schema.TypeBool,
+				Computed:    true,
+				Description: "Whether or not the iam api key is editable",
+			},
+			"creation_ip": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "The IP Address of the device which created the API key",
```
remyleone, line 75: Is it an IPv4 or an IPv6 address?

Issue #1337, opened 2022-06-21
Change priority of provider configuration methods


cryptobioz:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

According to the documentation, the provider can be configured _via_ 3 different methods, sorted by priority:
1. Environment variables
2. Static credentials
3. Shared configuration file

This priority order and usage of environment variables have two significant side effects.

#### Inability to use provider aliases

For example, if you want to deploy resources in multiple projects in the same Terraform workspace, you won't be able to make provider aliases work.
As the environment variables have the highest priority, all the resources configured to be deployed in different projects will be silently deployed in the project defined by the environment variable SCW_DEFAULT_PROJECT_ID.

#### Change of context

The second and the most serious impact of this priority order is the risk of a change of context. It's easier to control the code rather than the running environment. If we configured all our providers to use defined credentials and, by mistake or maliciously, someone runs a `terraform apply` with configured SCW_* environment variables, it would have a major impact on the deployed infrastructure.

#### A solution

This could be resolved quite easily changing the priority order to:

1. Static credentials
2. Environment variables
3. Shared configuration file

And can be achieved as easily by changing the arguments' positions here: https://github.com/scaleway/terraform-provider-scaleway/blob/master/scaleway/provider.go#L290

Would this change make sense to you?

PR #1350, opened 2022-06-29
add support for iam group resource

Mia-Cross:
I managed to keep things the way they worked before the breaking changes of the IAM API, all my tests with users. applications and both at the same time work fine without these attributes being required, I think it's more user-friendly this way.


scaleway/data_source_iam_group_test.go
```diff
@@ -0,0 +1,101 @@
...
+		Steps: []resource.TestStep{
+			{
+				Config: `
+					resource "scaleway_iam_group" "main_ds_basic" {
+						name        = "test_data_source_basic"
+					}
+			
+					data "scaleway_iam_group" "find_by_id_basic" {
+						group_id 	= scaleway_iam_group.main_ds_basic.id
+						organization_id = "08555df8-bb26-43bc-b749-1b98c5d02343"
```
remyleone, line 28: Can the organization_id be deduced from the provider settings? it seems cumbersome to specify it again

scaleway/data_source_iam_group_test.go
```diff
@@ -0,0 +1,101 @@
...
+				Config: `
+					resource "scaleway_iam_application" "app00" {
+						name = "app"
+					}
+					resource "scaleway_iam_group" "main_ds_mix" {
+						name = "test_data_source_mix"
+						application_ids = [
+							scaleway_iam_application.app00.id,
+						]
+						user_ids = [
```
remyleone, line 69: Can those user_ids be created using a terraform ressource to avoid having a hard-coded UUID?

scaleway/resource_iam_group.go
```diff
@@ -0,0 +1,221 @@
...
+	}
+
+	req := &iam.UpdateGroupRequest{
+		GroupID: group.ID,
+	}
+
+	if d.HasChange("name") {
+		req.Name = expandStringPtr(d.Get("name").(string))
+	} else {
+		req.Name = &group.Name
```
remyleone, line 146: Why should we have this else branch? Shouldn't we send only the thing we wish to modify?

scaleway/resource_iam_group.go
```diff
@@ -0,0 +1,221 @@
...
+	}
+
+	if d.HasChange("description") {
+		req.Description = expandStringPtr(d.Get("description").(string))
+	} else {
```
remyleone, line 151: This else branch should not exist in the first place.

scaleway/resource_iam_group_test.go
```diff
@@ -0,0 +1,485 @@
...
+		CheckDestroy: resource.ComposeTestCheckFunc(
+			testAccCheckScalewayIamGroupDestroy(tt),
+		),
+		Steps: []resource.TestStep{
+			{
+				Config: `
+						resource "scaleway_iam_group" "main_user" {
+							name = "iam_group_user"
+							user_ids = [
+								"29c31dd4-8ea1-4927-82d9-a0620e04773f"
```
remyleone, line 234: Can we have a data source about this? It would be nice to use something human-readable to understand who the user is.

scaleway/data_source_iam_group_test.go
Mia-Cross: For now it is not possible for requests that use query arguments like ListGroups, so it affects the sweeper functions and the search by name in datasources. After discussing it with the team, the solution we found was to temporarily make the organization_id required in datasources (every iam datasource is concerned).
Apparently, it's related to [this issue](https://github.com/scaleway/terraform-provider-scaleway/issues/1337).

scaleway/resource_iam_group.go
Mia-Cross: The Update request uses the PATCH method, not PUT so we have to send all information, even if we don't want to modify all of them. But indeed the way I wrote it could be improved so I've changed it a bit.

Mia-Cross:
The problem with the order of IDs has been resolved by using a TypeSet instead of a TypeList for the `application_ids` and `user_ids` attributes, so that we can check that the values are contained in the set without specifying the index. 
In the meantime, the AUTH team has handled the [issue](https://jira.infra.online.net/browse/AUTH-2127) I had opened about the description that could not be unset so every test now passes.

PR #1358, opened 2022-07-06
feat(serverless): add support for native domain name

remyleone:


docs/resources/container.md
```diff
@@ -84,6 +84,8 @@ The following arguments are optional:
 
 - `deploy` - (Optional) Boolean controlling whether the container is on a production environment.
 
+- `domain_name` - The native domain name of the container
```
Codelax, line 87: Only computed, should be in exported attributes

docs/resources/function.md
```diff
@@ -77,6 +77,8 @@ The following arguments are supported:
 
 - `zip_file` - Location of the zip file to upload containing your function sources
 
+- `domain_name` - The native domain name of the function
```
Codelax, line 80: Only computed, should be in exported attributes

Codelax (changes requested):

Codelax (approved):

PR #1388, opened 2022-07-22
fix: return nil for nil strings ptr

Codelax:


scaleway/helpers.go
```diff
@@ -391,6 +391,9 @@ func expandStringsPtr(data interface{}) *[]string {
 	for _, s := range data.([]interface{}) {
 		stringSlice = append(stringSlice, s.(string))
 	}
+	if stringSlice == nil {
+		return nil
```
kindermoumoute, line 395: return nil line 389 too

kindermoumoute (approved):
LGTM

PR #1396, opened 2022-07-28
fix(vpc-gw): bastion update

Monitob:


scaleway/resource_vpc_public_gateway_test.go
```diff
@@ -92,6 +92,54 @@ func TestAccScalewayVPCPublicGateway_Basic(t *testing.T) {
 	})
 }
 
+func TestAccScalewayVPCPublicGateway_Bastion(t *testing.T) {
+	tt := NewTestTools(t)
+	defer tt.Cleanup()
+	publicGatewayName := "public-gateway-test"
```
Codelax, line 98: Use different name for the test, it collides with the basic test

Codelax (changes requested):

Codelax (approved):

PR #1438, opened 2022-08-26
fix(rdb): acl flatten helper

Monitob:


scaleway/resource_rdb_acl.go
```diff
@@ -216,14 +216,22 @@ func rdbACLExpand(data []interface{}) ([]*rdb.ACLRuleRequest, error) {
 	var res []*rdb.ACLRuleRequest
 	for _, rule := range data {
 		r := rule.(map[string]interface{})
-		ip, err := expandIPNet(r["ip"].(string))
-		if err != nil {
-			return res, err
+
+		aclRule := &rdb.ACLRuleRequest{}
```
remyleone, line 220: Can we end up adding empty aclRule ? Is that normal? Or an acl with only a description and no ip

Monitob: Thank you for the feedback. I just updated the helper

remyleone (dismissed):

remyleone (approved):

PR #1555, opened 2022-10-27
tests(object): enable tests and handle s3 with cassettes

Codelax:


scaleway/validate_cassettes_test.go
```diff
@@ -53,6 +53,7 @@ func checkErrorCode(c *cassette.Cassette) error {
 func exceptionsCassettesCases() map[string]struct{} {
 	return map[string]struct{}{
 		"testdata/object-bucket-destroy-force.cassette.yaml":     {},
+		"testdata/object-bucket-cors-empty-origin.cassette.yaml": {},
```
remyleone, line 56: Do we have a way to document why there are exceptions in those cases?

Codelax: Look up the [related test](https://github.com/scaleway/terraform-provider-scaleway/blob/master/scaleway/resource_object_bucket_test.go#L578), it require an error

remyleone (approved):

PR #1570, opened 2022-11-03
feat(validate_cassette): check for all http errors except a list

Codelax:


scaleway/validate_cassettes_test.go
```diff
@@ -59,6 +59,25 @@ func exceptionsCassettesCases() map[string]struct{} {
...
+				return true
+			}
+		}
+		return false
+	}
+
+	return true
+}
+
 func checkErrCode(i *cassette.Interaction, c *cassette.Cassette, codes ...int) bool {
```
remyleone, line 81: Could you remove this function if it is not used anymore? or find another usage for it.

PR #1572, opened 2022-11-03
fix(k8s): add check for minor version with the autoupgrade option

Monitob:


scaleway/resource_k8s_cluster.go
```diff
@@ -222,6 +222,34 @@ func resourceScalewayK8SCluster() *schema.Resource {
...
+		CustomizeDiff: func(ctx context.Context, diff *schema.ResourceDiff, i interface{}) error {
+			autoUpgradeEnable, okAutoUpgradeEnable := diff.GetOkExists("auto_upgrade.0.enable")
+			_, okAutoUpgradeStartHour := diff.GetOkExists("auto_upgrade.0.maintenance_window_start_hour")
+			_, okAutoUpgradeDay := diff.GetOk("auto_upgrade.0.maintenance_window_day")
+
+			if okAutoUpgradeEnable {
+				// check if either all or none of the auto upgrade attribute are set.
+				// if one auto upgrade attribute is set, they all must be set.
+				// if none is set, auto upgrade attributes will be computed.
+				if !(okAutoUpgradeDay && okAutoUpgradeStartHour) {
```
Sh4d1, line 234: Hum the comment is not aligned with the code, the behaviour changed in https://github.com/scaleway/terraform-provider-scaleway/pull/730 
Here if both the `okAutoUpgradeDay` and the `okAutoUpgradeStartHour` are false, an error is raised. Guess it should also be modified in the apply section. a simple `if okAutoUpgradeDay != okAutoUpgradeStartHour` instead of this one should be enough (if that's what is wanted)

scaleway/resource_k8s_cluster.go
```diff
@@ -222,6 +222,34 @@ func resourceScalewayK8SCluster() *schema.Resource {
...
+			}
+
+			version := diff.Get("version").(string)
+			versionIsOnlyMinor := len(strings.Split(version, ".")) == 2
+
+			clusterAutoUpgradeEnabled := false
+
+			if okAutoUpgradeDay && okAutoUpgradeEnable && okAutoUpgradeStartHour {
+				clusterAutoUpgradeEnabled = autoUpgradeEnable.(bool)
+				if versionIsOnlyMinor != clusterAutoUpgradeEnabled {
```
Sh4d1, line 246: why not just `if versionIsOnlyMinor != autoUpgradeEnable.(bool)` ?

scaleway/resource_k8s_cluster.go
```diff
@@ -222,6 +222,34 @@ func resourceScalewayK8SCluster() *schema.Resource {
...
+					return fmt.Errorf("all field or zero field of auto_upgrade must be set")
+				}
+			}
+
+			version := diff.Get("version").(string)
+			versionIsOnlyMinor := len(strings.Split(version, ".")) == 2
+
+			clusterAutoUpgradeEnabled := false
+
+			if okAutoUpgradeDay && okAutoUpgradeEnable && okAutoUpgradeStartHour {
```
Sh4d1, line 244: This if could be mutualized with the one above, something like 
```go
if autoUpEnable {
    if !condition {
         return err
    }
    // do other checks
}
```

scaleway/resource_k8s_cluster.go
Monitob: ok, I've just removed it because it is already. checked on the schema

scaleway/resource_k8s_cluster.go
Monitob: idem, code removed

scaleway/resource_k8s_cluster.go
Monitob: Thanks, I simplified as you suggested

Sh4d1 (approved):
LG

scaleway/resource_k8s_cluster.go
```diff
@@ -222,6 +222,20 @@ func resourceScalewayK8SCluster() *schema.Resource {
 				Description: "The status of the cluster",
 			},
 		},
+		CustomizeDiff: func(ctx context.Context, diff *schema.ResourceDiff, i interface{}) error {
+			autoUpgradeEnable, okAutoUpgradeEnable := diff.GetOkExists("auto_upgrade.0.enable")
+
+			version := diff.Get("version").(string)
+			versionIsOnlyMinor := len(strings.Split(version, ".")) == 2
+
+			if okAutoUpgradeEnable {
```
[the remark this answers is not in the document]
Monitob: Thanks.  It's resolved

PR #1605, opened 2022-11-15
fix(container): set secret environment variables on creation

Codelax:


scaleway/helpers_container.go
```diff
@@ -62,6 +62,10 @@ func setCreateContainerRequest(d *schema.ResourceData, region scw.Region) (*cont
 		req.EnvironmentVariables = expandMapPtrStringString(envVariablesRaw)
 	}
 
+	if secretEnvVariablesRaw, ok := d.GetOk("secret_environment_variables"); ok {
```
Monitob, line 65: Can we add this to our example, please?

remyleone (approved):

PR #1619, opened 2022-11-21
Fix(vpcgw): add zone option to  datasource vpcgw

### LINKED ISSUE #1610: Allow zone argument for VPC Public Gateway data source
Currently, the `scaleway_vpc_public_gateway` data source does not allow passing in `zone` as an argument. Instead, it's parsed from the provider configuration context. See here:

https://github.com/scaleway/terraform-provider-scaleway/blob/59b339598a479e450aee27b5c94533b6b7617f05/scaleway/data_source_vpc_public_gateway.go#L36

This however, makes the data source quite difficult to use if you have VPC Public Gateways defined across multiple zones. Current work-around would be to define separate Scaleway providers with aliases and pass them to the data source as a [meta-argument](https://developer.hashicorp.com/terraform/language/meta-arguments/resource-provider).

Potential solution: allow users to pass the `zone` via an argument. Relevant code block:

https://github.com/scaleway/terraform-provider-scaleway/blob/59b339598a479e450aee27b5c94533b6b7617f05/scaleway/data_source_vpc_public_gateway.go#L13-L33

Monitob:
Close #1610

### DIFF
--- a/scaleway/data_source_vpc_public_gateway.go
+++ b/scaleway/data_source_vpc_public_gateway.go
@@ -6,7 +6,7 @@ import (
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
-	vpcgw "github.com/scaleway/scaleway-sdk-go/api/vpcgw/v1"
+	"github.com/scaleway/scaleway-sdk-go/api/vpcgw/v1"
 	"github.com/scaleway/scaleway-sdk-go/scw"
 )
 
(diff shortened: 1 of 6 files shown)


scaleway/data_source_vpc_public_gateway.go
```diff
@@ -15,7 +15,7 @@ func dataSourceScalewayVPCPublicGateway() *schema.Resource {
 	dsSchema := datasourceSchemaFromResourceSchema(resourceScalewayVPCPublicGateway().Schema)
 
 	// Set 'Optional' schema elements
-	addOptionalFieldsToSchema(dsSchema, "name")
+	addOptionalFieldsToSchema(dsSchema, "name", "zone")
```
remyleone, line 18: What is the purpose of this line? zone is already in the schema of `resourceScalewayVPCPublicGateway `

Monitob: Our data-source helper, `zone` remove the optional flag. We have to precise it if we want to keep it optional. Otherwise we got this error : ```Value for unconfigurable attribute```

remyleone (approved):

PR #1645, opened 2022-12-05
fix(iam): organization_id as optional and computed on iam_application

Monitob:


Codelax (changes requested):

scaleway/resource_iam_application.go
```diff
@@ -46,16 +46,22 @@ func resourceScalewayIamApplication() *schema.Resource {
 				Computed:    true,
 				Description: "Whether or not the application is editable.",
 			},
-			"organization_id": organizationIDSchema(),
+			"organization_id": {
```
kindermoumoute, line 49: Maybe could use a `organizationIDOptionalSchema` helper?

Codelax (approved):

PR #1671, opened 2023-01-02
fix(lb): frontend inherit zone from related lb

matth-:
a Terraform script with default zone different from Lb one, will fail to create the Lb frontend on the correct zone, leading to a error 400.

This PR set the frontend resource zone according to the related lb. Also add some zone related check.


Codelax:
Hi thanks for your contribution.

The issue with zoned and regional IDs is that we cannot enforce usage of this format.
The following snippets should work:

```hcl
resource "scaleway_lb_ip" "main" {
  zone = "fr-par-1"
}

resource "scaleway_lb" "base" {
  ip_id  = scaleway_lb_ip.main.id
  zone   = scaleway_lb_ip.main.zone
  type   = "LB-S"
}
```
and
```hcl
resource "scaleway_lb" "base" {
  ip_id  = "xxxx-xxxx-xxxx-xxxx"
  type   = "LB-S"
}
```

User is able to use a locality ID aswell as a bare ID.
Also, as we can see in [lb documentation](https://registry.terraform.io/providers/scaleway/scaleway/latest/docs/resources/lb) basic's example (which is the first snippet before), the recommended way is to pass down locality from one resource to another if you want to guarantee they are in the same one.

This is still a great issue that you have highlighted, we should provide more errors when validating a config.

matth-:
i see, then it needs to add a zone parameter to every lb ressource as it misse for frontend, backend and route.

Codelax:
Ok, thanks for this, I will make a pass on it.
It should be ok in the next release, sorry for the inconvenience.

Codelax (approved):

PR #1687, opened 2023-01-11
doc: update documentation to only use privateNICs

yfodil:


docs/resources/lb.md
```diff
@@ -63,10 +63,11 @@ resource "scaleway_instance_server" "main" {
     type        = "DEV1-S"
     image       = "debian_bullseye"
     enable_ipv6 = false
+}
 
-    private_network {
-        pn_id = scaleway_vpc_private_network.main.id
-    }
+resource "scaleway_instance_private_nic" "pnic01" {
```
Monitob, line 68: This is not the only example to update. Try to run also all the examples in local.

scaleway/resource_lb_test.go
```diff
@@ -479,10 +479,11 @@ func TestAccScalewayLbLb_WithPrivateNetworksOnDHCPConfig(t *testing.T) {
 					type        = "DEV1-S"
 					image       = "debian_bullseye"
 					enable_ipv6 = false
-				
-					private_network {
```
Monitob, line 483: Please check that waiting process is still working? may we just use the wait from private_nic now

docs/resources/lb.md
```diff
@@ -63,10 +63,11 @@ resource "scaleway_instance_server" "main" {
     type        = "DEV1-S"
     image       = "debian_bullseye"
     enable_ipv6 = false
+}
 
-    private_network {
```
Monitob, line 67: Aldo can you explain on the doc, Why we should be using instance_private_nic instead?

Monitob (changes requested):

PR #1782, opened 2023-02-20
feat(lb): add lbs datasource

yfodil:


docs/data-sources/lbs.md
```diff
@@ -0,0 +1,55 @@
...
+# scaleway_lbs
+
+Gets information about multiple Load Balancers.
+
+## Example Usage
+
+```hcl
+# Find LBs by name
+data "scaleway_lbs" "my_key" {
+  name = "foobar"
```
remyleone, line 16: Please add support for tags, it will help to select several load balancers in one go.

scaleway/data_source_lbs.go
```diff
@@ -0,0 +1,197 @@
...
+						},
+						"name": {
+							Computed: true,
+							Type:     schema.TypeString,
+						},
+						"type": {
+							Computed: true,
+							Type:     schema.TypeString,
+						},
+						"tags": {
```
remyleone, line 46: We should have a way to have selection through tags

docs/data-sources/lbs.md
yfodil: Tags support will be added afterward when its featured on api side

remyleone (approved):

PR #1808, opened 2023-02-27
fix(dns): API returns 403 instead of 404 on missing record

ndrpnt:
Whenever a record that doesn't exist is still present in Terraform state, we are stuck with the read failing because of a 403 error. 403 is used to indicate both that the record doesn't exist or that it exist be we are not allowed to access it.

This happens during normal operations because the deletion of a DNS record regularly fails with another error (that I have yet to investigate).

### DIFF
--- a/scaleway/resource_domain_record.go
+++ b/scaleway/resource_domain_record.go
@@ -321,7 +321,7 @@ func resourceScalewayDomainRecordRead(ctx context.Context, d *schema.ResourceDat
 			ID:      &recordID,
 		}, scw.WithAllPages(), scw.WithContext(ctx))
 		if err != nil {
-			if is404Error(err) {
+			if is404Error(err) || is403Error(err) {
 				d.SetId("")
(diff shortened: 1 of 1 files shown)


Codelax (approved):
Thank you for your contribution, this is a miss on our side. 

If domain or dns_zone does not exist, it will fail on read
```
╷
│ Error: scaleway-sdk-go: http error 403 Forbidden: subdomain not found
│ 
│   with scaleway_domain_record.record,
│   on main.tf line 13, in resource "scaleway_domain_record" "record":
│   13: resource "scaleway_domain_record" "record" {
│ 
╵
```

To reproduce:

```hcl
resource "scaleway_domain_record" "record" {
  dns_zone = "subdomain.mydomain.com"
  data     = "1.2.3.4"
  name     = "test"
  type     = "A"
}
```

Delete dns_zone `subdomain.mydomain.com` on console, then refresh state

PR #1815, opened 2023-02-28
feat(secret-version): resource secret version

Monitob:


scaleway/resource_secret_version.go
```diff
@@ -0,0 +1,204 @@
...
+
+func resourceScalewaySecretVersionCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	api, region, err := secretAPIWithRegion(d, meta)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	secretID := expandID(d.Get("secret_id").(string))
+	payloadSecret := []byte(d.Get("data").(string))
+	if isBase64Encoded(payloadSecret) {
```
kindermoumoute, line 89: proposal: return an error when it's not base64 and only manipulate base64 for this attribute

scaleway/helpers.go
```diff
@@ -696,6 +697,19 @@ func diffSuppressFuncTimeRFC3339(k, oldValue, newValue string, d *schema.Resourc
...
+	if oldValue == newValue {
+		return true
+	}
+
+	oldStr, err1 := base64.StdEncoding.DecodeString(oldValue)
+	newStr, err2 := base64.StdEncoding.DecodeString(newValue)
+	if err1 != nil || err2 != nil {
+		return false
+	}
+	return string(oldStr) == string(newStr)
```
quantumsheep, line 710:
```suggestion
	return oldValue == newValue
```

Two encoding of the same value outputs the same base64, why do we need the decoding part?

scaleway/helpers.go
```diff
@@ -696,6 +697,19 @@ func diffSuppressFuncTimeRFC3339(k, oldValue, newValue string, d *schema.Resourc
 	return t1.Equal(t2)
 }
 
+func diffSuppressFuncBase64(k string, oldValue string, newValue string, d *schema.ResourceData) bool {
```
quantumsheep, line 700:
```suggestion
func diffSuppressFuncBase64(oldValue string, newValue string) bool {
```

quantumsheep (changes requested):

scaleway/resource_secret_version.go
```diff
@@ -0,0 +1,204 @@
...
+		Schema: map[string]*schema.Schema{
+			"secret_id": {
+				Type:             schema.TypeString,
+				Required:         true,
+				Description:      "The secret ID associated with this version",
+				DiffSuppressFunc: diffSuppressFuncLocality,
+			},
+			"data": {
+				Type:             schema.TypeString,
+				Optional:         true,
```
kindermoumoute, line 36: Let's keep this required and recommand using datasources

scaleway/resource_secret_version.go
```diff
@@ -0,0 +1,204 @@
...
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "Status of the secret version",
+			},
+			"revision": {
+				Type:        schema.TypeInt,
+				Computed:    true,
+				Description: "The revision of secret version",
+			},
+			"with_access": {
```
kindermoumoute, line 60: can be removed in the resource since only datasource will actually access the resource

remyleone:
Why is it in a different resource ? could it be in the same ? I think the DX would be better

remyleone (approved):

scaleway/resource_secret_version.go
```diff
@@ -0,0 +1,189 @@
...
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	secretID := expandID(d.Get("secret_id").(string))
+	payloadSecretRaw := []byte(d.Get("data").(string))
+	if !isBase64Encoded(payloadSecretRaw) {
+		return diag.FromErr(fmt.Errorf("data payload mush be base64 encoded"))
+	}
+	payloadSecret, err := base64.StdEncoding.DecodeString(d.Get("data").(string))
```
Monitob, line 86: Warning double encode from SDK.

scaleway/resource_secret_version.go
```diff
@@ -0,0 +1,189 @@
...
+				Type:             schema.TypeString,
+				Required:         true,
+				Description:      "The data payload of your secret version on 64 mode",
+				Sensitive:        true,
+				ForceNew:         true,
+				DiffSuppressFunc: diffSuppressFuncBase64,
+				StateFunc: func(i interface{}) string {
+					return base64Encoded([]byte(i.(string)))
+				},
+			},
```
Codelax, line 44: The current issue with getting a base64 encoded value from terraform is that SDK will base64 the given data anyway, leading to a double base64 encoded data if we do not decode it before passing to SDK.
This make use requesting a base64 encoded data to decode it then encode it again...

Monitob: I've just removed the base64 restriction on the create  and only kept it on the state

docs/resources/secret_version.md
```diff
@@ -0,0 +1,68 @@
...
+
+- `secret_id` - (Required) The Secret ID associated wit the secret version.
+- `data` - (Optional) The data payload of the secret version. Must be no larger than 64KiB. (e.g. `my-secret-version-payload`). more on the [data section](#data)
+- `description` - (Optional) Description of the secret version (e.g. `my-new-description`).
+- `region` - (Defaults to [provider](../index.md#region) `region`) The [region](../guides/regions_and_zones.md#regions)
+  in which the resource exists.
+
+## Data
+
+Note: The `data` is already handle on the SDK, avoid double encoded.
```
quantumsheep, line 42:
```suggestion
Note: The `data` should be a base64 encoded string when sent from the API. **It is already handled by the provider so you don't need to encode it yourself.**
```

docs/resources/secret_version.md
```diff
@@ -0,0 +1,68 @@
...
+
+Updates to `data` will force new the secret version.
```
quantumsheep, line 44:
```suggestion
Updating `data` will force creating a new the secret version.
```

docs/resources/secret_version.md
```diff
@@ -0,0 +1,68 @@
...
+- `revision` - The revision for this Secret Version.
+- `status` - The status of the Secret Version.
+- `created_at` - Date and time of secret version's creation (RFC 3339 format).
+- `updated_at` - Date and time of secret version's last update (RFC 3339 format).
+
+## Import
+
+The Secret Version can be imported using the `{region}/{id}/{revision}`, e.g.
+
+~> **Important:** Be aware if you import with revision `latest` you will overwrite your version if you used before.
```
quantumsheep, line 64:
```suggestion
~> **Important:** Be aware if you import with revision `latest` you will overwrite the version you used before.
```

scaleway/helpers_secret.go
```diff
@@ -35,3 +36,27 @@ func secretAPIWithRegionAndID(m interface{}, id string) (*secret.API, scw.Region
...
+// secretVersionAPIWithRegionAndID returns a Secret API with locality and Nested ID extracted from the state
+func secretVersionAPIWithRegionAndID(m interface{}, id string) (*secret.API, scw.Region, string, string, error) {
+	meta := m.(*Meta)
+	api := secret.NewAPI(meta.scwClient)
+
+	region, id, revision, err := parseLocalizedNestedID(id)
+	if err != nil {
+		return nil, "", "", "", err
+	}
+	return api, scw.Region(region), id, revision, nil
```
quantumsheep, line 49:
```suggestion

	region, id, revision, err := parseLocalizedNestedID(id)
	if err != nil {
		return nil, "", "", "", err
	}

	api := secret.NewAPI(meta.scwClient)
	return api, scw.Region(region), id, revision, nil
```

Avoid allocating memory if it may be not used

quantumsheep (changes requested):

quantumsheep (changes requested):

quantumsheep (approved):

PR #1835, opened 2023-03-13
fix(rdb): replace ipnet as optional

Monitob:


scaleway/resource_instance_security_group.go
```diff
@@ -317,7 +317,7 @@ func updateSecurityGroupeRules(ctx context.Context, d *schema.ResourceData, zone
 
 			// This happens when there is more rule in state than in the api. We create more rule in API.
 			setGroupRules = append(setGroupRules, &instance.SetSecurityGroupRulesRequestRule{
-				Zone:         zone,
+				Zone:         &zone,
```
remyleone, line 320: how is this change related? Please do two PR if the two subjects are different.

remyleone (changes requested):

scaleway/resource_instance_security_group.go
Monitob: Please find the PR related [here](https://github.com/scaleway/terraform-provider-scaleway/pull/1838)

remyleone (approved):

PR #1857, opened 2023-03-23
feat(function_domain): resolve domain before creating it

Codelax:


quantumsheep (approved):

scaleway/resource_function_domain.go
```diff
@@ -63,10 +64,16 @@ func resourceScalewayFunctionDomainCreate(ctx context.Context, d *schema.Resourc
 		return diag.FromErr(err)
 	}
 
+	hostname := d.Get("hostname").(string)
+
+	if !isFunctionDomainResolved(ctx, api, hostname, d.Timeout(schema.TimeoutCreate), functionID, region) {
```
quantumsheep, line 69: You can pass the returned function from `waitForFunction` so you don't do another get request later on

quantumsheep (changes requested):

PR #1862, opened 2023-03-29
Fix scaleway_rdb_instance using name as a pattern instead of an exact match

jvasseur:
When trying to import a RDB instance, the name parameter is directly passed to the API which treat it as a pattern instead of an exact match leading to a multiple instance with the name error if other instance contains the name of the searched instance.

This PR adds an additional step of matching to ensure only exact match are returned.

The code is heavily inspired by what is donne inside the `scaleway_instance_server` data source.


remyleone:
@jvasseur Thanks a lot for your contribution

Issue #1882, opened 2023-04-06
Missing attributes on scaleway_tem_domain


Nainterceptor:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

As a devOps engineer, I want to be able to hardcode the strict minimum of things, and automate the maximum of steps.
I have to, currently, hardcode SMTP host, username and port.
I have to, currently,  hardcode DKIM key (content is available).
I have to, currently, validate configuration step on console.
(Password will be fetch from IAM ressource creation)

### New or Affected Resource(s)

* scaleway_tem_domain

### Potential Terraform Configuration

No idea to pass steps, but host, username, ... can be added as attributes

### References

No

quantumsheep:
Hello! I'm not sure to understand how you want to use the resource, can you give me an example?

About the validation step we have #1733 which will be handled soon.

Currently you should be able to create your infrastructure like this:
```hcl

resource "scaleway_tem_domain" "main" {
  name       = "example.com"
  accept_tos = true
}

resource "scaleway_domain_record" "dkim" {
  dns_zone = scaleway_tem_domain.main.name
  name     = "dkim._domainkey"
  type     = "TXT"
  data     = scaleway_tem_domain.main.dkim_config
  ttl      = 3600
}

resource "scaleway_domain_record" "spf" {
  dns_zone = scaleway_tem_domain.main.name
  name     = "@"
  type     = "TXT"
  data     = "v=spf1 ${scaleway_tem_domain.main.spf_config} -all"
  ttl      = 3600
}

resource "scaleway_domain_record" "mx" {
  dns_zone = scaleway_tem_domain.main.name
  name     = "@"
  type     = "MX"
  data     = "."
  ttl      = 3600
  priority = 0
}
```

Nainterceptor:
Hi, that's ok to create resources, but then what ?
I want to inject tem information into CI/CD variables, or kubernetes secrets, or something else to let an application send email via tem.

In the following snippet, I want to get information, basically going to `gitlab_project_variable` from scaleway resources

```hcl
resource "scaleway_tem_domain" "main" {
[...]

PR #1892, opened 2023-04-11
feat(tem): add smtp host and ports in resource state

quantumsheep:
Relates to #1882


scaleway/resource_tem_domain.go
```diff
@@ -140,6 +170,12 @@ func resourceScalewayTemDomainRead(ctx context.Context, d *schema.ResourceData,
 	_ = d.Set("last_error", domain.LastError)
 	_ = d.Set("spf_config", domain.SpfConfig)
 	_ = d.Set("dkim_config", domain.DkimConfig)
+	_ = d.Set("smtp_host", "smtp.tem.scw.cloud")
```
remyleone, line 173: Why is it a static string? Could it be read from the API instead of being hardcoded?

scaleway/resource_tem_domain.go
```diff
@@ -140,6 +170,12 @@ func resourceScalewayTemDomainRead(ctx context.Context, d *schema.ResourceData,
...
+	_ = d.Set("smtp_port_unsecure", 25)
```
remyleone, line 174: Anyhow, if the product API cannot return those constants, they should live in the SDK as they are constant regardless of the devtools using them.

scaleway/resource_tem_domain.go
quantumsheep: There is no API to fetch this value. Same for ports.

remyleone (approved):

PR #1988, opened 2023-06-14
feat(lb): health_check block

Monitob:


scaleway/resource_lb_backend_test.go
```diff
@@ -105,6 +105,140 @@ func TestAccScalewayLbBackend_Basic(t *testing.T) {
...
+					
+					resource "scaleway_lb_backend" "bkd01" {
+					  lb_id            = scaleway_lb.lb01.id
+					  name             = "bkd01"
+					  forward_protocol = "tcp"
+					  forward_port     = 80
+					
+					  health_check {
+						protocol      = "mysql"
+						database_user = "devtools"
```
remyleone, line 228: Should a password be defined here?

scaleway/resource_lb_backend_test.go
```diff
@@ -105,6 +105,140 @@ func TestAccScalewayLbBackend_Basic(t *testing.T) {
...
+					  }
+					}
+					`,
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("scaleway_lb_backend.bkd01", "health_check.#", "1"),
```
remyleone, line 233: Check that the health check in created

remyleone (commented):
I think that this was started without a lot of agreement with the product team. Could we have a check with them and with @yfodil too?

yfodil (approved):

PR #1989, opened 2023-06-14
feat(k8s): Kapsule VPCs go from zoned to regional

Mia-Cross:
The private networks used with Kapsule clusters are now regional (when they were zoned before).

### DIFF
--- a/scaleway/resource_k8s_cluster.go
+++ b/scaleway/resource_k8s_cluster.go
@@ -245,6 +245,11 @@ func resourceScalewayK8SCluster() *schema.Resource {
 			},
 			func(ctx context.Context, diff *schema.ResourceDiff, i interface{}) error {
 				if diff.HasChange("private_network_id") {
+					clusterType := diff.Get("type")
+					isKapsule := clusterType == "" || strings.HasPrefix(clusterType.(string), "kapsule")
+					if !isKapsule {
+						return fmt.Errorf("only Kapsule clusters support private networks")
+					}
 					actual, planned := diff.GetChange("private_network_id")
 					if actual == "" {
 						// If no private network has been set yet, migrate the cluster in the Update function
@@ -437,7 +442,7 @@ func resourceScalewayK8SClusterCreate(ctx context.Context, d *schema.ResourceDat
 	// Private network configuration
 
 	if pnID, ok := d.GetOk("private_network_id"); ok {
-		req.PrivateNetworkID = scw.StringPtr(expandZonedID(pnID.(string)).ID)
+		req.PrivateNetworkID = scw.StringPtr(expandRegionalID(pnID.(string)).ID)
 	}
 
 	// Cluster creation
--- a/scaleway/resource_k8s_cluster_test.go
+++ b/scaleway/resource_k8s_cluster_test.go
@@ -2,6 +2,7 @@ package scaleway
 
 import (
 	"fmt"
+	"strings"
 	"testing"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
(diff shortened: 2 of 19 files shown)


scaleway/resource_k8s_cluster.go
```diff
@@ -229,6 +229,34 @@ func resourceScalewayK8SCluster() *schema.Resource {
 				Computed:    true,
 				Description: "The status of the cluster",
 			},
+			"private_network": {
+				Type:        schema.TypeList,
+				Computed:    true,
+				Optional:    true,
```
Codelax, line 235: Is it optional ? It means the user can set this attribute.

scaleway/resource_k8s_cluster.go
```diff
@@ -436,8 +464,12 @@ func resourceScalewayK8SClusterCreate(ctx context.Context, d *schema.ResourceDat
...
 	// Private network configuration
 
+	isKapsule := clusterType == "" || strings.HasPrefix(clusterType.(string), "kapsule")
 	if pnID, ok := d.GetOk("private_network_id"); ok {
-		req.PrivateNetworkID = scw.StringPtr(expandZonedID(pnID.(string)).ID)
+		if !isKapsule {
+			return diag.FromErr(fmt.Errorf("only Kapsule clusters support private networks"))
+		}
+		req.PrivateNetworkID = scw.StringPtr(expandRegionalID(pnID.(string)).ID)
 	}
```
Codelax, line 473: Can this check be achieved in `CustomizeDiff` ?
You can also return a full diagnostic to improve error output
```go
diag.Diagnostic{
	Severity:      diag.Error,
	Summary:       "Private networks are not supported for current cluster type",
	Detail:        "Cluster type must be kapsule to use a private network",
	AttributePath: cty.GetAttrPath("private_network_id"),
}
```

scaleway/resource_k8s_cluster.go
```diff
@@ -229,6 +229,32 @@ func resourceScalewayK8SCluster() *schema.Resource {
 				Computed:    true,
 				Description: "The status of the cluster",
 			},
+			"private_network": {
```
Codelax, line 232: Is that needed rather than using private_network [datasource](https://registry.terraform.io/providers/scaleway/scaleway/latest/docs/data-sources/vpc_private_network) ?

scaleway/resource_k8s_pool_test.go
```diff
@@ -460,7 +462,7 @@ func testAccCheckScalewayK8SPoolServersAreInPrivateNetwork(tt *TestTools, cluste
 			}
 
 			server, err := instanceAPI.GetServer(&instance.GetServerRequest{
-				Zone:     zone,
+				Zone:     scw.Zone(providerIDSplit[3]),
```
Codelax, line 465: Could you add a comment to give an example of the providerID format ?

Mia-Cross: Done -> line 460

scaleway/resource_k8s_cluster_test.go
```diff
@@ -664,6 +664,7 @@ func testAccCheckScalewayK8SClusterConfigPrivateNetworkNotLinked(version string)
 	return fmt.Sprintf(`
 resource "scaleway_vpc_private_network" "private_network" {
   name       = "k8s-private-network"
+  is_regional = true
```
remyleone, line 667: I don't think this field will stay once vpc v1 is deprecated. Is there a way to manage it transparently for the customer?

Mia-Cross: It is transparent for the customer because both v1 and v2 work for the moment, so they can choose if they want to add `is_regional = true` to use v2 or keep using v1.
The goal of this PR is to set up everything for v2 while still supporting v1, that's why I updated the test so that line 536 uses the regional function rather than the zoned one.

remyleone (approved):

PR #1995, opened 2023-06-15
feat(vpc): add support for regional pn on datasource

yfodil:


scaleway/data_source_vpc_private_network.go
```diff
@@ -25,6 +26,11 @@ func dataSourceScalewayVPCPrivateNetwork() *schema.Resource {
 		ValidateFunc:  validationUUIDorUUIDWithLocality(),
 		ConflictsWith: []string{"name"},
 	}
+	dsSchema["is_regional"] = &schema.Schema{
+		Type:        schema.TypeBool,
+		Optional:    true,
+		Description: "Whether this is a regional or zonal private network",
+	}
 
 	return &schema.Resource{
```
Codelax, line 35: Outside of current code change but `zone` and `region` cannot be configured for this datasource.
Those fields should be added to `addOptionalFieldsToSchema`

scaleway/data_source_vpc_private_network.go
```diff
@@ -71,5 +84,47 @@ func dataSourceScalewayVPCPrivateNetworkRead(ctx context.Context, d *schema.Reso
...
+				),
+			)
+		}
+		privateNetworkID = res.PrivateNetworks[0].ID
+	}
+
+	regionalID := datasourceNewRegionalizedID(privateNetworkID, region)
+	d.SetId(regionalID)
+	_ = d.Set("private_network_id", regionalID)
+	return resourceScalewayVPCPrivateNetworkRegionalRead(ctx, d, meta)
```
Codelax, line 129: Data source does not fail if given `private_network_id` does not exist

```suggestion
	diags := resourceScalewayVPCPrivateNetworkRegionalRead(ctx, d, meta)
	if diags != nil {
		        return append(diags, diag.Errorf("failed to read private network state")...)
	}
	
	if d.Id() == nil {
		return diag.Errorf("private network (%s) not found", regionalID)
	}
	
	return nil
```

docs/data-sources/vpc_private_network.md
```diff
@@ -21,17 +23,30 @@ data "scaleway_vpc_private_network" "my_id" {
 }
 ```
 
+### Regional
+
+```hcl
+# Get info by name
+data "scaleway_vpc_private_network" "my_name" {
+  name = "foobar"
+  is_regional = true
```
remyleone, line 32: Is there a way for us to guess it? I'm concerned because going forward we won't have zonal vpc anymore. therefore this field got good chance to be deprecated. In case there is an ambiguous PN that got both a version in zonal and regional maybe we should have a warning and call for the user to rename it?

PR #2018, opened 2023-06-22
feat(k8s): skip reading kubeconfig if unauthorized

### LINKED ISSUE #2015: Skip reading kubeconfig if not authorized instead of returning the error right away
### Description

Reading the kubeconfig when reading the cluster could result in a 403. In that case we should skip it and display a warning rather than returning an error.

* scaleway_k8s_cluster

Mia-Cross:
Closes #2015

### DIFF
--- a/scaleway/resource_k8s_cluster.go
+++ b/scaleway/resource_k8s_cluster.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/hashicorp/go-cty/cty"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/customdiff"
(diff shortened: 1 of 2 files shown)


scaleway/resource_k8s_cluster.go
```diff
@@ -521,38 +522,23 @@ func resourceScalewayK8SClusterRead(ctx context.Context, d *schema.ResourceData,
...
+	kubeconfig, err := flattenKubeconfig(ctx, k8sAPI, region, clusterID)
 	if err != nil {
-		return diag.FromErr(err)
-	}
-
-	kubeconfigToken, err := kubeconfig.GetToken()
-	if err != nil {
-		return diag.FromErr(err)
+		if !is403Error(err) {
+			return diag.FromErr(err)
```
remyleone, line 529: I would flip this conditionnal, add the diag in it and return any other error.

remyleone (approved):

PR #2024, opened 2023-06-26
feat(lb): add support for configuring backend servers max connections

yfodil:


scaleway/resource_lb_backend.go
```diff
@@ -274,6 +275,17 @@ E.g. 'failover-website.s3-website.fr-par.scw.cloud' if your bucket website URL i
 				Optional:    true,
 				Default:     false,
 			},
+			"max_connections": {
+				Type:        schema.TypeInt,
+				Optional:    true,
+				Description: "Maximum number of connections allowed per backend server",
```
remyleone, line 281: Is there a max for this max connection? If that is the case, could we have a validator here?

remyleone (approved):

PR #2067, opened 2023-07-25
feat(iam): add iam_group_membership resource

Codelax:


scaleway/resource_iam_group.go
```diff
@@ -139,7 +148,7 @@ func resourceScalewayIamGroupUpdate(ctx context.Context, d *schema.ResourceData,
 		}
 	}
 
-	if d.HasChanges("application_ids", "user_ids") {
+	if !d.Get("external_membership").(bool) && d.HasChanges("application_ids", "user_ids") {
```
quantumsheep, line 151: If we changes from `external_membership = false` to `external_membership = true`, shouldn't the previously set apps and users be deleted?

scaleway/resource_iam_group_membership.go
```diff
@@ -0,0 +1,162 @@
...
+	groupID, userID, applicationID, err := expandGroupMembershipID(d.Id())
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	group, err := api.GetGroup(&iam.GetGroupRequest{
+		GroupID: groupID,
+	}, scw.WithContext(ctx))
+	if err != nil {
+		return diag.FromErr(err)
```
quantumsheep, line 79:
```suggestion
		if is404Error(err) {
			d.SetId("")
			return nil
		}
		return diag.FromErr(err)
```

If the group doesn't exists, neither do the membership

scaleway/resource_iam_group_membership.go
```diff
@@ -0,0 +1,162 @@
...
+	if userID != "" {
+		req.UserID = &userID
+	} else if applicationID != "" {
+		req.ApplicationID = &applicationID
+	}
+
+	_, err = api.RemoveGroupMember(req, scw.WithContext(ctx))
+	if err != nil {
+		return diag.FromErr(err)
+	}
```
quantumsheep, line 133: If the group doesn't exists then it should not return an error

quantumsheep (changes requested):

scaleway/resource_iam_group.go
Monitob: mmmm maybe just clean the state

Codelax: Done as in security group: https://github.com/scaleway/terraform-provider-scaleway/blob/8cdd5f333e5d2f244d7367103f7365f83f2d21cc/scaleway/resource_instance_security_group.go#L292

quantumsheep (approved):

PR #2077, opened 2023-07-28
fix(k8s): set PNID to empty string in read if no PN

Mia-Cross:
It has been reported that when the migration from a legacy cluster to a cluster with private network fails, the private_network_id gets stored in the state regardless, so when attempting a new migration, Terraform detects a change of private_network_id, resulting in a ForceNew of the cluster and the loss of data.

This can be avoided by setting the private_network_id to an empty string in the read function if there is no PN.


scaleway/resource_k8s_cluster.go
```diff
@@ -520,6 +520,7 @@ func resourceScalewayK8SClusterRead(ctx context.Context, d *schema.ResourceData,
 	_ = d.Set("auto_upgrade", clusterAutoUpgradeFlatten(cluster))
 
 	// private_network
+	_ = d.Set("private_network_id", "")
 	if cluster.PrivateNetworkID != nil {
 		_ = d.Set("private_network_id", cluster.PrivateNetworkID)
 	}
```
quantumsheep, line 526:
```suggestion
	_ = d.Set("private_network_id", flattenStringPtr(cluster.PrivateNetworkID))
```

You can use `flattenStringPtr`

quantumsheep (changes requested):

quantumsheep (approved):

PR #2097, opened 2023-08-17
fix(instance): avoid ForceNew if server.Image.ID matches the image label

### LINKED ISSUE #2094: instance: image change wrongly detected
### Affected Resource(s)

* scaleway_instance_server

### Expected Behavior

No difference should be detected if the image label corresponds to the local-image ID stored in the state.

### Actual Behavior

During development of Terraform support for kOps I have come across a case in which Terraform detects a change in instance image when actually there are none, the value is just not of the same form (zoned ID of local image vs image's label).

### Steps to Reproduce

1. create an instance with terraform --> the value for the image that gets stored in the state is in the form of a zoned ID of the local image.
2. destroy the instance outside of terraform (by a rolling-update of the kOps cluster)
3. `terraform state rm scaleway_instance_server.$NAME`
4. `terraform import scaleway_instance_server.$NAME $NEW_INSTANCE_ID`
5. `terraform plan` --> the value for the image in the plan is in the form of an image label, so a difference is detected and this forces replacement.

Mia-Cross:
Closes #2094


scaleway/resource_instance_server_test.go
```diff
@@ -1404,3 +1404,43 @@ func TestAccScalewayInstanceServer_MigrateInvalidLocalVolumeSize(t *testing.T) {
...
+				Config: `
+					data "scaleway_marketplace_image" "jammy" {
+						label = "ubuntu_jammy"
+					}
+					resource "scaleway_instance_server" "main" {
+						image = data.scaleway_marketplace_image.jammy.id
+						type = "DEV1-S"
+					}
+				`,
+				ExpectNonEmptyPlan: false,
```
Codelax, line 1438: `ExpectNonEmptyPlan` is false per default. You should look into `PlanOnly` to ensure that the plan is empty.
Here the test could be successful if the instance is re-created.

Codelax (approved):

PR #2135, opened 2023-09-26
fix(lb): update health_check_port when not set by user

yfodil:


scaleway/resource_lb_backend.go
```diff
@@ -613,6 +613,13 @@ func resourceScalewayLbBackendUpdate(ctx context.Context, d *schema.ResourceData
 		updateHCRequest.TCPConfig = expandLbHCTCP(d.Get("health_check_tcp"))
 	}
 
+	rawConfig := d.GetRawConfig().AsValueMap()
```
remyleone, line 616: Why is it necessary to use a GetRawConfig here? Why not extract it from the state?

yfodil: There is the case where for example we specify in the create both forward_port=8080 and health_check_port=8080,
when updating, if you change forward_port to 80, if we extract the config from the state the provider won't know whether health_check_port was set to 8080 by default or whether the user specified its value manually (if it was set manually we don't want to update it)

remyleone (approved):

PR #2182, opened 2023-10-24
feat(ipam): upgrade to v1 API & add missing filters

yfodil:


scaleway/data_source_ipam_ip.go
```diff
@@ -64,8 +70,18 @@ func dataSourceScalewayIPAMIP() *schema.Resource {
...
-
+			"tags": {
+				Type: schema.TypeList,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+				Optional:    true,
+				Description: "The tags associated with the IP",
+			},
+			"zonal":           zoneSchema(),
```
remyleone, line 81: Should it be zone?

yfodil: referred as `zonal` on [api side](https://github.com/scaleway/scaleway-sdk-go/blob/master/api/ipam/v1/ipam_sdk.go#L397)

remyleone (approved):

PR #2188, opened 2023-10-27
feat(ipam): add ipam ip resource

yfodil:


scaleway/helpers_vpc.go
```diff
@@ -51,13 +51,13 @@ func vpcAPIWithRegionAndID(m interface{}, id string) (*v2.API, scw.Region, strin
 	return vpcAPI, region, ID, err
 }
 
-func vpcAPI(m interface{}) (*v1.API, error) {
+func vpcAPI(m interface{}) (*v2.API, error) {
```
remyleone, line 54: Can the import on vpc v1 be removed? That way we have a full consistant and vpc v2 api usage?

docs/resources/ipam_ip.md
```diff
@@ -0,0 +1,113 @@
...
+
+resource "scaleway_vpc_private_network" "pn01" {
+  vpc_id = scaleway_vpc.vpc01.id
+  ipv4_subnet {
+    subnet = "172.16.32.0/22"
+  }
+}
+
+resource "scaleway_ipam_ip" "ip01" {
+  address = "172.16.32.7/22"
```
remyleone, line 48: Is it a single IP /32 or a range?

scaleway/resource_ipam_ip_test.go
```diff
@@ -0,0 +1,282 @@
...
+					resource scaleway_ipam_ip ip01 {
+					 address = "172.16.32.7"
+					 source {
+						private_network_id = scaleway_vpc_private_network.pn01.id
+					  }
+					}
+				`,
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckScalewayIPAMIPExists(tt, "scaleway_ipam_ip.ip01"),
+					resource.TestCheckResourceAttr("scaleway_ipam_ip.ip01", "address", "172.16.32.7/22"),
```
remyleone, line 118: Is it normal that address is a IPv4 without CIDR and in the test is with a CIDR?

docs/resources/ipam_ip.md
yfodil: It's a single IP in CIDR notation that falls in the range of the PN's subnet `172.16.32.0/22`

scaleway/resource_ipam_ip_test.go
yfodil: Yes its normal, the API permits to the user to book an IP with CIDR and without it (in this case it needs to fall within the correct range of the allocatable addresses of the PN's subnet), the API always returns the IP with CIDR in the response.

remyleone (approved):

PR #2189, opened 2023-10-27
feat(object): add SCW_S3_ENDPOINT for custom s3 endpoint

Sh4d1:


scaleway/resource_object_bucket.go
```diff
@@ -533,8 +533,6 @@ func resourceScalewayObjectBucketRead(ctx context.Context, d *schema.ResourceDat
 
 	_ = d.Set("cors_rule", flattenBucketCORS(corsResponse))
 
-	_ = d.Set("endpoint", fmt.Sprintf("https://%s.s3.%s.scw.cloud", bucketName, region))
```
remyleone, line 536: It seems like a breaking change :/ if people are using this field, this will break their code.

Sh4d1: Nop, check a few lines above

remyleone (approved):

PR #2293, opened 2023-12-08
fix(apple-silicon): add M2-L type in validate func

yfodil:


scaleway/resource_apple_silicon_server.go
```diff
@@ -38,7 +38,8 @@ func resourceScalewayAppleSiliconServer() *schema.Resource {
 				Required:    true,
 				ForceNew:    true,
 				ValidateFunc: validation.StringInSlice([]string{
```
remyleone, line 40: Should we remove this entierly? Otherwise we would be force to update everytime a new type comes along.

remyleone (approved):

PR #2302, opened 2023-12-12
fix(object): getting a 403 when reading bucket should not exit the provider

Mia-Cross:
Blocks #2280

### DIFF
--- a/scaleway/helpers_object.go
+++ b/scaleway/helpers_object.go
@@ -593,7 +593,7 @@ func normalizeOwnerID(id *string) *string {
 	return &tab[0]
 }
 
-func addErrorDiagnostic(diags *diag.Diagnostics, err error, resource string, awsResourceNotFoundCode string) (bool, bool) {
+func addReadBucketErrorDiagnostic(diags *diag.Diagnostics, err error, resource string, awsResourceNotFoundCode string) (bucketFound bool, resourceFound bool) {
 	switch {
 	case isS3Err(err, s3.ErrCodeNoSuchBucket, ""):
 		*diags = append(*diags, diag.Diagnostic{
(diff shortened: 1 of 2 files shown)


scaleway/helpers_object.go
```diff
@@ -590,3 +592,31 @@ func normalizeOwnerID(id *string) *string {
 
 	return &tab[0]
 }
+
+func addErrorDiagnostic(diags *diag.Diagnostics, err error, resource string, awsResourceNotFoundCode string) (bool, bool) {
```
Codelax, line 596: could the function be more object specific in the naming ?
addBucketErrorDiagnostic ?

Codelax: document or name the return values

scaleway/resource_object_bucket.go
```diff
@@ -468,15 +470,13 @@ func resourceScalewayObjectBucketRead(ctx context.Context, d *schema.ResourceDat
 		Bucket: aws.String(bucketName),
 	})
 	if err != nil {
-		switch {
-		case isS3Err(err, ErrCodeObjectLockConfigurationNotFoundError, ""):
-			_ = d.Set("object_lock_enabled", false)
-		case isS3Err(err, s3.ErrCodeNoSuchBucket, ""):
-			tflog.Error(ctx, fmt.Sprintf("Bucket %q was not found - removing from state!", bucketName))
+		bucketFound, objectLockFound := addErrorDiagnostic(&diags, err, "object lock configuration", ErrCodeObjectLockConfigurationNotFoundError)
```
Codelax, line 473: the `resource` value here can lead to `cty.GetAttrPath` on an invalid input.
This is the same after, I skip creating multiple comments on this

PR #2308, opened 2023-12-15
fix(rdb): update endpoints handling

### LINKED ISSUE #2281: Integrate changes to endpoints in RDB instance
The RDB API had changes concerning endpoints, it is now possible to create an instance without any public endpoint, which makes the `disable_public_endpoint` field deprecated as it was used to delete the public endpoint after its creation by default with the instance.
And as the behavior changed, the nightly tests for rdb_instance with a private network fail for now.

To integrate this change, we will also need to make changes to the schema of the resource in the way endpoints are defined by the user. This should look like that for configuring both a public and a private endpoint:
```
resource scaleway_vpc_private_network pn {}

resource scaleway_rdb_instance example {
    [...]
    private_network {
         pn_id = scaleway_vpc_private_network.pn.id
    }
    load_balancer {}
}
```

### LINKED ISSUE #2299: RDB update PN doesn't use IPAM
Hello, 

When creating a rdb instance with a PN and IPAM, it works like expected, the config send to the API is using ipam_config flag. 
When updating the PN id to a new one using another IPAM subnet, it seems to drop and recreate the instance endpoint but reusing the service_ip already defined with the previous configuration. 
In case of update of a PN, we should check the client request and do not push the service_ip field if the client request an ipam_config. 
service_ip is still necessary if the endpoint is configured with a service_ip (without ipam)

Thanks a lot, 

### Community Note

### Affected Resource(s)

* scaleway_rdb_instance

```hcl
provider "scaleway" {
  access_key      = var.secret.scaleway.access_key
  secret_key      = var.secret.scaleway.secret_key
  organization_id = var.scaleway.organization_id
  project_id      = var.scaleway.project_id
  region          = var.scaleway.region
  zone            = var.scaleway.zone
}

resource "scaleway_vpc" "demo_vpc" {
  name       = "demo_vpc"
  region     = var.scaleway.region
  project_id = var.scaleway.project_id
}

resource "scaleway_vpc_private_network" "subnet_0" {
  name       = "subnet_0"
  vpc_id     = scaleway_vpc.demo_vpc.id
  project_id = var.scaleway.project_id
  region     = var.scaleway.region
  ipv4_subnet {
    subnet = "192.168.0.0/24"
  }
}

resource "scaleway_vpc_private_network" "subnet_1" {
  name       = "subnet_1"
  vpc_id     = scaleway_vpc.demo_vpc.id
  project_id = var.scaleway.project_id
  region     = var.scaleway.region
  ipv4_subnet {
    subnet = "192.168.1.0/24"
  }
}

resource "scaleway_rdb_instance" "demo" {
  project_id     = var.scaleway.project_id
  name           = "rdb_migrate_pn_bug_demo"
  region         = var.scaleway.region
  node_type      = "DB-DEV-S"
  engine         = "MySQL-8"
  disable_backup = true
  is_ha_cluster  = false
  user_name      = "demo"
  password       = "Azety123!"
  private_network {
    pn_id = scaleway_vpc_private_network.subnet_1.id
  }
}
```

### Steps to Reproduce

1. `terraform apply`

### References

* #0000

Mia-Cross:
This PR changes the way endpoints are handled to integrate the latest changes in the RDB API: if a private network is defined, there is no more default public endpoint. So if the user wants both a private and a public endpoint, they must explicitly declare a `load_balancer {}` empty block alongside the `private_network`. This also means there is no need for the `disable_public_endpoint` field anymore.
Closes #2281 

Also there was a problem with the updates of private network endpoints, that did not always use IPAM as they should.
Closes #2299

The PR should also fix the nightly tests.

### DIFF
--- a/scaleway/resource_rdb_instance.go
+++ b/scaleway/resource_rdb_instance.go
@@ -147,6 +147,12 @@ func resourceScalewayRdbInstance() *schema.Resource {
 							DiffSuppressFunc: diffSuppressFuncLocality,
 							Description:      "The private network ID",
 						},
+						"enable_ipam": {
+							Type:        schema.TypeBool,
+							Optional:    true,
+							Computed:    true,
+							Description: "Whether or not the private network endpoint should be configured with IPAM",
+						},
 						// Computed
 						"endpoint_id": {
 							Type:        schema.TypeString,
@@ -185,12 +191,6 @@ func resourceScalewayRdbInstance() *schema.Resource {
 						"zone": zoneSchema(),
 					},
 				},
-			},
-			"disable_public_endpoint": {
-				Type:        schema.TypeBool,
-				Optional:    true,
-				Default:     false,
-				Description: "Whether the instance should have a public endpoint if it has a Private Network attached",
 			},
 			// Computed
 			"endpoint_ip": {
@@ -235,6 +235,7 @@ func resourceScalewayRdbInstance() *schema.Resource {
 			},
 			"load_balancer": {
 				Type:        schema.TypeList,
+				Optional:    true,
 				Computed:    true,
 				Description: "Load balancer of the database instance",
 				Elem: &schema.Resource{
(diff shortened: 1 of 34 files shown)


Mia-Cross:
Examples :
```
scaleway_rdb_instance example {
    private_network {
        pn_id = xxx
        (enable_ipam = true)
    }
}
```
--> Defines an IPAM configured private network endpoint

```
scaleway_rdb_instance example {
    private_network {
        pn_id = xxx
        enable_ipam = false
    }
    load_balancer {}
}
```
--> Defines a static private network endpoint and a load_balancer public endpoint

```
scaleway_rdb_instance example {
}
```
--> Defines a default load_balancer public endpoint

scaleway/resource_rdb_instance.go
```diff
@@ -186,12 +192,6 @@ func resourceScalewayRdbInstance() *schema.Resource {
 					},
 				},
 			},
-			"disable_public_endpoint": {
```
remyleone, line 189: Could you add examples of configuration in the documentation? I guess that examples need to be updated to showcase

remyleone (approved):

PR #2309, opened 2023-12-18
feat(k8s): add warning if cluster type is not prefixed by kapsule or multicloud

pablo-ruth:
This PR add a warning if cluster type is not prefixed by kapsule or multicloud, instead of returning an error.

### DIFF
--- a/scaleway/resource_k8s_cluster.go
+++ b/scaleway/resource_k8s_cluster.go
@@ -3,7 +3,6 @@ package scaleway
 import (
 	"context"
 	"fmt"
-	"slices"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


scaleway/resource_k8s_cluster.go
```diff
@@ -335,6 +335,14 @@ func resourceScalewayK8SClusterCreate(ctx context.Context, d *schema.ResourceDat
 		clusterType = ""
 	}
 
+	if !slices.ContainsFunc([]string{"kapsule", "multicloud"}, func(s string) bool { return strings.HasPrefix(clusterType.(string), s) }) {
+		diags = append(diags, diag.Diagnostic{
+			Severity: diag.Warning,
+			Summary:  "Unexpected cluster type",
+			Detail:   fmt.Sprintf("The expected cluster type is one of %v, but got %s", []string{"kapsule", "multicloud", "kapsule-dedicated-*", "multicloud-dedicated-*"}, clusterType.(string)),
+		})
```
Codelax, line 343: You can add attribute path to have a better output:
```
AttributePath: cty.GetAttrPath("type"),
```

scaleway/resource_k8s_cluster.go
```diff
@@ -3,6 +3,7 @@ package scaleway
 import (
 	"context"
 	"fmt"
+	"slices"
```
Codelax, line 6: We should bump to go 1.21

scaleway/resource_k8s_cluster.go
pablo-ruth: done :)

scaleway/resource_k8s_cluster.go
pablo-ruth: ok I replaced the call to slices by a simple if as there is only two elements

Codelax (approved):

PR #2318, opened 2023-12-19
feat(container): add triggers attribute

Codelax:


Codelax:
There already is a `registry_sha256` to trigger a deployment, it is missing in documentation

PR #2360, opened 2024-01-11
tests(vpc): add test and doc for filtering by vpc_id in datasource

yfodil:


scaleway/data_source_vpc_private_network.go
```diff
@@ -52,16 +52,15 @@ func dataSourceScalewayVPCPrivateNetworkRead(ctx context.Context, d *schema.Reso
...
 		if err != nil {
 			return diag.FromErr(err)
 		}
 
-		vpcID, hasVpcIDFilter := d.GetOk("vpc_id")
-
 		foundPN, err := findExact(
 			res.PrivateNetworks,
-			func(s *vpc.PrivateNetwork) bool { return s.Name == pnName && (!hasVpcIDFilter || s.VpcID == vpcID) },
+			func(s *vpc.PrivateNetwork) bool { return s.Name == pnName },
```
remyleone, line 63: There is no unicity with name on a private network. Could you add a test that returns an error in case several private networks within the same vpc share the same name?

remyleone (approved):

PR #2438, opened 2024-03-04
test(redis): remove subnet from vpc to avoid `subnet overlap in vpc`

Mia-Cross:
The Redis nightly tests failed because of an error during the creation of the private network (`subnets is invalid for unexpected reason, subnet_overlaps_in_vpc: {}`), removing said subnet from the configuration may resolve the issue as it is not mandatory.


fenollp (approved):

Mia-Cross:
@Monitob The examples for Redis don't have any `vpc_private_network` resource with its `ipv4_subnet` attribute set, it was in the tests only. There is one occurrence in the `rdb_instance` documentation also, but I just tested it and it works :+1:

remyleone (approved):

PR #2461, opened 2024-03-14
fix(rdb): allow updates from bssd to lssd if new node type has enough space

Mia-Cross:
I noticed a case that did not occur in the tests but there was a bug when changing the volume type from `bssd` to `lssd` as the `volume_size_in_gb` attribute goes from the given value to 0 (the volume size depends on the node type). And since we carry out the upgrades starting from the volume and ending with the node type, that resulted in the API rejecting the update of the `volume_type` because the `node_type` should be updated first.

To fix the issue, this PR adds the comparing of the actual and the wanted storage capacities to see if the node_type has to be updated first.

### DIFF
--- a/scaleway/resource_rdb_instance.go
+++ b/scaleway/resource_rdb_instance.go
@@ -593,7 +593,7 @@ func resourceScalewayRdbInstanceUpdate(ctx context.Context, d *schema.ResourceDa
 
 	// If we are switching to local storage, we have to make sure that the node_type upgrade is done first
 	if d.HasChange("volume_type") {
-		_, wantedVolumeType := d.GetChange("volume_type")
+		wantedVolumeType := d.Get("volume_type")
 		if wantedVolumeType == rdb.VolumeTypeLssd.String() {
 			for i, req := range upgradeInstanceRequests {
 				if req.NodeType != nil && i != 0 {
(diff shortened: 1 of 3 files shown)


scaleway/resource_rdb_instance.go
```diff
@@ -498,7 +498,7 @@ func resourceScalewayRdbInstanceUpdate(ctx context.Context, d *schema.ResourceDa
 	////////////////////
 	// Upgrade instance
 	////////////////////
-	upgradeInstanceRequests := []rdb.UpgradeInstanceRequest(nil)
+	upgradeInstanceRequests := make(map[string]*rdb.UpgradeInstanceRequest)
```
Codelax, line 501: Order was respected with a slice. Map are unordered.
Is order of upgrade's requests important other than `node_type` ?

scaleway/resource_rdb_instance.go
```diff
@@ -583,22 +579,42 @@ func resourceScalewayRdbInstanceUpdate(ctx context.Context, d *schema.ResourceDa
...
+			EnableHa:   scw.BoolPtr(d.Get("is_ha_cluster").(bool)),
+		}
 	}
 
-	// Carry out the upgrades
-	for i := range upgradeInstanceRequests {
+	// Carry out the upgrades in the correct order
+	// First we have to check if we have enough storage capacity on the actual node.
+	actualStorage := rdbInstance.Volume.Size
+	wantedStorage := scw.Size(d.Get("volume_size_in_gb").(int) * 10e9)
```
Codelax, line 592: Constants in code should be avoided. You can use our size helpers (`scw.GB`, `scw.MB`, ...)

Codelax (approved):

PR #2509, opened 2024-04-05
refactor(sweepers): change sweepers format to include dependencies

Codelax:


internal/services/block/blocktestfuncs/sweep.go
```diff
@@ -0,0 +1,78 @@
+package blocktestfuncs
```
remyleone, line 1: testfuncs sounds like a better package/folder name

PR #2536, opened 2024-04-16
feat(rdb): add log policy in instance

jremy42:
close #2047


internal/services/rdb/helpers.go
```diff
@@ -145,3 +147,44 @@ func getIPAMConfigRead(resource interface{}, m interface{}) (bool, error) {
 	}
 	return false, nil
 }
+
+func flattenSizePtr(i *scw.Size) interface{} {
```
remyleone, line 151: Should be migrated to types as it is not related to rdb per se

internal/services/rdb/helpers.go
```diff
@@ -18,8 +18,10 @@ import (
 )
 
 const (
-	defaultInstanceTimeout   = 15 * time.Minute
-	defaultWaitRetryInterval = 30 * time.Second
+	defaultInstanceTimeout           = 15 * time.Minute
+	defaultWaitRetryInterval         = 30 * time.Second
+	defaultMaxAgeRetention    uint32 = 30
```
remyleone, line 23: They seem to be un-used

testdata/function-basic.cassette.yaml
remyleone, line 1: Why does this cassette is modified?

remyleone (changes requested):

internal/services/rdb/helpers.go
```diff
@@ -145,3 +145,28 @@ func getIPAMConfigRead(resource interface{}, m interface{}) (bool, error) {
 	}
 	return false, nil
 }
+
+func expandInstanceLogsPolicy(i interface{}) *rdb.LogsPolicy {
```
remyleone, line 149: should be in rdb types file

remyleone (approved):

PR #2543, opened 2024-04-18
feat(cockpit): add datasources

yfodil:


internal/services/cockpit/datasource.go
```diff
@@ -0,0 +1,147 @@
...
+			return nil
+		}
+		return diag.FromErr(err)
+	}
+
+	_ = d.Set("name", res.Name)
+	_ = d.Set("type", res.Type.String())
+	_ = d.Set("url", res.URL)
+	_ = d.Set("origin", res.Origin)
+	_ = d.Set("synchronized_with_grafana", res.SynchronizedWithGrafana)
```
Codelax, line 123: Should it contains a waiter for `synchronized_with_grafana` ? Is this required to use the datasource ?

yfodil: We don't have to wait for this field for the resource to be ready

internal/services/cockpit/helpers_cockpit.go
```diff
@@ -21,14 +24,34 @@ const (
 )
 
 // NewAPI returns a new cockpit API.
-func NewAPI(m interface{}) (*cockpit.API, error) {
-	api := cockpit.NewAPI(meta.ExtractScwClient(m))
+func NewAPI(m interface{}) (*cockpitv1beta1.API, error) {
```
remyleone, line 27: Could we manage only the v1 api? Is there a premium to maintain both APIs?

yfodil: As we want to keep the legacy cockpit resource for some time, we have to leave the v1beta1 api to make the calls, in v1 there's no longer any notion of cockpit

PR #2575, opened 2024-05-21
docs: add instance example

Codelax:
Init a new examples folder. Examples will be tested by CI in the future

### DIFF
--- a/examples/instance-servers/outputs.tf
+++ b/examples/instance-servers/outputs.tf
@@ -1,0 +1,3 @@
+output "ips_v4" {
+  value = scaleway_instance_server.server[*].public_ip
+}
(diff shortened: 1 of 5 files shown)


examples/instance-servers/outputs.tf
```diff
@@ -0,0 +1,3 @@
+output "ips_v4" {
+  value = scaleway_instance_server.server[*].public_ip
+}
```
yfodil, line 3:
```suggestion
output "ips_v4" {
  description = "The public IPv4 addresses of the created instance servers"
  value       = scaleway_instance_server.server[*].public_ip
}

```

yfodil (approved):

PR #2577, opened 2024-05-23
feat(vpc): add support for enable_routing

yfodil:


internal/services/vpc/vpc.go
```diff
@@ -58,6 +65,20 @@ func ResourceVPC() *schema.Resource {
 				Description: "The date and time of the last update of the private network",
 			},
 		},
+		CustomizeDiff: func(_ context.Context, diff *schema.ResourceDiff, _ interface{}) error {
+			rawConfig := diff.GetRawConfig().AsValueMap()
+			before, after := diff.GetChange("enable_routing")
+
+			// If the field was set to true and now it is either false or not set, force recreation
```
Codelax, line 72: Doesn't `not set` defaults to false ?

Codelax: This was made so remove `enable_routing` is the same as setting it to false

Codelax (approved):

PR #2590, opened 2024-05-30
feat(core): data source config reader

Mia-Cross:


internal/services/scwconfig/test_config.yaml
```diff
@@ -0,0 +1,81 @@
+# Scaleway configuration file
```
Codelax, line 1: This test file could be moved to a testfixture folder, similar to how it's done with [function](https://github.com/scaleway/terraform-provider-scaleway/tree/1e5cb73e47eaf85e5e2f6baeaa73c13ec306f430/internal/services/function/testfixture)

Codelax (approved):

PR #2600, opened 2024-06-04
feat: add enum validation

yfodil:


internal/verify/enum.go
```diff
@@ -0,0 +1,41 @@
...
+
+type EnumValues[T ~string] interface {
+	~string
+	Values() []T
+}
+
+// ValidateEnum creates a schema validation function for the provided enum type
+func ValidateEnum[T EnumValues[T]]() schema.SchemaValidateDiagFunc {
+	values := filterUnknownValues(getValues[T]())
+	return validation.ToDiagFunc(validation.StringInSlice(values, false))
```
Codelax, line 18: `ignoreCase` now default to `false` where it was previously `true`

internal/verify/enum.go
```diff
@@ -0,0 +1,41 @@
...
+import (
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+)
+
+type EnumValues[T ~string] interface {
+	~string
+	Values() []T
```
Codelax, line 12: Interface could have been `Values() string` to provide a more idiomatic Go. This would allow a simple interface rather than generics.
This change require fixing sdk generation.

internal/verify/enum.go
yfodil: `ignoreCase` has always been defaulted to false, the only case I found where its set to true is on [pat_rule's protocol field](https://github.com/scaleway/terraform-provider-scaleway/blob/master/internal/services/vpcgw/pat_rule.go#L72)

Codelax (approved):

PR #2615, opened 2024-06-14
docs(k8s_cluster): bump k8s version

Codelax:


docs/guides/multicloud_cluster_with_baremetal_servers.md
```diff
@@ -26,7 +26,7 @@ as a node by the apiserver. This can be achieved manually ([method A](#method-a-
 resource "scaleway_k8s_cluster" "multicloud" {
   name    = "multicloud-cluster"
   type    = "multicloud"
-  version = "1.27.1"
```
remyleone, line 29: We would change this all the time, why not using a data source with latest to have an example that is always up to date and can be copy and pasted easily?

docs/resources/k8s_cluster.md
```diff
@@ -60,7 +60,7 @@ resource "scaleway_vpc_private_network" "hedy" {}
 resource "scaleway_k8s_cluster" "john" {
```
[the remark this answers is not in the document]
Codelax: I renamed resources

docs/guides/multicloud_cluster_with_baremetal_servers.md
Codelax: There is multiple snippets that should work with a simple copy/paste, using a valid version help that.
We could use a minor version `X.Y` but it would require enabling auto-upgrade

remyleone (approved):

PR #2641, opened 2024-07-01
feat(instance_server): support sbs_volume as root volume

Codelax:


docs/resources/instance_server.md
```diff
@@ -163,6 +163,20 @@ resource "scaleway_instance_server" "from_snapshot" {
 }
 ```
 
+#### Using Scaleway Block Storage (SBS) volume
+
+```terraform
+resource "scaleway_instance_server" "server" {
+  type = "PLAY2-MICRO"
+  image = "ubuntu_jammy"
+  root_volume {
```
remyleone, line 172: How is it possible to find this volume from a data source? Does it have a name/id ?

Codelax: This creates a new volume, why would we use a data source ?

remyleone (approved):

PR #2658, opened 2024-07-15
tests(lb): fix unexpected ip count

yfodil:
Moved resources to another zone to avoid unexpected ip count from the other tests

### DIFF
--- a/internal/services/lb/ips_data_source_test.go
+++ b/internal/services/lb/ips_data_source_test.go
@@ -62,40 +62,43 @@ func TestAccDataSourceIPs_WithType(t *testing.T) {
 		Steps: []resource.TestStep{
 			{
 				Config: `
-					resource scaleway_lb_ip ip1 {
+					resource "scaleway_lb_ip" "ip1" {
 					  zone = "nl-ams-1"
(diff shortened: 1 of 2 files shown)


internal/services/lb/ips_data_source_test.go
```diff
@@ -63,31 +63,38 @@ func TestAccDataSourceIPs_WithType(t *testing.T) {
...
 					resource scaleway_lb_ip ip3 {
+					  zone = "nl-ams-1"
 					}
 
 					data "scaleway_lb_ips" "ips_by_type" {
-						ip_type = "ipv4"
-						depends_on = [scaleway_lb_ip.ip1, scaleway_lb_ip.ip2, scaleway_lb_ip.ip3]
+                      ip_type    = "ipv4"
+					  zone       = "nl-ams-1"
+					  depends_on = [scaleway_lb_ip.ip1, scaleway_lb_ip.ip2, scaleway_lb_ip.ip3]
```
Codelax, line 97: It looks like a mix of tab and space. Indentation should be fixed

Codelax (approved):

PR #2673, opened 2024-07-31
feat(tem): add support for webhook

jremy42:


internal/services/tem/webhook.go
```diff
@@ -0,0 +1,193 @@
...
+				Required:     true,
+				Description:  "SNS ARN",
+				ValidateFunc: validation.StringLenBetween(3, 127),
+			},
+			"organization_id": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "The organization id",
+			},
+			"id": {
```
remyleone, line 63: When is this Id updated? Why you put an id in the schema explicitely while it is not the case in other resources?

internal/services/tem/webhook.go
```diff
@@ -0,0 +1,184 @@
...
+	api, region, err := temAPIWithRegion(d, m)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	eventTypesInterface := d.Get("event_types").([]interface{})
+	eventTypes := make([]tem.WebhookEventType, len(eventTypesInterface))
+	for i, v := range eventTypesInterface {
+		eventTypes[i] = tem.WebhookEventType(v.(string))
+	}
```
Codelax, line 83: Can this be extracted to a function ? It could be used in Update too.
```
eventTypes := expandWebhookEventTypes(d.Get("event_types"))
```

Codelax (approved):

PR #2674, opened 2024-07-31
feat(iam): add support for condition

remyleone:


internal/services/iam/policy.go
```diff
@@ -108,6 +108,11 @@ func ResourcePolicy() *schema.Resource {
 								Type: schema.TypeString,
 							},
 						},
+						"condition": {
+							Type:        schema.TypeString,
+							Description: "Conditions of the policy written in CEL (https://github.com/google/cel-spec/blob/master/doc/langdef.md)",
```
remyleone, line 113: Adding documentation with examples would be useful

docs/resources/iam_policy.md
```diff
@@ -90,6 +90,29 @@ resource "scaleway_iam_policy" "iam_tf_storage_policy" {
...
+### Create a policy with a particular condition
+
+IAM policy rule can use a condition to be applied.
+This condition is written in a [CEL](https://github.com/google/cel-spec/blob/master/doc/langdef.md) format.
+The following variables are available:
+- `request.ip`
+- `request.user_agent`
+- `request.time`
+- `session.bearer_id`
+- `session.staff`
```
aerialls, line 102: To be removed, they are not public.

aerialls (changes requested):

internal/services/iam/policy.go
quantumsheep, line 113: We should not specify that the format is CEL

Codelax (dismissed):

Codelax (approved):

aerialls (approved):

quantumsheep (approved):

PR #2676, opened 2024-08-01
feat(rdb): add encryption at rest for instance

jremy42:


internal/services/rdb/instance.go
```diff
@@ -302,6 +302,11 @@ func ResourceInstance() *schema.Resource {
 					},
 				},
 			},
+			"encryption_at_rest": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Description: "Enable or disable encryption at rest for the database instance",
```
yfodil, line 308: This field requires a Forcenew since it can't be updated

internal/services/rdb/instance.go
```diff
@@ -302,6 +302,11 @@ func ResourceInstance() *schema.Resource {
 					},
 				},
 			},
+			"encryption_at_rest": {
```
yfodil, line 305: Documentation is missing

internal/services/rdb/instance_test.go
```diff
@@ -1216,6 +1216,94 @@ func TestAccInstance_Endpoints(t *testing.T) {
...
+				Config: fmt.Sprintf(`
+					resource scaleway_rdb_instance main {
+						name               = "test-rdb-no-encryption"
+						node_type          = "db-dev-s"
+						engine             = %q
+						is_ha_cluster      = false
+						disable_backup     = true
+						user_name          = "my_initial_user_no_enc"
+						password           = "thiZ_is_v&ry_s3cret"
+						encryption_at_rest = false
```
remyleone, line 1284: Could you test if it could be updated? passing from false to true ? Does it require un ForceNew?

internal/services/rdb/instance.go
remyleone, line 305: Could you add documentation?

remyleone (approved):

PR #2695, opened 2024-08-07
fix(iam_ssh_key): do not read public key in resource

Codelax:
This caused the value to differ from requested input and this can lead to issue in terraform if the value is used in other resources. 
Also as it cannot be updated, it seems right to assume the local key is the correct key

### DIFF
--- a/internal/services/iam/ssh_key.go
+++ b/internal/services/iam/ssh_key.go
@@ -55,6 +55,14 @@ func ResourceSSKKey() *schema.Resource {
 					areEqual := bytes.Equal(marshalledOldValue, marshalledNewValue)
 					return areEqual
 				},
+				StateFunc: func(v interface{}) string {
+					switch v := v.(type) {
+					case string:
+						return strings.TrimSpace(v)
+					default:
+						return ""
+					}
+				},
 			},
 			"fingerprint": {
 				Type:        schema.TypeString,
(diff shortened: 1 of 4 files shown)


internal/services/iam/ssh_key.go
```diff
@@ -55,6 +55,14 @@ func ResourceSSKKey() *schema.Resource {
 					areEqual := bytes.Equal(marshalledOldValue, marshalledNewValue)
 					return areEqual
 				},
+				StateFunc: func(v interface{}) string {
```
remyleone, line 58: Could you add a bit of documentation explaining that the last part of the public key will be trimmed.

remyleone:
Could you do additional testing to ensure that the formatting happening in the terraform state is consistant with the key stored on the IAM backend?

Codelax:
My test config to reproduce issue:

```tf
provider "scaleway" {
  alias = "project"
}

resource "scaleway_account_project" "project" {
  provider = scaleway.project
}

provider "scaleway" {
  project_id = scaleway_account_project.project.id
}

resource "scaleway_iam_ssh_key" "key1" {
  public_key = "<key with comment>"
}

resource "scaleway_iam_ssh_key" "key2" {
  public_key = "<key with comment>"
}

locals {
  ssh_keys_hash = sha256(join(",", [
    scaleway_iam_ssh_key.key1.public_key,
    scaleway_iam_ssh_key.key2.public_key,
  ]))
}

resource "scaleway_vpc_public_gateway_ip" "ip" {}

resource "scaleway_vpc_public_gateway" "main" {
  ip_id = scaleway_vpc_public_gateway_ip.ip.id
  name             = "public_gateway_demo"
  type             = "VPC-GW-S"
  tags             = ["demo", "terraform"]
  bastion_enabled  = true
  bastion_port     = 61000
  refresh_ssh_keys = local.ssh_keys_hash
}

```

Codelax:
Terraform does not allow required fields to be computed. This means that we cannot rely on API for its formatting of the key.
I don't think we should try to implement API behavior locally in a `StateFunc`. I don't see any other solution, I'm open to ideas.

In my given config, we can use a variable to store key and use this variable for key creation and hashing.

PR #2700, opened 2024-08-12
feat:(cockpit): add push urls to cockpit datasources

jremy42:
close [#2563](https://github.com/scaleway/terraform-provider-scaleway/issues/2563)

### DIFF
--- a/internal/services/cockpit/types.go
+++ b/internal/services/cockpit/types.go
@@ -63,7 +63,6 @@ func createCockpitPushURLList(endpoints []map[string]interface{}) []map[string]i
 			result = append(result, newEndpoint)
 		}
 	}
-
 	return result
 }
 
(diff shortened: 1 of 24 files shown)


internal/services/cockpit/types.go
```diff
@@ -65,6 +65,17 @@ func createCockpitPushURL(endpoints []map[string]interface{}) []map[string]inter
...
+	switch sourceType {
+	case cockpit.DataSourceTypeMetrics:
+		return url + pathMetricsURL
+	case cockpit.DataSourceTypeLogs:
+		return url + pathLogsURL
+	default:
+		return ""
+	}
+}
+
```
yfodil, line 78:
```suggestion
func createCockpitPushURL(sourceType cockpit.DataSourceType, url string) (string, error) {
	switch sourceType {
	case cockpit.DataSourceTypeMetrics:
		return url + pathMetricsURL, nil
	case cockpit.DataSourceTypeLogs:
		return url + pathLogsURL, nil
	default:
		return "", fmt.Errorf("invalid data source type: %v", sourceType)
	}
}
```

yfodil (approved):

PR #2702, opened 2024-08-13
feat(vpc): add custom route resource

yfodil:


internal/services/vpc/route.go
```diff
@@ -0,0 +1,230 @@
...
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "The date and time of the last update of the route",
+			},
+		},
+	}
+}
+
+func ResourceRouteCreate(ctx context.Context, d *schema.ResourceData, m interface{}) diag.Diagnostics {
+	vpcAPI, region, err := vpcAPIWithRegion(d, m)
```
remyleone, line 82: Should an EnableRouting happen here before trying to create a route?

remyleone: It is now enabled by default

PR #2736, opened 2024-09-11
feat(inference): add support CRUD deployment

Laure-di:


internal/provider/provider.go
```diff
@@ -2,6 +2,7 @@ package provider
 
 import (
 	"context"
+	"github.com/scaleway/terraform-provider-scaleway/v2/internal/services/inference"
```
remyleone, line 5: Please make this import align nicely with the other using gofumpt

internal/services/inference/helpers_inference.go
```diff
@@ -0,0 +1,39 @@
+package inference
+
+import (
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	inference "github.com/scaleway/scaleway-sdk-go/api/inference/v1beta1"
+	"github.com/scaleway/scaleway-sdk-go/scw"
+	"github.com/scaleway/terraform-provider-scaleway/v2/internal/locality/regional"
+	"github.com/scaleway/terraform-provider-scaleway/v2/internal/meta"
+	"time"
```
remyleone, line 9: Need to be gofumpt

internal/services/inference/deployment.go
```diff
@@ -0,0 +1,271 @@
...
+	inference "github.com/scaleway/scaleway-sdk-go/api/inference/v1beta1"
+	"github.com/scaleway/terraform-provider-scaleway/v2/internal/httperrors"
+	"github.com/scaleway/terraform-provider-scaleway/v2/internal/locality/regional"
+	"github.com/scaleway/terraform-provider-scaleway/v2/internal/services/account"
+	"github.com/scaleway/terraform-provider-scaleway/v2/internal/types"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/scaleway/scaleway-sdk-go/scw"
+	_ "time"
```
remyleone, line 14: Why is this import ignored?

internal/services/inference/deployment.go
```diff
@@ -0,0 +1,271 @@
...
+
+import (
+	"context"
+	inference "github.com/scaleway/scaleway-sdk-go/api/inference/v1beta1"
+	"github.com/scaleway/terraform-provider-scaleway/v2/internal/httperrors"
+	"github.com/scaleway/terraform-provider-scaleway/v2/internal/locality/regional"
+	"github.com/scaleway/terraform-provider-scaleway/v2/internal/services/account"
+	"github.com/scaleway/terraform-provider-scaleway/v2/internal/types"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
```
remyleone, line 11: gofumpt

internal/services/inference/deployment.go
```diff
@@ -0,0 +1,271 @@
...
+
+	if maxSize, ok := d.GetOk("max_size"); ok {
+		req.MaxSize = scw.Uint32Ptr(uint32(maxSize.(int)))
+	}
+
+	if minSize, ok := d.GetOk("min_size"); ok {
+		req.MaxSize = scw.Uint32Ptr(uint32(minSize.(int)))
+	}
+
+	req.Endpoints = []*inference.EndpointSpec{&endpoint}
```
remyleone, line 169: Move the code that is referring to endpoints closer together to reduce cognitive load

internal/services/inference/deployment_test.go
```diff
@@ -0,0 +1,129 @@
...
+	"github.com/scaleway/terraform-provider-scaleway/v2/internal/acctest"
+	"github.com/scaleway/terraform-provider-scaleway/v2/internal/httperrors"
+	"github.com/scaleway/terraform-provider-scaleway/v2/internal/services/inference"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+)
+
+func TestAccDeployment_Basic(t *testing.T) {
```
remyleone, line 15: Could you add a test with a private endpoint?

internal/services/inference/deployment.go
```diff
@@ -0,0 +1,308 @@
...
+						"public_endpoint": {
+							Type:        schema.TypeBool,
+							Description: "Set the endpoint as public",
+							Optional:    true,
+						},
+						"private_endpoint": {
+							Type:        schema.TypeString,
+							Description: "The id of the private network",
+							Optional:    true,
+						},
```
Codelax, line 146: Having `public_endpoint` as a bool and `private_endpoint` as an ID is misleading.
These fields should have names that suit their use, example: `is_public`, `private_network_id`

internal/services/inference/deployment.go
```diff
@@ -0,0 +1,308 @@
...
+	_ = d.Set("model_name", deployment.ModelName)
+	_ = d.Set("min_size", int(deployment.MinSize))
+	_ = d.Set("max_size", int(deployment.MaxSize))
+	_ = d.Set("size", int(deployment.Size))
+	_ = d.Set("status", deployment.Status)
+	_ = d.Set("model_id", deployment.ModelID)
+	_ = d.Set("created_at", types.FlattenTime(deployment.CreatedAt))
+	_ = d.Set("updated_at", types.FlattenTime(deployment.UpdatedAt))
+
+	for _, endpoint := range deployment.Endpoints {
```
Codelax, line 241: Should this export as `endpoints` ? User configure its `endpoints` but get different fields after StateRead.

internal/services/inference/deployment.go
```diff
@@ -0,0 +1,308 @@
...
+				Description: "The endpoint public ID",
+			},
+			"endpoint_private_id": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "The endpoint private ID",
+			},
+			"disable_auth_public": {
+				Type:        schema.TypeBool,
+				Computed:    true,
```
remyleone, line 118: Should an optional be present here?

internal/services/inference/deployment.go
```diff
@@ -0,0 +1,308 @@
...
+				Description: "The endpoint public URL",
+			},
+			"endpoint_private_url": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "The endpoint private URL",
+			},
+			"disable_auth_private": {
+				Type:        schema.TypeBool,
+				Computed:    true,
```
remyleone, line 103: Should an optional be present here?

internal/services/inference/deployment.go
remyleone: It doesn't make much sense to have a field at the root of the ressource being updated with a sub resource downstream :/ would it be better to have all information about endpoint at the same place?

internal/services/inference/deployment.go
```diff
@@ -0,0 +1,349 @@
...
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"id": {
+							Type:        schema.TypeString,
+							Description: "The id of the public endpoint",
+							Computed:    true,
+						},
+						"is_enabled": {
+							Type:        schema.TypeBool,
+							Description: "Set the endpoint as public",
```
Codelax, line 150:
```suggestion
							Description: "Enable or disable public endpoint",
```

internal/services/inference/deployment.go
```diff
@@ -0,0 +1,349 @@
...
+				AtLeastOneOf: []string{"public_endpoint"},
+				Description:  "List of endpoints",
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"id": {
+							Type:        schema.TypeString,
+							Description: "The id of the private endpoint",
+							Computed:    true,
+						},
+						"private_endpoint_id": {
```
Codelax, line 115:
```suggestion
						"private_network_id": {
```
You will also need to change in the rest of the code.

internal/services/inference/deployment.go
```diff
@@ -0,0 +1,349 @@
...
+			publicEndpoints = append(publicEndpoints, publicEndpointSpec)
+		}
+	}
+
+	if len(privateEndpoints) > 0 {
+		_ = d.Set("private_endpoint", privateEndpoints)
+	}
+	if len(publicEndpoints) > 0 {
+		_ = d.Set("public_endpoint", publicEndpoints)
+	}
```
Codelax, line 292: MaxItems is 1 for `private_endpoint` and `public_endpoint`. This code make it seems multiple endpoints are supported. Should we let user specify multiple endpoints in terraform ? If API allows it, terraform provider would be ready.

internal/services/inference/deployment.go
```diff
@@ -0,0 +1,349 @@
...
+		return diag.FromErr(err)
+	}
+	req := &inference.UpdateDeploymentRequest{
+		Region:       region,
+		DeploymentID: deployment.ID,
+	}
+
+	if d.HasChange("name") {
+		req.Name = types.ExpandUpdatedStringPtr(d.Get("name"))
+	}
```
Codelax, line 317: Api allows more field in Update: https://github.com/scaleway/scaleway-sdk-go/blob/a2c1e85de57a916ea17d27c4efe2646fc5f33d11/api/inference/v1beta1/inference_sdk.go#L786-L803

internal/services/inference/deployment.go
```diff
@@ -0,0 +1,349 @@
...
+				Type:        schema.TypeList,
+				Elem:        &schema.Schema{Type: schema.TypeString},
+				Optional:    true,
+				Description: "The tags associated with the deployment",
+			},
+			"min_size": {
+				Type:        schema.TypeInt,
+				Optional:    true,
+				Computed:    true,
+				Description: "The minimum size of the pool",
```
remyleone, line 73: Should add validation about the int being positive. Can the `min_size` be 0?

internal/services/inference/deployment.go
Laure-di: You can have only one private network and only one public

internal/services/inference/deployment.go
```diff
@@ -0,0 +1,349 @@
...
+							Type:        schema.TypeString,
+							Description: "The id of the public endpoint",
+							Computed:    true,
+						},
+						"is_enabled": {
+							Type:        schema.TypeBool,
+							Description: "Set the endpoint as public",
+							Optional:    true,
+						},
+						"disable_auth": {
```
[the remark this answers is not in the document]
Laure-di: by default, deployments are protected by IAM authentication.

So by default the disable_auth is false if not specified

internal/services/inference/deployment.go
remyleone: I would prefer it to maintain the computed

PR #2739, opened 2024-09-13
tests: rename test resources to expected test name

Codelax:
These resources were not deleted by sweepers


yfodil (approved):

internal/services/iot/routes_test.go
```diff
@@ -86,7 +86,7 @@ func TestAccRoute_S3(t *testing.T) {
 	}
 	tt := acctest.NewTestTools(t)
 	defer tt.Cleanup()
-	bucketName := sdkacctest.RandomWithPrefix("test-acc-scaleway-iot-route-s3")
+	bucketName := sdkacctest.RandomWithPrefix("tf-tests-scaleway-iot-route-s3")
```
remyleone, line 89: not the correct prefix I think. Could you extract the prefix and reference it here?

remyleone: https://github.com/scaleway/terraform-provider-scaleway/blob/master/internal/services/object/testfuncs/sweep.go#L36

remyleone: In that case, I would suggest changing testSweepStorageObjectBucket to use IsTestResource instead of strings.HasPrefix

PR #2741, opened 2024-09-17
fix(cockpit): add grafana url manual creation

jremy42:


internal/services/cockpit/types.go
```diff
@@ -18,6 +18,10 @@ var scopeMapping = map[string]cockpit.TokenScope{
 	"write_traces":        cockpit.TokenScopeWriteOnlyTraces,
 }
 
+func createGrafanaURL(projectID string) string {
+	return fmt.Sprintf("https://%s.dashboards.obs.fr-par.scw.cloud", projectID)
```
remyleone, line 22: Can the region be also passed? Otherwise I think we will have to change it down the road, it can be configured by default with `fr-par`

jremy42: Done

remyleone (approved):

PR #2777, opened 2024-10-17
feat(mongodb): add support for v1alpha1

jremy42:


internal/services/mongodb/instance.go
```diff
@@ -0,0 +1,405 @@
...
+			return diag.FromErr(err)
+		}
+	}
+
+	req := &mongodb.UpdateInstanceRequest{
+		Region:     region,
+		InstanceID: ID,
+	}
+
+	if d.HasChange("name") {
```
remyleone, line 332: Introduce a variable that is changed like `should_update` or `has_changed` and if this variable is set then perform the update. That way the conditional line 342 won't get bigger

internal/services/mongodb/instance.go
```diff
@@ -0,0 +1,405 @@
...
+		Region:     region,
+		InstanceID: ID,
+	}
+
+	if d.HasChange("password") {
+		password := d.Get("password").(string)
+		updateUserRequest.Password = &password
+	}
+
+	if updateUserRequest.Password != nil {
```
remyleone, line 364: Same here, introduce a variable and check this variable to ensure than an update should be performed

internal/services/mongodb/snapshot.go
```diff
@@ -0,0 +1,212 @@
...
+			Update:  schema.DefaultTimeout(defaultMongodbSnapshotTimeout),
+			Delete:  schema.DefaultTimeout(defaultMongodbSnapshotTimeout),
+			Default: schema.DefaultTimeout(defaultMongodbSnapshotTimeout),
+		},
+		Importer: &schema.ResourceImporter{
+			StateContext: schema.ImportStatePassthroughContext,
+		},
+		SchemaVersion: 0,
+		Schema: map[string]*schema.Schema{
+			"id": {
```
remyleone, line 35: Might be confusing with the SetId method. Why is this required?

internal/services/mongodb/snapshot.go
```diff
@@ -0,0 +1,212 @@
...
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	updateReq := &mongodb.UpdateSnapshotRequest{
+		SnapshotID: snapshotID,
+		Region:     region,
+	}
+
+	if d.HasChange("name") {
```
remyleone, line 167: Introduce variable to trigger a change if this boolean value changed. Like `has_changed`

internal/services/mongodb/instance.go
```diff
@@ -0,0 +1,405 @@
...
+			"node_type": {
+				Type:             schema.TypeString,
+				Required:         true,
+				Description:      "Type of node to use for the instance",
+				DiffSuppressFunc: dsf.IgnoreCase,
+			},
+			"user_name": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				ForceNew:    true,
```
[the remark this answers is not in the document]
jremy42: Yes, the user_name cannot be updated at the moment, so changing it will indeed trigger the recreation of the whole instance.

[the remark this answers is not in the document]
jremy42: doing nothing be a better solution?

internal/services/mongodb/snapshot.go
jremy42: You are right

docs/data-sources/mongodb_instance.md
```diff
@@ -0,0 +1,68 @@
...
+
+- `region` - (Defaults to [provider](../index.md#region) `region`) The [region](../guides/regions_and_zones.md#zones) in which the MongoDB® Instance exists.
+
+- `organization_id` - (Defaults to [provider](../index.md#organization_id) `organization_id`) The ID of the organization the MongoDB® instance is in.
+
+## Attributes Reference
+
+In addition to all above arguments, the following attributes are exported:
+
+- `id` - The ID of the MongoDB® Instance.
```
remyleone, line 50: Will this ID still be present?

internal/services/mongodb/instance.go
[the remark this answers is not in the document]
remyleone: Maybe having a dedicated resource for user, it is early but I would feel more comfortable with that rather than having a forcenew that could have large consequences if handled badly

docs/data-sources/mongodb_instance.md
jremy42: That's normal, id is always exported, right?

remyleone (approved):

PR #2781, opened 2024-10-23
fix(baremetal): change offer in tests

Laure-di:


internal/services/baremetal/offer_data_source_test.go
```diff
@@ -13,46 +13,55 @@ import (
 	"github.com/scaleway/terraform-provider-scaleway/v2/internal/services/baremetal"
 )
 
+const (
+	OfferName = "EM-B220E-NVME"
+	Region    = "fr-par-1"
```
Codelax, line 18:
```suggestion
	Zone    = "fr-par-1"
```
This will impact whole file, you probably want to do it manually in your IDE.

Codelax (approved):

PR #2872, opened 2025-01-14
feat(rdb): creating instance from snapshot

jremy42:


internal/services/rdb/snapshots.go
```diff
@@ -0,0 +1,230 @@
...
+			},
+			"updated_at": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "Expiration date of the snapshot in ISO 8601 format (RFC 3339).",
+			},
+			"node_type": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "The type of database instance you want to create",
```
remyleone, line 64: Should be renamed to focus more on the node_type the snapshot come from?

remyleone (dismissed):

remyleone (approved):

PR #2899, opened 2025-02-04
ci: bump go to 1.23

Codelax:


go.mod
```diff
@@ -2,7 +2,7 @@ module github.com/scaleway/terraform-provider-scaleway/v2
 
 go 1.22.0
```
remyleone, line 3:
```suggestion
go 1.23.5
```
?

.github/workflows/unit-tests.yaml
```diff
@@ -8,7 +8,7 @@ jobs:
   tests:
     strategy:
       matrix:
```
remyleone, line 10: Could it be simplified? it is a matrix with a single element

Codelax: Can we do this in a new PR ?

remyleone (approved):

PR #2927, opened 2025-02-19
docs: importing a secret is not correct

al26p:
The documentation was not correct. 

Importing a secret with `{region}/{id}` will result in the following: 

```
Error: cant parse localized id: fr-par/11111111-1111-1111-1111-111111111111
```

Turns out the expected format is `{region}/{id}/{version}`

### DIFF
--- a/docs/resources/secret.md
+++ b/docs/resources/secret.md
@@ -69,8 +69,8 @@ In addition to all arguments above, the following attributes are exported:
 
 ## Import
 
-This section explains how to import a secret using the `{region}/{id}` format.
+This section explains how to import a secret using the `{region}/{id}/{version}` format.
 
 ```bash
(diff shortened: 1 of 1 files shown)


remyleone (approved):

remyleone:
Thanks a lot for your contribution @al26p :)

remyleone:
That being said, are you sure you are not mixing up `scaleway_secret` and `scaleway_secret_version`? https://registry.terraform.io/providers/scaleway/scaleway/latest/docs/resources/secret_version#import

al26p:
Let me double check. It looks like it could be the case

al26p:
You are definitively right on this. I should have been more careful on this. 
Sorry for that and for loosing your time. (and good catch !)

Will definitively owe you a beer if we meet !

PR #2937, opened 2025-02-25
docs(baremetal): use UUID for offer instead of label name

Laure-di:


docs/resources/baremetal_server.md
```diff
@@ -18,7 +18,7 @@ data "scaleway_iam_ssh_key" "main" {
 
 resource "scaleway_baremetal_server" "base" {
   zone		  = "fr-par-2"
-  offer       = "GP-BM1-S"
+  offer       = "fd7605a7-7493-49ba-9019-0e9c2897d4a0"
```
remyleone, line 21: Could you provide an example with a data source ?

Laure-di: There are examples with datasource. Should I transform all examples with the data source on the document?

remyleone (approved):

PR #2979, opened 2025-03-12
feat(secret_manager): add list of secret versions in secret datasource

### LINKED ISSUE #2926: Add list of secret versions to scaleway_secret datasource
### Community Note

### Description

Upon retrieval of a SecretManager secret using the `scaleway_secret` datasource, there is an attribute `version_count` available, which can be used to retrieve multiple versions of a secret. However, this only works if no (intermediate) versions have been deleted, as then the `revision` is out of sync with the `version_count`.

It would be much nicer if the `scaleway_secret` datasource would return the complete list of available versions (like [this API call](https://www.scaleway.com/en/developers/api/secret-manager/#path-secret-versions-list-versions-of-a-secret-using-the-secrets-id)) in an attribute `versions`, instead of a `version_count` attribute, which need to be itterated/fetched manually.

* scaleway_secret

mmrqs:
We would like to add the list of secret_versions in the secret datasource.
Fix #2926


internal/services/secret/secret.go
```diff
@@ -192,19 +231,47 @@ func ResourceSecretRead(ctx context.Context, d *schema.ResourceData, m interface
 		_ = d.Set("tags", types.FlattenSliceString(secretResponse.Tags))
 	}
 
+	versions, err := api.ListSecretVersions(&secret.ListSecretVersionsRequest{
```
luxifer, line 234: This API call can be paginated, here we should loop over all the pages. I know that most of the secrets have only one version but this may not be the case every time.

mmrqs: Good remark, I've added the option : "scw.WithAllPages()" to request all the pages. Thanks !

internal/services/secret/secret.go
```diff
@@ -192,19 +231,47 @@ func ResourceSecretRead(ctx context.Context, d *schema.ResourceData, m interface
...
 	_ = d.Set("project_id", secretResponse.ProjectID)
 	_ = d.Set("path", secretResponse.Path)
 	_ = d.Set("protected", secretResponse.Protected)
 	_ = d.Set("ephemeral_policy", flattenEphemeralPolicy(secretResponse.EphemeralPolicy))
 	_ = d.Set("type", secretResponse.Type)
 
+	versionsList := make([]map[string]interface{}, 0, len(versions.Versions))
+	for _, version := range versions.Versions {
+		versionsList = append(versionsList, map[string]interface{}{
+			"revision":    strconv.Itoa(int(version.Revision)),
```
ojroques, line 264: Would be nice to also have the secret ID here.

mmrqs: Good remark, i've added it ! Thanks !

ojroques (approved):

jremy42 (approved):

PR #2985, opened 2025-03-18
docs(lb): clarify health check URI description

ahamez:
The `uri` parameter takes a path, not a full URI. Otherwise, the health check doesn't work and the servers are always reported as down.


remyleone (approved):

remyleone:
Thanks @ahamez for your contribution :)

PR #3034, opened 2025-04-14
fix(mongodb): delete pn during delete for fix test

jremy42:


internal/services/container/container_test.go
```diff
@@ -13,6 +13,8 @@ import (
 	containerchecks "github.com/scaleway/terraform-provider-scaleway/v2/internal/services/container/testfuncs"
 )
 
+const namespaceResourceType = "scaleway_container_namespace"
```
jremy42, line 16:
```suggestion
```

internal/services/container/container_test.go
```diff
@@ -13,6 +13,7 @@ import (
 	containerchecks "github.com/scaleway/terraform-provider-scaleway/v2/internal/services/container/testfuncs"
 )
 
+
```
jremy42, line 16:
```suggestion
```

PR #3035, opened 2025-04-14
feat(edge-services): support route and waf references on cache & tls stages

yfodil:


internal/services/container/container_test.go
```diff
@@ -575,7 +575,7 @@ func isContainerPresent(tt *acctest.TestTools, n string) resource.TestCheckFunc
 func isContainerDestroyed(tt *acctest.TestTools) resource.TestCheckFunc {
 	return func(state *terraform.State) error {
 		for _, rs := range state.RootModule().Resources {
-			if rs.Type != "scaleway_container_namespace" { //nolint:goconst
+			if rs.Type != "scaleway_container_namespace" {
```
Gnoale, line 578: why not use the constant here too ?

Gnoale (approved):

Gnoale (approved):

PR #3093, opened 2025-05-16
container: use project_id from provider configuration

### LINKED ISSUE #3068: container_namespace should infer project_id from provider but doesn't
### Terraform Version

```bash
Terraform v1.5.7
on darwin_arm64
+ provider registry.terraform.io/scaleway/scaleway v2.53.0
```

### Affected Resource(s)

- `scaleway_container_namespace`

```hcl
provider "scaleway" {
  #should be the ID of a non-default project
  project_id = "REDACTED"
  region     = "nl-ams"
  zone       = "nl-ams-1"
}

resource "scaleway_container_namespace" "foo" {
  name       = "foo"
}

```

### Expected Behavior
Create a Serverless Containers namespace in the project matching the ID from the provider.

### Actual Behavior
Creates a Serverless Containers namespace in the default project.

### Steps to Reproduce
1. Apply
2. Destroy
3. Update the `scaleway_container_namespace` resource, adding in the `project_id` argument
4. Apply again. It will now create the namespace in the correct place.

Gnoale:
As a user bring recently, in most of the modules, we ignore the `project_id` parameter set at the provider level in some cases, because the environment takes precedence over the provider config [here](https://github.com/scaleway/terraform-provider-scaleway/blob/ee4eaf3c1e02119c3c7ed37a572337fb999e9de3/internal/meta/meta.go#L225C83-L225C93)

partly fixes https://github.com/scaleway/terraform-provider-scaleway/issues/3068

The provider configuration should probably override other configuration sources but this could be impactful / breaking for users used to provide environment variables

The change imply to explicitly find a `project_id` at the resource / client level and raise an error if not found, instead of relying on the SDK logic to eventually find it or use an empty string  in some cases.

=> Now when running the test in CI, we need to inject a project ID in the [meta configuration](https://github.com/scaleway/terraform-provider-scaleway/pull/3093/files#diff-7a0b4bba956f7f07d7f1b0b1b3f83f28dda76464bf8a7bc0c98d406a1029efdbR45).

But this reveal another corner case: resources calling HTTP GET query with the optional `project_id` parameter that where not using it, [namespace container datasource](https://github.com/scaleway/terraform-provider-scaleway/blob/ee4eaf3c1e02119c3c7ed37a572337fb999e9de3/internal/services/container/testdata/data-source-namespace-basic.cassette.yaml#L71) for instance, are now using the fake project_id set in the test suite (`&project_id=1111111`).
So [we strip this parameter](https://github.com/scaleway/terraform-provider-scaleway/pull/3093/files#diff-7a0b4bba956f7f07d7f1b0b1b3f83f28dda76464bf8a7bc0c98d406a1029efdbR70) from the query to address these kind of cases

### DIFF
--- a/internal/services/account/project_data_source_test.go
+++ b/internal/services/account/project_data_source_test.go
@@ -95,30 +95,6 @@ func TestAccDataSourceProject_Default(t *testing.T) {
 	})
 }
 
-func TestAccDataSourceProject_Extract(t *testing.T) {
-	tt := acctest.NewTestTools(t)
-	defer tt.Cleanup()
-
-	projectID, projectIDExists := tt.Meta.ScwClient().GetDefaultProjectID()
-	if !projectIDExists {
-		t.Skip("no default project ID")
-	}
-
-	resource.ParallelTest(t, resource.TestCase{
-		PreCheck:          func() { acctest.PreCheck(t) },
-		ProviderFactories: tt.ProviderFactories,
-		Steps: []resource.TestStep{
-			{
-				Config: `data scaleway_account_project "project" {}`,
-				Check: resource.ComposeTestCheckFunc(
-					resource.TestCheckResourceAttr("data.scaleway_account_project.project", "id", projectID),
-					resource.TestCheckResourceAttrSet("data.scaleway_account_project.project", "name"),
-				),
-			},
-		},
-	})
-}
-
 func TestAccDataSourceProject_List(t *testing.T) {
 	tt := acctest.NewTestTools(t)
 	defer tt.Cleanup()
--- a/internal/services/container/container_data_source.go
+++ b/internal/services/container/container_data_source.go
@@ -9,6 +9,7 @@ import (
 	"github.com/scaleway/scaleway-sdk-go/scw"
 	"github.com/scaleway/terraform-provider-scaleway/v2/internal/datasource"
 	"github.com/scaleway/terraform-provider-scaleway/v2/internal/locality"
+	"github.com/scaleway/terraform-provider-scaleway/v2/internal/meta"
 	"github.com/scaleway/terraform-provider-scaleway/v2/internal/types"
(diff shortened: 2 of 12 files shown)


internal/services/container/container_data_source.go
```diff
@@ -58,11 +59,16 @@ func DataSourceContainerRead(ctx context.Context, d *schema.ResourceData, m inte
 	if !ok {
 		containerName := d.Get("name").(string)
 
+		projectID, _, err := meta.ExtractProjectID(d, m)
```
remyleone, line 62: Is there a way to ensure that it is taken into account with a test?

Gnoale: the test would break without the `ForceProjectID` in meta.Config

internal/services/account/project_data_source_test.go
```diff
@@ -95,30 +95,6 @@ func TestAccDataSourceProject_Default(t *testing.T) {
 	})
 }
 
-func TestAccDataSourceProject_Extract(t *testing.T) {
```
remyleone, line 98: Why is this test removed?

Gnoale: In CI it was always skipped because the projectID was not set
But now that it is forced to a dummy value, the query is not found

I could remove the skip part, re record it and use the   `acctest.NewTestToolsWithoutDefaultProjectID`

Gnoale: in fact that does not makes much sense because the projectID is totally dependent on the user default one, and we explicitly override that in the tests with `00000000-0000-0000-0000-000000000000`

`Error: scaleway-sdk-go: error executing request: Get "https://api.scaleway.com/account/v3/projects/00000000-0000-0000-0000-000000000000": Get "https://api.scaleway.com/account/v3/projects/00000000-0000-0000-0000-000000000000": requested interaction not found`

Gnoale:
Closing for now

PR #3094, opened 2025-05-19
fix(all): private_ip(s) should only be set when no error occurred

Mia-Cross:
This PR makes the `private_ip` or `private_ips` attributes optional as not every resource has private IPs.

It also adds the project ID as query parameter for fetching IPs from IPAM in order to force permissions checks. If unauthorized, the field will not be set (and a warning will be displayed).

This PR also prevents the Read functions to exit early if errors occur while reading private IPs. Warnings will be collected and displayed afterwards, but the feature should be non-blocking.


internal/services/redis/cluster.go
```diff
@@ -368,54 +369,72 @@ func ResourceClusterRead(ctx context.Context, d *schema.ResourceData, m interfac
...
-				diags = append(diags, diag.Diagnostic{
-					Severity:      diag.Warning,
-					Summary:       err.Error(),
-					Detail:        "Got 403 while reading private IP from IPAM API, please check your IAM permissions",
-					AttributePath: cty.GetAttrPath("private_ips"),
-				})
-			}
+				privateIPs, err := ipam.GetResourcePrivateIPs(ctx, m, region, opts)
+
+				switch {
```
Gnoale, line 406: If the logic to get theses private IP is similar, maybe you could factorize it in one function ?

Mia-Cross: I tried but the logic changes a little bit for each resource, and I would also need to pass a large amount of parameters to the factorized function in order to display precise error messages.

remyleone (approved):

PR #3100, opened 2025-05-21
fix(mongodb): enable public & private endpoints definition

jremy42:
…te doc


internal/services/mongodb/instance_test.go
```diff
@@ -373,6 +373,51 @@ func TestAccMongoDBInstance_UpdatePrivateNetwork(t *testing.T) {
...
+				  private_network {
+				    pn_id = scaleway_vpc_private_network.pn01.id
+				  }
+
+				  public_network {}
+				}
+				`,
+				Check: resource.ComposeTestCheckFunc(
+					isMongoDBInstancePresent(tt, "scaleway_mongodb_instance.main"),
+					resource.TestCheckResourceAttr("scaleway_mongodb_instance.main", "public_network.#", "1"),
```
remyleone, line 411: Could you add test that private network are also defined here?

remyleone (approved):

PR #3121, opened 2025-05-28
doc: format examples and relocate inference model datasource

coutantal:
Following hahiscorp guidelines, resource type and name should be wrapped in double quotes.
https://developer.hashicorp.com/terraform/language/style#code-style

`scaleway_inference_model` looks isolated due to a typo : 

image


Gnoale (approved):
Thanks for your contribution

PR #3138, opened 2025-06-03
feat(object): add content_type attribute

### LINKED ISSUE #2898: scaleway_object resource not setting the HTTP Content-Type header correctly
### Description

Currently, the scaleway_object resource only accepts a generic metadata map. When trying to set the HTML content type (for example, to render an HTML page in a bucket website), using:
```hcl
metadata = {
  content-type = "text/html"
}
```
results in the provider sending `x-amz-meta-content-type: text/html` instead of setting the actual HTTP Content-Type header. As a result, browsers receive a default Content-Type (such as `application/octet-stream`), which causes the HTML file to be downloaded rather than rendered.

In contrast, when using the Scaleway Console, adding a metadata key `content-type` with the value `text/html` properly sets the HTTP header on the object.

### Community Note

### Affected Resource(s)

* scaleway_object

```hcl
resource "scaleway_object" "main" {
  bucket     = scaleway_object_bucket.main.name
  key        = "index.html"
  file       = "index.html"
  visibility = "public-read"
  metadata = {
    content-type = "text/html"
  }
}
```

Laure-di:
close #2898


internal/services/object/object_test.go
```diff
@@ -161,7 +200,6 @@ func TestAccObject_Hash(t *testing.T) {
 				`, bucketName, objectTestsMainRegion),
 				Check: resource.ComposeTestCheckFunc(
 					objectchecks.CheckBucketExists(tt, "scaleway_object_bucket.base-01", true),
-					testAccCheckObjectExists(tt, "scaleway_object.file"),
```
remyleone: Why is this line removed?

PR #3140, opened 2025-06-04
fix(baremetal): remove non-idempotent condition

Laure-di:


internal/services/baremetal/testdata/data-source-offer-subscription-period-hourly.cassette.yaml
remyleone, line 1: Why is this cassettes recorded if there are no change in the test ?

Laure-di: The monthly cassette was removed as there were no changes. However, the hourly cassette was recorded because it did include changes

PR #3141, opened 2025-06-04
feat(file): add support v1beta1

Laure-di:


docs/resources/file_filesystem.md
```diff
@@ -0,0 +1,52 @@
...
+  associated with.
+- `organization_id` - (Defaults to [provider](../index.md#organization_id) `organization_id`) The ID of the organization the user is associated with.
+
+## Attributes Reference
+
+In addition to all arguments above, the following attributes are exported:
+
+- `id` - The ID of the filesystem.
+- `status` - The current status of the filesystem. Possible values include creating, available, etc.
+- `number_of_attechments` - The number of active attachments (mounts) on the filesystem.
```
remyleone:
```suggestion
- `number_of_attachments` - The number of active attachments (mounts) on the filesystem.
```

docs/resources/file_filesystem.md
```diff
@@ -0,0 +1,52 @@
...
+
+### Basic
+
+```terraform
+resource scaleway_file_filesystem file {
+  name = "my-nfs-filesystem"
+  size = 100000000000 # 100 GB
+}
+```
+
```
remyleone, line 22: Could you add an example that would show case how to have a filesystem with an instance interconnected? At the moment I don't see how to mount and use the file system. 
Also how does it work from a security perspective?

Laure-di: I’m currently waiting for the team to allow an attribute in public so I can finish my PR for the integration with instance.

Could you clarify your last question?

remyleone (approved):

PR #3142, opened 2025-06-05
fix(iam): handle http 409 transient error

Gnoale:
* exponentialy retry Add and Remove group member requests

### DIFF
--- a/internal/services/iam/group_membership.go
+++ b/internal/services/iam/group_membership.go
@@ -4,12 +4,14 @@ import (
 	"context"
 	"fmt"
 	"strings"
+	"time"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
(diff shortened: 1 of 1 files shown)


internal/services/iam/group_membership.go
```diff
@@ -178,3 +180,59 @@ func ExpandGroupMembershipID(id string) (groupID string, userID string, applicat
...
+		}
+
+		return nil, fmt.Errorf("failed to remove group member after %d retries", maxRetries)
+
+	default:
+		return nil, fmt.Errorf("invalid request type: %T", req)
+	}
+}
+
+func handleTransientError(err error, groupID string, retryInterval time.Duration, maxRetries int) bool {
```
remyleone: Could `transport.RetryOnTransientStateError` be used here?

remyleone (approved):

PR #3143, opened 2025-06-05
feat(vpc): add support for ResourceIdentity

yfodil:


internal/services/vpc/vpc.go
```diff
@@ -21,7 +23,56 @@ func ResourceVPC() *schema.Resource {
 		UpdateContext: ResourceVPCUpdate,
 		DeleteContext: ResourceVPCDelete,
 		Importer: &schema.ResourceImporter{
-			StateContext: schema.ImportStatePassthroughContext,
+			StateContext: func(
```
jremy42: Isn't there a way to make a global function, since it'll basically be the same across all resources, right?

internal/locality/regional/ids.go
```diff
@@ -81,3 +83,50 @@ func ParseID(regionalID string) (region scw.Region, id string, err error) {
 
 	return
 }
+
+func ResolveRegionAndID(
+	d *schema.ResourceData,
+	fallbackDefaultRegion func(*schema.ResourceData) (scw.Region, error),
+) (scw.Region, string, error) {
+	if identity, err := d.Identity(); err == nil && identity != nil {
```
remyleone: Could you have a more flatten structure with returning early in case of an error? It would be helpful to generate logs here using tflog and passing the context.

internal/locality/regional/ids.go
```diff
@@ -81,3 +85,67 @@ func ParseID(regionalID string) (region scw.Region, id string, err error) {
...
+
+func ResolveRegionAndID(
+	ctx context.Context,
+	d *schema.ResourceData,
+	fallbackDefaultRegion func(*schema.ResourceData) (scw.Region, error),
+) (scw.Region, string, error) {
+	identity, err := d.Identity()
+	if err != nil {
+		tflog.Warn(ctx, fmt.Sprintf("failed to read identity from ResourceData: %v", err))
+	} else if identity != nil {
```
remyleone, line 97: Could you flatten more and resolve early in case : 
- identity is nil, 
- identity does not contain id and resolve with a specific diagnostic
- identity does not contain region and resolve with a specific diagnostic

PR #3144, opened 2025-06-06
feat(meta): warn the user if there are multiple variable sources used

Gnoale:
```
│ Warning: Multiple credential sources detected
│
│   with provider["registry.terraform.io/scaleway/scaleway"],
│   on main.tf line 21, in provider "scaleway":
│   21: provider "scaleway" {
│
│ Variable	AvailableSources						Using
│ SCW_ACCESS_KEY	Profile defined in provider{} block, Environment variable	Environment variable
│
```


internal/meta/meta.go
```diff
@@ -73,6 +77,41 @@ func (m Meta) ZoneSource() string {
...
+			if len(values) > 1 {
+				fmt.Fprintf(w, "%s\t%s\t%s\n", variable, strings.Join(values, ", "), values[len(values)-1]) //nolint:errcheck
+
+				multiple = true
+			}
+		}
+	}
+
+	if err := w.Flush(); err != nil {
+		panic(err) // lintignore: R009
```
remyleone: I'm not ok to add a panic statement here, could you add diags and fail later on in a different way? I do not want a panic to occur in the provider.

Gnoale: Yes I removed the "nolint" and return an error

internal/provider/provider.go
```diff
@@ -357,7 +357,28 @@ func Provider(config *Config) plugin.ProviderFunc {
 				return nil, diag.FromErr(err)
 			}
 
-			return m, nil
+			var diags diag.Diagnostics
+
+			ok, message, err := m.HasMultipleVariableSources()
+			if err != nil {
+				diags = append(diags, diag.Diagnostic{
+					Severity: diag.Error,
```
remyleone: Could it be a warning and not an error? I would not like to break existing deployment that work. I would like just to warn them. Why do you think an error is necessary here?

Gnoale: because this is in case the tab formatting return an error with `fmt` or buffer.flush() for instance  

I think this is weird to display anything to the user in this particular case because it would bring even more confusion I suppose 

that's why in the first place i didn't want to handle any error, but I agree it could be a warn, it should never happens anyway 🤞

remyleone (approved):

PR #3153, opened 2025-06-11
feat(lb): add private network resource

yfodil:


internal/services/lb/private_network.go
```diff
@@ -0,0 +1,210 @@
...
+				ForceNew:    true,
+				Description: "The load-balancer ID to attach the private network to",
+			},
+			"private_network_id": {
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+				Description: "The private network ID to attach",
+			},
+			"ipam_ip_ids": {
```
remyleone, line 46: Why is it plural if I can only have one?

internal/services/lb/private_network.go
```diff
@@ -0,0 +1,210 @@
...
+		return diag.FromErr(err)
+	}
+
+	_ = d.Set("private_network_id", regional.NewIDString(region, foundPN.PrivateNetworkID))
+	_ = d.Set("lb_id", zonal.NewIDString(zone, foundPN.LB.ID))
+	_ = d.Set("ipam_ip_ids", regional.NewIDStrings(region, foundPN.IpamIDs))
+	_ = d.Set("status", foundPN.Status.String())
+	_ = d.Set("created_at", types.FlattenTime(foundPN.CreatedAt))
+	_ = d.Set("updated_at", types.FlattenTime(foundPN.UpdatedAt))
+	_ = d.Set("zone", zone)
```
remyleone: Could you use foundPN.Zone here if available?

internal/services/lb/private_network.go
yfodil: It's also plural on the API side, which suggests that it may eventually support multiple IPs in the future

remyleone (dismissed):

internal/services/lb/private_network_test.go
```diff
@@ -0,0 +1,63 @@
...
+	"github.com/scaleway/terraform-provider-scaleway/v2/internal/acctest"
+)
+
+func TestAccLBPrivateNetwork_Basic(t *testing.T) {
+	tt := acctest.NewTestTools(t)
+	defer tt.Cleanup()
+
+	resource.ParallelTest(t, resource.TestCase{
+		ProtoV6ProviderFactories: tt.ProviderFactories,
+		CheckDestroy:             isLbDestroyed(tt),
```
remyleone: Should we also check for the destruction of the VPC and the private network ?

PR #3158, opened 2025-06-12
feat(container): vpc integration

Mia-Cross:


internal/services/container/testdata/namespace-vpc-integration.cassette.yaml
```diff
@@ -0,0 +1,2904 @@
...
+                - Scaleway API Gateway (fr-par-2;edge02)
+            Strict-Transport-Security:
+                - max-age=63072000
+            X-Content-Type-Options:
+                - nosniff
+            X-Frame-Options:
+                - DENY
+            X-Request-Id:
+                - d3b7c8ba-8e96-476b-970e-27ba92704224
+        status: 400 Bad Request
```
remyleone, line 940: I'm not sure we should merge that. This request does not look ok

internal/acctest/validate_cassettes_test.go
```diff
@@ -29,6 +29,7 @@ func exceptionsCassettesCases() map[string]struct{} {
 		"../services/secret/testdata/secret-version-type.cassette.yaml":                      {},
 		"../services/file/testdata/file-system-invalid-size-granularity-fails.cassette.yaml": {},
 		"../services/file/testdata/file-system-size-too-small-fails.cassette.yaml":           {},
+		"../services/container/testdata/namespace-vpc-integration.cassette.yaml":             {},
```
remyleone, line 32: I'm not sure we should add an exception here.

internal/services/container/testdata/namespace-vpc-integration.cassette.yaml
Mia-Cross: I wanted to test the fact that legacy namespaces are not compatible with VPC and that activating the feature in the namespace should recreate it with ForceNew, therefore changing its ID.
I can remove the 2nd step if we prefer not to test whether the error occurs, but only the resource recreation.

remyleone (approved):

remyleone (approved):

PR #3160, opened 2025-06-13
feat(function): vpc integration

Mia-Cross:


internal/services/function/testdata/function-namespace-vpc-integration.cassette.yaml
```diff
@@ -0,0 +1,2708 @@
...
+                - Scaleway API Gateway (fr-par-2;edge01)
+            Strict-Transport-Security:
+                - max-age=63072000
+            X-Content-Type-Options:
+                - nosniff
+            X-Frame-Options:
+                - DENY
+            X-Request-Id:
+                - de7f23b5-6a57-4b20-ae67-e329b161e02d
+        status: 400 Bad Request
```
remyleone, line 1136: Is it normal to have a bad request here?

Mia-Cross: I wanted to test the fact that legacy namespaces are not compatible with VPC and that activating the feature in the namespace should recreate it with ForceNew, therefore changing its ID. 
I can remove the 2nd step if we prefer not to test whether the error occurs, but only the resource recreation.

remyleone (approved):

PR #3198, opened 2025-07-03
feat(keymanager): add Key Manager Key resource

jremy42:


internal/services/keymanager/helpers.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	if u.AsymmetricSigning != nil {
+		return "asymmetric_signing"
+	}
+
+	return ""
+}
+
+func TimeToRFC3339(t *time.Time) string {
```
remyleone: Why not using a types.FlattenTime ?

internal/services/keymanager/helpers.go
```diff
@@ -0,0 +1,77 @@
...
+	"fmt"
+	"strings"
+	"time"
+
+	key_manager "github.com/scaleway/scaleway-sdk-go/api/key_manager/v1alpha1"
+	"github.com/scaleway/scaleway-sdk-go/scw"
+	"github.com/scaleway/terraform-provider-scaleway/v2/internal/meta"
+)
+
+func ExpandStringList(v any) []string {
```
remyleone: What is the difference with `types.ExpandStrings`

internal/services/keymanager/key_resource.go
```diff
@@ -0,0 +1,270 @@
...
+				}, false),
+			},
+			// Computed fields
+			"id":             {Type: schema.TypeString, Computed: true},
+			"state":          {Type: schema.TypeString, Computed: true},
+			"created_at":     {Type: schema.TypeString, Computed: true},
+			"updated_at":     {Type: schema.TypeString, Computed: true},
+			"rotation_count": {Type: schema.TypeInt, Computed: true},
+			"protected":      {Type: schema.TypeBool, Computed: true},
+			"locked":         {Type: schema.TypeBool, Computed: true},
```
remyleone: could you add description?

internal/services/keymanager/key_resource.go
```diff
@@ -0,0 +1,270 @@
...
+			"rotated_at":     {Type: schema.TypeString, Computed: true},
+			"origin_read":    {Type: schema.TypeString, Computed: true},
+			"region_read":    {Type: schema.TypeString, Computed: true},
```
remyleone: What is this field?

remyleone (changes requested):

internal/services/keymanager/helpers.go
```diff
@@ -0,0 +1,52 @@
...
+import (
+	"fmt"
+	"strings"
+
+	key_manager "github.com/scaleway/scaleway-sdk-go/api/key_manager/v1alpha1"
+	"github.com/scaleway/scaleway-sdk-go/scw"
+	"github.com/scaleway/terraform-provider-scaleway/v2/internal/meta"
+)
+
+func UsageToString(u *key_manager.KeyUsage) string {
```
remyleone, line 14: Why is this required when the enum got a .String function?

internal/services/keymanager/key_resource.go
```diff
@@ -0,0 +1,273 @@
...
+		Unprotected:    unprotected,
+		Origin:         key_manager.KeyOrigin(origin),
+	}
+
+	key, err := client.CreateKey(createReq)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	d.SetId(fmt.Sprintf("%s/%s", key.Region, key.ID))
```
remyleone: Why not using a regional.NewIDString ?

remyleone (changes requested):

internal/services/keymanager/helpers.go
remyleone: Could this util be moved to the SDK? It seems that it might get useful for other tools as well

jremy42: So with reflect then? Because the function is very contextual

jremy42: At the limit, in protocgen, generate a function that converts the oneof

internal/services/keymanager/helpers.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	if u.AsymmetricSigning != nil {
+		return "asymmetric_signing"
+	}
+
+	return ""
+}
+
+func ExtractRegionAndKeyID(id string) (scw.Region, string, error) {
```
remyleone: Why not use `regional.ParseID` ?

remyleone (approved):

PR #3216, opened 2025-07-16
feat(lb): add support for enable_access_logs

yfodil:


internal/services/lb/data_source_lb_frontends.go
```diff
@@ -74,6 +74,14 @@ func DataSourceFrontends() *schema.Resource {
 							Computed: true,
 							Type:     schema.TypeBool,
 						},
+						"connection_rate_limit": {
+							Computed: true,
```
remyleone: Could you add a description with the unit expected here?

internal/services/lb/data_source_lb_frontends.go
```diff
@@ -74,6 +74,14 @@ func DataSourceFrontends() *schema.Resource {
...
+							Type:     schema.TypeInt,
+						},
+						"enable_access_logs": {
+							Computed: true,
+							Type:     schema.TypeBool,
```
remyleone: Could you add a description here?

remyleone (approved):

Issue #3219, opened 2025-07-17
Expose TLS certificate for scaleway_mongodb_instance as output or via data source


GeryDeKocliko:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

When creating a managed MongoDB instance using scaleway_mongodb_instance, Scaleway automatically generates a TLS certificate for secure connections. However, this certificate is not currently exposed via outputs nor accessible through a data source.
It would be extremely beneficial for build pipelines and automated workflows if the certificate could be retrieved directly via:
	•	A tls_certificate attribute in the `scaleway_mongodb_instance` resource, or
	•	A new data source such as `data.scaleway_mongodb_instance_tls_certificate`

This would enable fully automated retrieval of the certificate for CI/CD deployments, scripting, and Terraform module integration.

### New or Affected Resource(s)
	•	scaleway_mongodb_instance
	•	Potentially a new data source: data.scaleway_mongodb_instance_tls_certificate

### Potential Terraform Configuration

Option A: Expose via resource attribute & output

```hcl
resource "scaleway_mongodb_instance" "db" {
  name    = "my-mongo"
  engine  = "4.4"
  version = "4.4.6"
  zone    = "fr-par-1"
  # …additional settings
}

output "mongo_tls_cert" {
  value       = scaleway_mongodb_instance.db.tls_certificate
  description = "PEM-encoded TLS certificate for MongoDB"
}
```

Option B: New data source to fetch the certificate

```hcl
data "scaleway_mongodb_instance_tls_certificate" "db_cert" {
  instance_id = scaleway_mongodb_instance.db.id
}

resource "local_file" "mongo_cert" {
  content  = data.scaleway_mongodb_instance_tls_certificate.db_cert.tls_pem
  filename = "mongo-ca.pem"
}
```

## Use Case

Currently, users must:
	1.	Log in to the Scaleway Console, navigate to their MongoDB instance, and manually download the TLS certificate
	2.	Securely store the certificate
	3.	Create scripts or external tooling to integrate the certificate with the deployment process

[...]

PR #3228, opened 2025-07-22
feat(mongodb): add TLS certificate

jremy42:
close [#3219](https://github.com/scaleway/terraform-provider-scaleway/issues/3219)


internal/services/mongodb/instance.go
```diff
@@ -233,6 +236,11 @@ func ResourceInstance() *schema.Resource {
 			// Common
 			"region":     regional.Schema(),
 			"project_id": account.ProjectIDSchema(),
+			"tls_certificate": {
```
remyleone, line 237: Could you add documentation ?

jremy42: done

internal/services/mongodb/instance.go
```diff
@@ -460,6 +467,20 @@ func ResourceInstanceRead(ctx context.Context, d *schema.ResourceData, m any) di
...
+
+	var buf bytes.Buffer
+
+	if err == nil && cert != nil {
+		_, copyErr := io.Copy(&buf, cert.Content)
+		if copyErr == nil {
+			_ = d.Set("tls_certificate", buf.String())
+		}
+	}
+
```
yfodil: Errors are checked but silently ignored, if the certificate fetch or read fails, no diagnostic or return occurs.

internal/services/mongodb/instance.go
```diff
@@ -460,6 +467,20 @@ func ResourceInstanceRead(ctx context.Context, d *schema.ResourceData, m any) di
 
 	_ = d.Set("settings", map[string]string{})
 
+	cert, err := mongodbAPI.GetInstanceCertificate(&mongodb.GetInstanceCertificateRequest{
+		Region:     region,
+		InstanceID: ID,
+	}, scw.WithContext(ctx))
+
+	var buf bytes.Buffer
```
yfodil: Nit: Consider using `io.ReadAll` instead of `io.Copy` into a buffer

internal/services/mongodb/instance.go
```diff
@@ -460,6 +466,18 @@ func ResourceInstanceRead(ctx context.Context, d *schema.ResourceData, m any) di
...
 	_ = d.Set("settings", map[string]string{})
 
+	cert, err := mongodbAPI.GetInstanceCertificate(&mongodb.GetInstanceCertificateRequest{
+		Region:     region,
+		InstanceID: ID,
+	}, scw.WithContext(ctx))
+
+	if err == nil && cert != nil {
+		certBytes, readErr := io.ReadAll(cert.Content)
+		if readErr == nil {
```
remyleone, line 476: Could you add a diagnostic here in case readErr is not nil?

jremy42: done

PR #3229, opened 2025-07-24
doc(backend_guide): add documentation for scaleway object storage as tfstate backend

tcarecolin:
Hello, as a Scaleway user i encountered an issue when i attempted to use S3 backend not with AWS's3 but with Scaleway object-storage bucket
i had to find a workaround to make Scaleway credentials be correctly read by terraform

the issue is : by default Terraform backend S3 will read aws shared config file 
>~/$HOME/.aws/credentials

instead of using the correct scw config file

>~/$HOME/.config/scw/config.yaml

 i thought usefull for other people facing the same issue to complete the backend guide with a chatper on "S3 Backend"

### DIFF
--- a/docs/guides/backend_guide.md
+++ b/docs/guides/backend_guide.md
@@ -246,9 +246,9 @@ profiles:
     insecure: false
 ```
 
-actualy terraform backend "s3" is not aware of any other kind of s3 compatible bucket and is by default assuming you ar using aws's S3 service
+actually terraform backend "s3" is not aware of any other kind of s3 compatible bucket and is by default assuming you ar using aws's S3 service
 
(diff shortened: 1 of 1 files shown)


Laure-di:
Hello @tcarecolin,

Thank you so much for your contribution!

The documentation for the S3 backend already exists here:
https://registry.terraform.io/providers/scaleway/scaleway/latest/docs#store-terraform-state-on-scaleway-s3-compatible-object-storage

That said, I completely agree that having backend-related information in multiple places — or worse, missing entirely from where users expect it — can be confusing. I’ll fix that by consolidating the content and adding more details about credential usage.

Thanks again for highlighting this!

PR #3231, opened 2025-07-24
test(mongodb):  regroup tests

jremy42:
…ed tests


internal/services/mongodb/instance_test.go
```diff
@@ -13,7 +13,7 @@ import (
 	"github.com/scaleway/terraform-provider-scaleway/v2/internal/services/mongodb"
 )
 
-func TestAccMongoDBInstance_Basic(t *testing.T) {
+func TestAccMongoDBInstance_CombinedBasicUpdate(t *testing.T) {
```
remyleone: I don't think a change of name in necessary

remyleone (commented):
LGTM with a nit

remyleone (changes requested):
Could you also add an import test step to ensure it is working properly?

PR #3251, opened 2025-08-04
feat(baremetal): support protected server

### LINKED ISSUE #3070: Elastic Metal: Add protected feature for Elastic Metal
Similarly to the recently added `protected` option for Virtual Instances (#3022), I believe it would be useful to offer the same functionality for Elastic Metal servers.

However, unlike the situation in issue #3022, I haven't found any "protected" option for Elastic Metal in the Scaleway console.
Therefore, I'm aware that this issue doesn't exclusively concern the Terraform Provider Scaleway, but perhaps a Scaleway developer could create sub-issues for the internal teams.

This feature would improve the security of Elastic Metal resources by preventing their accidental deletion, as is already possible for Virtual Instances.

Do you think this feature could be implemented in the short or medium term?

Laure-di:
close #3070


internal/services/baremetal/server.go
```diff
@@ -148,6 +148,12 @@ If this behaviour is wanted, please set 'reinstall_on_ssh_key_changes' argument
 				ValidateFunc: validation.StringLenBetween(0, 255),
 				Description:  "Some description to associate to the server, max 255 characters",
 			},
+			"protected": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Default:     false,
+				Description: "If true, the baremetak server is protected against accidental deletion via the Scaleway API.",
```
yfodil:
```suggestion
				Description: "If true, the baremetal server is protected against accidental deletion via the Scaleway API",
```

internal/services/baremetal/server.go
```diff
@@ -148,6 +148,12 @@ If this behaviour is wanted, please set 'reinstall_on_ssh_key_changes' argument
 				ValidateFunc: validation.StringLenBetween(0, 255),
 				Description:  "Some description to associate to the server, max 255 characters",
 			},
+			"protected": {
```
yfodil, line 151: Can you please add documentation to `baremetal_server.md`

yfodil (approved):

PR #3281, opened 2025-08-26
docs(cockpit): mark scaleway_cockpit resource as deprecated

jremy42:


benjick:
Should this be Jan 1st **2026**?

PR #3322, opened 2025-09-09
chore: add mux to support terraform framework

remyleone:


internal/meta/extractors.go
```diff
@@ -18,12 +19,35 @@ type terraformResourceData interface {
...
+	GetRawConfig() cty.Value
+	GetRawConfigAt(valPath cty.Path) (cty.Value, diag.Diagnostics)
+}
+
+var (
+	_ terraformResourceData = (*schema.ResourceData)(nil)
+	_ terraformResourceData = (*schema.ResourceDiff)(nil)
+)
+
+func ExtractRawConfigString(d terraformResourceData, field string) (string, diag.Diagnostics) {
```
yfodil: I think that this existing function can do the same trick https://github.com/scaleway/terraform-provider-scaleway/blob/master/internal/meta/extractors.go#L162

remyleone: done

yfodil (approved):

PR #3323, opened 2025-09-09
chore: remove computed from locality schemas

remyleone:


internal/locality/regional/schemas.go
```diff
@@ -31,7 +31,11 @@ func Schema() *schema.Schema {
...
 		Optional:         true,
 		ForceNew:         true,
-		Computed:         true,
 		ValidateDiagFunc: locality.ValidateStringInSliceWithWarning(allRegions(), "region"),
+		DiffSuppressFunc: suppressSDKNullAssignment,
 	}
 }
+
+func suppressSDKNullAssignment(k, old, new string, d *schema.ResourceData) bool {
+	return new == "" && old != ""
```
jremy42, line 40:
```suggestion
	return (new == "" && old != "") || (old == "" && new != "")
```

PR #3387, opened 2025-10-02
fix(ipam): only suppress diffs when IP host part is identical

yfodil:


internal/services/ipam/ip.go
```diff
@@ -38,7 +38,7 @@ func ResourceIP() *schema.Resource {
 				ForceNew:         true,
 				Description:      "Request a specific IP in the requested source pool",
 				ValidateFunc:     validation.IsIPAddress,
-				DiffSuppressFunc: diffSuppressFuncStandaloneIPandCIDR,
+				DiffSuppressFunc: DiffSuppressFuncStandaloneIPandCIDR,
```
remyleone: maybe it could be migrated to dsf to be reused with other networking products

PR #3390, opened 2025-10-06
feat: add support for instance action

remyleone:
Requires https://github.com/scaleway/terraform-provider-scaleway/issues/3427 to have proper testing

### DIFF
--- a/internal/services/instance/action_server_action_test.go
+++ b/internal/services/instance/action_server_action_test.go
@@ -2,7 +2,6 @@ package instance_test
 
 import (
 	"errors"
-	"fmt"
 	"regexp"
 	"strings"
 	"testing"
@@ -90,7 +89,7 @@ func TestAccActionServer_Basic(t *testing.T) {
 							}
 						}
 
-						return fmt.Errorf("did not found the reboot event")
+						return errors.New("did not found the reboot event")
 					},
 				),
 			},


internal/services/instance/action_server_action_test.go
```diff
@@ -0,0 +1,43 @@
...
+					  lifecycle {
+						action_trigger {
+						  events  = [after_create]
+						  actions = [action.scaleway_instance_server_action.main]
+						}
+					  }
+					}
+
+					action "scaleway_instance_server_action" "main" {
+						config {
```
bbasata, line 41:
```suggestion
						config {
						  action = "reboot"
						
```

remyleone: @bbasata I've even included a test to verify that an incorrect verb is refused and it is not the case :/ I'm not sure to understand how to get started with this test

remyleone: @bbasata do you have examples of implementation and testing with actions? I'm not sure to understand how to get started. When I put a breakpoint on the invoke method, it does not get triggered

remyleone: @bbasata Also I think it would be helpful to have terraform in 1.13 or previous version to throw an error in case the lifecycle is configured with `action_trigger`. Because this keyword is not known in 1.13 I expect to have an error if I have an incorrect lifecycle attribute.

bbasata: > Also I think it would be helpful to have terraform in 1.13 or previous version to throw an error in case the lifecycle is configured with `action_trigger`. Because this keyword is not known in 1.13 I expect to have an error if I have an incorrect lifecycle attribute.

@remyleone: I agree, I would expect an error from Terraform 1.13 and earlier. I will confirm.

bbasata: > @remyleone: I agree, I would expect an error from Terraform 1.13 and earlier. I will confirm.

I've confirmed that I see an error in `go test` when I try actions syntax with earlier Terraform versions:

```console
 $ TF_ACC=1 TF_ACC_TERRAFORM_PATH=$HOME/bin/terraform_1.11.0 go test ./internal/cloudlite/machines -count=1
--- FAIL: TestRestart (0.15s)
    restart_test.go:13: Step 1/1 error: Error running pre-apply plan: exit status 1

        Error: Unsupported block type

          on terraform_plugin_test.tf line 14, in resource "terraform_data" "x":
          14:                                                                           action_trigger {

        Blocks of type "action_trigger" are not expected here.

        Error: Unsupported block type

          on terraform_plugin_test.tf line 21:
          21:                                           action "cloudlite_machine_restart" "power_cycle" {

        Blocks of type "action" are not expected here.
    panic.go:615: Error retrieving state, there may be dangling resources: exit status 1

        Error: Unsupported block type

          on terraform_plugin_test.tf line 14, in resource "terraform_data" "x":
          14:                                                                           action_trigger {

        Blocks of type "action_trigger" are not expected here.

        Error: Unsupported block type

          on terraform_plugin_test.tf line 21:
          21:                                           action "cloudlite_machine_restart" "power_cycle" {

        Blocks of type "action" are not expected here.
FAIL
FAIL    github.com/bbasata/terraform-provider-cloudlite/internal/cloudlite/machines     0.477s
FAIL
```

bbasata: > do you have examples of implementation and testing with actions?

For implementation: this is a recent blog post from a teammate: https://danielmschmidt.de/posts/2025-09-26-writing-a-terraform-action/

provider/framework.go
```diff
@@ -5,27 +5,47 @@ import (
...
 	"github.com/hashicorp/terraform-plugin-framework/resource"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/scaleway/terraform-provider-scaleway/v2/internal/meta"
+	"github.com/scaleway/terraform-provider-scaleway/v2/internal/services/instance"
 )
 
-var _ provider.Provider = &ScalewayProvider{}
+var (
+	_ provider.Provider            = &ScalewayProvider{}
+	_ provider.ProviderWithActions = (*ScalewayProvider)(nil)
```
estellesoulard, line 21: I might be missing something but I don't see any calls for this var? (could be some golang pattern I'm missing)

internal/services/instance/action_server_action_test.go
```diff
@@ -0,0 +1,126 @@
...
+							return fmt.Errorf("could not parse events.# as integer: %w", err)
+						}
+
+						if count < 1 {
+							return fmt.Errorf("expected events count > 1, got %d", count)
+						}
+
+						return nil
+					},
+				),
```
estellesoulard, line 94: - Shouldn't the check be `if count < 2`? At this point we expect a `poweron` followed by a `reboot` (and 2 matching events).
- Add check on the latest event's `request_body` matching `"action": "reboot"`
- Maybe add a PreConfig wait to this test (like [this one](https://github.com/scaleway/terraform-provider-scaleway/blob/1d2fc00402f86f84df3b2547773348ffacd10c71/internal/services/audittrail/event_data_source_test.go#L44)) to wait for events to cleanly pop up and secure the test results?

provider/framework.go
remyleone: It is completely normal. It is the standard way to prove that a type implements an interface :) https://developer.hashicorp.com/terraform/plugin/framework/actions/implementation#add-action-to-provider

internal/services/instance/action_server_action_test.go
remyleone: I've fixed the checks and now they are actively looking for a reboot event

internal/services/instance/action_server_action_test.go
```diff
@@ -80,18 +80,17 @@ func TestAccActionServer_Basic(t *testing.T) {
 							return errors.New("not found: data.scaleway_audit_trail_event.instance")
 						}
 
-						countStr := rs.Primary.Attributes["events.#"]
-
-						count, err := strconv.Atoi(countStr)
-						if err != nil {
-							return fmt.Errorf("could not parse events.# as integer: %w", err)
-						}
+						for key, value := range rs.Primary.Attributes {
```
estellesoulard, line 82: the request_body to check is [in each event ](https://github.com/scaleway/terraform-provider-scaleway/blob/master/docs/data-sources/audit_trail_event.md) , i think you might need something like
`for key, value := range rs.Primary.Attributes["events"] {` to iterate on each event

Issue #2870, opened 2025-01-10
RDB improve update version management


Laure-di:
The actual behaviour is to force_new a new instance. It would be interesting to clone the instance like the console does.

PR #3408, opened 2025-10-16
feat(RDB): implement RDB engine upgrade using MajorUpgradeWorkflow

jremy42:
close [#2870](https://github.com/scaleway/terraform-provider-scaleway/issues/2870)


internal/services/rdb/instance.go
```diff
@@ -911,21 +951,86 @@ func ResourceRdbInstanceUpdate(ctx context.Context, d *schema.ResourceData, m an
...
+	if d.HasChange("engine") {
+		oldEngine, newEngine := d.GetChange("engine")
+		newEngineStr := newEngine.(string)
+
+		targetVersionID := ""
+
+		var availableVersions []string
+		for _, version := range rdbInstance.UpgradableVersion {
+			availableVersions = append(availableVersions, version.Name)
+			if version.ID == newEngineStr || version.Version == newEngineStr || version.Name == newEngineStr {
```
estellesoulard: If my understanding is correct, l.62 'engine' is described as 'Database's engine version id.' but it seems here it could either be an id, version or name. If we keep this possibility, shouldn't it be reflected in the description? 
Regardless, from a Terraform standpoint, it would seem coherent to me for the 'engine' to be expected to be version.Name in this context (expecially since we list available version names in case of errors). Let me know what you think!

PR #3426, opened 2025-10-27
tests(instance): bump go-vcr to v4 for instance package

Mia-Cross:
This PR starts the migration of tests cassettes from `go-vcr.v3` to `go-vcr.v4` with the instance package.
Both versions of go-vcr can coexist in the same go.mod so we decided to migrate one package at a time.

This migration will allow us to have smaller cassettes thanks to the `omitempty` tag that was added to the Request and Response structs in v4, so empty fields like `transfer_encoding` and `trailer` will not appear in the cassettes.


internal/acctest/acctest.go
```diff
@@ -25,6 +33,75 @@ type TestTools struct {
...
+			return true
+		}
+	}
+
+	return false
+}
+
+// s3Encoder encodes binary payloads as base64 because serialization changed on go-vcr.v4
+func s3Encoder(i *cassette.Interaction) error {
+	if strings.HasSuffix(i.Request.Host, "scw.cloud") {
```
remyleone: Could you have an early return pattern here to avoid having nested if?

Issue #2762, opened 2024-10-04
Privilege RDB "readwrite" switch to custom after a certain amount of time


SebUndefined:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform Version

```console
terraform -v
Terraform v1.6.5
on darwin_arm64

Your version of Terraform is out of date! The latest version
is 1.9.7. You can update by downloading from https://www.terraform.io/downloads.html
```

### Affected Resource(s)

* scaleway_rdb_privilege

### Terraform Configuration Files

```hcl
resource "scaleway_rdb_privilege" "my_privilege" {
  instance_id   = MY_ID
  user_name     = USERNAME
  database_name = DB_NAME
  permission    = "readwrite"
}
```

### Actual Behavior

The privilege is actually well defined if we look at the scw console to "Permissions (Read/Write)" on the "DB_NAME" for a few minutes. After some time, it switch to "custom" in the console without any reason and cause a  permission denied error if we try to query or insert an element in some table. 

### Steps to Reproduce

Deploy the resource

```hcl
resource "scaleway_rdb_privilege" "my_privilege" {
  instance_id   = MY_ID
  user_name     = USERNAME
  database_name = DB_NAME
  permission    = "readwrite"
}
```

Run `terraform apply`

Wait some minutes/seconds and see the privileges switching to custom

Note, if we assign back the privilege to "Permissions (Read/Write)" manually (scw console), it seems to be ok and not going back to custom. 

### Important Factoids

None

SebUndefined:
ok sorry, it is apparently due to the table creation I did 
If I refer to this [link](https://www.scaleway.com/en/docs/managed-databases/postgresql-and-mysql/how-to/manage-permissions/#set-permissions-for-a-new-user) 
```
Permissions are applied on existing items at the moment the configuration is set, and not applied to any items created afterward. When you modify your Database configuration, your configured permissions are automatically set to Custom. The Custom setting ensures that your permissions are still valid for items created before you reconfigured your Database. Any items created after you change your configuration will not have permissions applied to them, you must set new permissions.
```
[...]

PR #3429, opened 2025-10-30
fix(rdb): surface effective_permission and permission_status for privileges drifts

jremy42:
close [#2762](https://github.com/scaleway/terraform-provider-scaleway/issues/2762)

### DIFF
--- a/internal/services/rdb/privilege.go
+++ b/internal/services/rdb/privilege.go
@@ -220,11 +220,7 @@ func readPrivilegeIntoState(ctx context.Context, d *schema.ResourceData, m any) 
 	effectivePermission := string(privilege.Permission)
 	configuredPermission := d.Get("permission").(string)
 
-	_ = d.Set("database_name", privilege.DatabaseName)
-	_ = d.Set("user_name", privilege.UserName)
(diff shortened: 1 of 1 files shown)


internal/services/rdb/privilege.go
```diff
@@ -184,13 +199,47 @@ func ResourceRdbPrivilegeRead(ctx context.Context, d *schema.ResourceData, m any
...
 	_ = d.Set("user_name", privilege.UserName)
-	_ = d.Set("permission", privilege.Permission)
 	_ = d.Set("instance_id", regional.NewIDString(region, instanceID))
 	_ = d.Set("region", region)
+	_ = d.Set("permission", privilege.Permission)
+	_ = d.Set("effective_permission", effectivePermission)
+
+	var diags diag.Diagnostics
+
+	if effectivePermission != configuredPermission {
```
remyleone, line 228: Is there a way to add a test for this?

jremy42: Testing drift requires SQL executed outside Terraform, which is tricky to automate in acceptance tests :(

PR #3445, opened 2025-11-04
feat(audit_trail): add new data source audit_trail_event

estellesoulard:
https://github.com/issues/assigned?issue=scaleway%7Cterraform-provider-scaleway%7C3427 

- I had initially added some tests without any filters, but this led to a huge cassette with a lot of info (listing all events on a whole organization for the past hour). This is why I opted to always include a filter on the test project to limit this.


templates/data-sources/audit_trail_event.md.tmpl
```diff
@@ -0,0 +1,79 @@
+---
```
remyleone, line 1: I think it would be helpful to give this file a tmpl file extension.

templates/data-sources/audit_trail_event.md.tmpl
```diff
@@ -0,0 +1,79 @@
...
+subcategory: "Audit Trail"
+page_title: "Scaleway: scaleway_audit_trail_event"
+---
+
+# scaleway_audit_trail_event
+
+Use this data source to get a list of existing Audit Trail events.
```
remyleone, line 8: I think it would be helpful to specify the range of time considered if nothing is specified.

internal/services/audittrail/event_data_source.go
```diff
@@ -0,0 +1,227 @@
...
+				Type:             schema.TypeString,
+				Description:      "Project containing the listed events",
+				Optional:         true,
+				ValidateDiagFunc: verify.IsUUID(),
+			},
+			"resource_type": {
+				Type:             schema.TypeString,
+				Description:      "Type of the scaleway resources associated with the listed events",
+				Optional:         true,
+				ValidateDiagFunc: verify.ValidateEnum[audittrailSDK.ResourceType](),
```
remyleone: Could you ensure that is it only a warning and not an error that is returned here? It would be helpful in case new resource_type are recognised by the API but not yet available in the provider and/or in a outdated provider version.

templates/data-sources/audit_trail_event.md.tmpl
estellesoulard: I'll add the recorded_after and recorded_before filters.  I left out a few other filters, should I add any of these additionally?
- method_name
- status
- order_by
- service_name
- principal_id
- source_ip

remyleone: Yes I think it would be helpful

internal/services/audittrail/event_data_source.go
estellesoulard: :heavy_check_mark:

templates/data-sources/audit_trail_event.md.tmpl
estellesoulard: Done :heavy_check_mark:

internal/services/audittrail/event_data_source.go
estellesoulard: I wanted to add a test for this warning, but : 
- terraform-plugin-testing is apparently missing support for warning check
- it triggers a 400 on API side, which is expected, given the resource_type I provide does not actually exist,but having a 400 in the tests breaks the cassette validation.  

This test did not seem worth fighting for, which is why I just removed it altogether  (see [this sad commit](https://github.com/scaleway/terraform-provider-scaleway/pull/3445/commits/a8b738e6223ca34f9e3c9963aaffba4c86ca5c50))

internal/services/audittrail/event_data_source_test.go
```diff
@@ -0,0 +1,268 @@
...
+				Config: fmt.Sprintf(`
+					resource "scaleway_secret" "main" {
+					  name        = "%s"
+					  description = "DataSourceAuditTrail test description"
+					  project_id  = "%s"
+					}
+
+					data "scaleway_audit_trail_event" "by_id" {
+						project_id = scaleway_secret.main.project_id
+						resource_id = split("/", scaleway_secret.main.id)[1]
```
remyleone, line 63: Could the data source be more flexible and accept a scaleway_secret_main and then do the extraction on its own?

internal/services/audittrail/event_data_source.go
```diff
@@ -0,0 +1,354 @@
...
+					}
+
+					return verify.ValidateStringInSliceWithWarning(resourceTypeStringValues, "resourceType")(i, p)
+				},
+			},
+			"resource_id": {
+				Type:             schema.TypeString,
+				Description:      "ID of the Scaleway resource associated with the listed events",
+				Optional:         true,
+				ValidateDiagFunc: verify.IsUUID(),
```
remyleone: I think we could be more flexible here and accept resources that are in the format locality/id

internal/services/audittrail/event_data_source.go
```diff
@@ -0,0 +1,354 @@
...
+			},
+			"product_name": {
+				Type:        schema.TypeString,
+				Description: "Scaleway product associated with the listed events in a hyphenated format",
+				Optional:    true,
```
remyleone, line 63: Is there a list of those product somewhere? It could be helpful to provide hint

internal/services/audittrail/event_data_source.go
```diff
@@ -0,0 +1,354 @@
...
+			},
+			"service_name": {
+				Type:        schema.TypeString,
+				Description: "Name of the service of the API call performed",
```
remyleone, line 67: Is there a list of those service_name? It could be helpful to provide hinting

internal/services/audittrail/event_data_source.go
```diff
@@ -0,0 +1,354 @@
...
+				Optional:    true,
+			},
+			"method_name": {
+				Type:        schema.TypeString,
+				Description: "Name of the method of the API call performed",
+				Optional:    true,
+			},
+			"principal_id": {
```
remyleone, line 75: Could you add a validator function here?

internal/services/audittrail/event_data_source.go
```diff
@@ -0,0 +1,354 @@
...
+				Type:        schema.TypeString,
+				Description: "ID of the User or IAM application at the origin of the event",
+				Optional:    true,
+			},
+			"source_ip": {
```
remyleone, line 81: Is there a validator for IPv4/IPv6 field?

internal/services/audittrail/event_data_source_test.go
```diff
@@ -0,0 +1,268 @@
...
+					}
+				`, secretName, project.ID),
+				Check: createEventDataSourceChecks("data.scaleway_audit_trail_event.by_id", orgID),
+			},
+			{
+				PreConfig: func() { waitForAuditTrailEvents(t, ctx, auditTrailAPI, project) },
+				Config: fmt.Sprintf(`
```
remyleone: Why all this step are required? You could have a single event and then define several blocks of data resource and perform checks on each of them.

internal/services/audittrail/event_data_source_test.go
estellesoulard: These tests are on the audit trail events for a secret resource, but the secret id format ('fr-par/11111111-1111-1111-1111-111111111111')  does not match the proto validation itself ("^[a-zA-Z0-9._-]+$", so no '/' accepted as delimiter). 
Accepting other formats with the locality (such as 'fr-par_11111111-1111-1111-1111-111111111111' or 'fr-par-11111111-1111-1111-1111-111111111111') could work for products with ids formatted as such, but still not for secret resources.
Should we update the proto?

estellesoulard: Erratum: realized I might have misunderstood what you meant. I'll see if we can parse the uuid on the provider side before querying the api

internal/services/audittrail/event_data_source.go
estellesoulard: There appears to be no public list, and they're only pattern validated (no enum on sdk side). 
I have contacted the team to see if they could add a list to their docs for those entries. 
In the meantime I built a list from the sources and I included it in the .mds.

internal/services/audittrail/event_data_source.go
estellesoulard: There appears to be no public list, and they're only pattern validated (no enum on sdk side). 
I have contacted the team to see if they could add a list to their docs for those entries. 
In the meantime I built a list from the sources and I included it in the .mds.

internal/services/audittrail/event_data_source_test.go
estellesoulard: Indeed!! I have reworked the tests, should be better now :)

internal/services/audittrail/event_data_source.go
estellesoulard: The list is available from their ListProducts endpoint, however i'm not sure we'd want to add an API call in the Datasource function purely for informative value. Leaving it as is for now, let me know what you think!

docs/data-sources/audit_trail_event.md
```diff
@@ -0,0 +1,100 @@
...
+
+Use this data source to get a list of existing Audit Trail events.
+For more information refer to the [Audit Trail API documentation](https://www.scaleway.com/en/developers/api/audit-trail/).
+
+## Example Usage
+
+```hcl
+# Retrieve all audit trail events on the default organization
+data "scaleway_audit_trail_event" "find_all" {
+  region = "fr-par"
```
remyleone: I'm not sure to understand why we want a region filter here if we want all the events in the organization. Will it filter event not happening on `fr-par`

docs/data-sources/audit_trail_event.md
```diff
@@ -0,0 +1,100 @@
...
+}
+
+# Retrieve audit trail events on a specific organization
+data "scaleway_audit_trail_event" "find_by_org" {
+  organization_id = "11111111-1111-1111-1111-111111111111"
+}
+
+# Retrieve audit trail events on a specific project
+data "scaleway_audit_trail_event" "find_by_project" {
+  region = "fr-par"
```
remyleone: I think the region should be removed here if the focus of this example is the project_id

docs/data-sources/audit_trail_event.md
```diff
@@ -0,0 +1,100 @@
...
+}
+
+# Retrieve audit trail for a specific resource
+data "scaleway_audit_trail_event" "find_by_resource_id" {
+  resource_id = "11111111-1111-1111-1111-111111111111"
+}
+
+# Retrieve audit trail for a specific Scaleway product
+data "scaleway_audit_trail_event" "find_by_product_name" {
+  region = "nl-ams"
```
remyleone: Will it merge the results across all regions if the region flag is missing? I think this region attribute should be removed.

internal/services/audittrail/helpers.go
```diff
@@ -0,0 +1,27 @@
...
+// newAPIWithRegionAndProjectID returns a new Audit Trail API, with region and projectID
+func newAPIWithRegion(d *schema.ResourceData, m any) (*audittrailSDK.API, scw.Region, error) {
+	api := audittrailSDK.NewAPI(meta.ExtractScwClient(m))
+
+	region, err := meta.ExtractRegion(d, m)
+	if err != nil {
+		return nil, "", err
+	}
+	// In audit trail sdk-go so far only fr-par and nl-ams are supported
+	if !slices.Contains(api.Regions(), region) {
```
remyleone: I would prefer not to send an error from the provider side. In case we add new regions and the provider is not yet updated.

docs/data-sources/audit_trail_event.md
estellesoulard: Audittrail API is regional, only one region's result are returned. If the region flag is missing the default provider region is used.

docs/data-sources/audit_trail_event.md
estellesoulard: Nice catch, this region and the others in the doc are just oversights, i cleared them :heavy_check_mark:

internal/services/audittrail/helpers.go
estellesoulard: Agreed! i removed this check :heavy_check_mark:

PR #3474, opened 2025-11-25
chore: bump scaleway-sdk-go and fix compatibility issues

jremy42:


internal/services/file/waiter.go
```diff
@@ -16,7 +16,7 @@ func waitForFileSystem(ctx context.Context, fileAPI *file.API, region scw.Region
 	}
 
 	fileSystem, err := fileAPI.WaitForFileSystem(&file.WaitForFileSystemRequest{
-		FileSystemID:  id,
+		FilesystemID:  id,
```
remyleone, line 19: @yfodil I think there is an issue in this waiter

internal/services/k8s/waiters.go
```diff
@@ -50,7 +50,6 @@ func waitClusterStatus(ctx context.Context, k8sAPI *k8s.API, cluster *k8s.Cluste
 	cluster, err := k8sAPI.WaitForCluster(&k8s.WaitForClusterRequest{
 		ClusterID:     cluster.ID,
 		Region:        cluster.Region,
-		Status:        status,
```
remyleone, line 53: Why is this line removed I think it is the main focus of this function!

jremy42: The Status field was present in WaitForClusterRequest but doesn't appear to have been used in the implementation. When waiters were [auto-generated ](https://github.com/scaleway/scaleway-sdk-go/commit/febdd5aefa584a83d2229f94eab43f071f09e5d9#diff-f42151d5483d9f1fbe2bee9b532209f1390140f1ea7361017ab6e4c0697a689c )it was removed, likely because the generator only keeps fields that are actually used. So the Status: status, line was dropped — it may not have had any effect before.

jremy42: The Status field existed in WaitForClusterRequest but wasn't used in the implementation. The function checked cluster.Status against a hardcoded map of terminal states, ignoring req.Status.

internal/services/file/waiter.go
yfodil: The generated waiter is picking the field name from the get request https://github.com/scaleway/scaleway-sdk-go/blob/master/api/file/v1alpha1/file_sdk.go#L269

The previous waiter that was manually implemented had it written as [FileSystemID](https://github.com/scaleway/scaleway-sdk-go/commit/febdd5aefa584a83d2229f94eab43f071f09e5d9#diff-922a226c259968306de99aa430239178e567b2d83f51d69778e23725fdbf5bdeL17) instead 

It's a little break, this waiter is relatively recent and I don't think that it's used that much

Issue #3337, opened 2025-09-16
Add support for actions in mongodb


remyleone:
It would be helpful to add support in mongodb for actions:

- CreateSnapshot

PR #3494, opened 2025-12-02
feat(mongodb): add mongodb instance snapshot action

jremy42:
close [#3337](https://github.com/scaleway/terraform-provider-scaleway/issues/3337)


internal/services/mongodb/action_instance_snapshot_action.go
```diff
@@ -0,0 +1,202 @@
...
+			"region": schema.StringAttribute{
+				Optional:    true,
+				Description: "Region of the MongoDB instance. If not set, the region is derived from the instance_id when possible or from the provider configuration.",
+			},
+			"name": schema.StringAttribute{
+				Optional:    true,
+				Description: "Name of the snapshot. If not set, a name will be generated.",
+			},
+			"expires_at": schema.StringAttribute{
+				Required:    true,
```
remyleone: Why should it be required in case I don't want it to expire ?

jremy42: fixed

internal/services/mongodb/action_instance_snapshot_action.go
```diff
@@ -0,0 +1,229 @@
...
+}
+
+func (a *InstanceSnapshotAction) Schema(_ context.Context, _ action.SchemaRequest, resp *action.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Attributes: map[string]schema.Attribute{
+			"instance_id": schema.StringAttribute{
+				Required:    true,
+				Description: "MongoDB instance ID to snapshot. Can be a plain UUID or a regional ID.",
+				Validators: []validator.String{
+					stringvalidator.LengthAtLeast(1),
```
remyleone: I don't think this validator is very helpful. It should be either removed or replaced with a proper validator that can check for a localized/unlocalized id

internal/services/mongodb/action_instance_snapshot_action.go
```diff
@@ -0,0 +1,229 @@
...
+			"Missing region",
+			"Could not determine region for MongoDB snapshot. Please set the region attribute, use a regional instance_id, or configure a default region in the provider.",
+		)
+
+		return
+	}
+
+	snapshotName := data.Name.ValueString()
+	if snapshotName == "" {
+		snapshotName = "tf-mongodb-snapshot"
```
remyleone: could you add action also here to have more context in this default value?

Issue #3332, opened 2025-09-16
Add support for actions in cockpit


remyleone:
It would be helpful to have support for actions related to cockpit ressources:

- TriggerTestAlert
- ResetGrafanaUserPassword
- SyncGrafanaDataSources

PR #3499, opened 2025-12-04
feat(cockpit): add support for cockpit actions

jremy42:
close [#3332](https://github.com/scaleway/terraform-provider-scaleway/issues/3332)

### DIFF
--- a/internal/locality/regional/schemas_framework.go
+++ b/internal/locality/regional/schemas_framework.go
@@ -1,9 +1,7 @@
 package regional
 
 import (
-	"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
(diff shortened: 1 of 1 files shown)


internal/locality/regional/schemas_framework.go
```diff
@@ -0,0 +1,29 @@
...
+	return regions
+}
+
+// SchemaAttribute returns a Plugin Framework schema attribute for a region field
+func SchemaAttribute() schema.StringAttribute {
+	return schema.StringAttribute{
+		Optional:    true,
+		Description: "The region you want to attach the resource to",
+		Validators: []validator.String{
+			stringvalidator.OneOf(AllRegions()...),
```
remyleone: I would like only a warning and not a blocker

Issue #3336, opened 2025-09-16
Add support for actions in key manager


remyleone:
It would be helpful to add support for action in the provider for key manager:

- RotateKeyRequest
- GenerateDataKey
- Encrypt
- Decrypt
- Sign
- Verify

estellesoulard:
Note: While actions are appropriate for RotateKeyRequest, other calls require interaction with the payload and are more aligned with the Ephemeral Resource pattern.

estellesoulard:
Encrypt: https://github.com/scaleway/terraform-provider-scaleway/commit/ebff5d371791656a582efecd83091031e5ee242b

PR #3517, opened 2025-12-08
feat(keymanager): add RotateKeyAction

estellesoulard:
https://github.com/scaleway/terraform-provider-scaleway/issues/3336

### DIFF
--- a/docs/actions/key_manager_key_rotate_action.md
+++ b/docs/actions/key_manager_key_rotate_action.md
@@ -15,3 +15,5 @@ page_title: "Scaleway: scaleway_key_manager_key_rotate_action"
 ### Optional
 
 - `region` (String) Region of the key. If not set, the region is derived from the key_id when possible or from the provider configuration.
+
+


internal/services/keymanager/action_rotate_key_action.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+func (a *RotateKeyAction) Metadata(ctx context.Context, req action.MetadataRequest, resp *action.MetadataResponse) {
+	resp.TypeName = req.ProviderTypeName + "_key_manager_key_rotate_action"
+}
+
+type RotateKeyActionModel struct {
+	Region types.String `tfsdk:"region"`
+	KeyID  types.String `tfsdk:"key_id"`
+}
```
estellesoulard, line 54: I have not included a `wait` argument because we do not have a `WaitFor()` in the key_manager sdk. Should I add one? (i'm still unclear on the conditions required for a WaitFor)

internal/services/keymanager/action_rotate_key_action_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/terraform"
+	kmSDK "github.com/scaleway/scaleway-sdk-go/api/key_manager/v1alpha1"
+	"github.com/scaleway/scaleway-sdk-go/scw"
+	"github.com/scaleway/terraform-provider-scaleway/v2/internal/acctest"
+	"github.com/scaleway/terraform-provider-scaleway/v2/internal/services/keymanager"
+)
+
+func isKeyRotated(tt *acctest.TestTools, resourceName string) resource.TestCheckFunc {
```
estellesoulard: Using this check function because we do not have a datasource for keys yet.

docs/actions/key_manager_key_rotate_action.md
```diff
@@ -0,0 +1,8 @@
+---
+subcategory: "Key Manager"
+page_title: "Scaleway: scaleway_key_manager_key_rotate_action"
+---
+
+# scaleway_key_manager_key_rotate_action (Action)
+
+<no value>
```
remyleone: I would expect the schema to show up here

internal/services/keymanager/action_rotate_key_action.go
remyleone: We can only have a wait if we have a status. I don't think `KeyState` help us for that

internal/services/keymanager/action_rotate_key_action_test.go
remyleone: I think data source should come first. We should have a data source before the action

internal/services/keymanager/action_rotate_key_action_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+func TestAccActionRotateKey_Basic(t *testing.T) {
+	if acctest.IsRunningOpenTofu() {
+		t.Skip("Skipping TestAccActionRotateKey_Basic because actions are not yet supported on OpenTofu")
+	}
+
+	tt := acctest.NewTestTools(t)
+	defer tt.Cleanup()
+
+	region := "fr-par"
```
remyleone: not very useful

internal/services/keymanager/action_rotate_key_action.go
estellesoulard: Thanks, that confirms my suspicion that waitFor() is necessarily status dependent ! My idea was to setup a waitFor() that would watch the RotationCount and return once it's increased, but that now sounds like a bit of a stretch for a waitFor(). Closing

internal/services/keymanager/action_rotate_key_action_test.go
estellesoulard: oops, leftover from a previous implem

internal/services/keymanager/action_rotate_key_action_test.go
```diff
@@ -0,0 +1,85 @@
...
+							}
+						}
+					}
+
+					action "scaleway_key_manager_key_rotate_action" "main" {
+						config {
+							key_id = scaleway_key_manager_key.main.id
+							region = scaleway_key_manager_key.main.region
+						}
+					}
```
estellesoulard, line 79: Right now the action does not trigger at all (rotate API call does not happen in the test, nor with a dev_override)

estellesoulard: ...EBKAC error. Had a local kerfuffle of terraform bins.

internal/services/keymanager/action_rotate_key_action.go
```diff
@@ -0,0 +1,135 @@
...
+		Attributes: map[string]schema.Attribute{
+			"region": schema.StringAttribute{
+				Optional:    true,
+				Description: "Region of the key. If not set, the region is derived from the key_id when possible or from the provider configuration.",
+			},
+			"key_id": schema.StringAttribute{
+				Required:    true,
+				Description: "ID of the key to rotate (UUID format)",
+			},
+		},
```
estellesoulard, line 71: Maybe add validation

remyleone (approved):

PR #3524, opened 2025-12-10
feat(keymanager): add key datasource

estellesoulard:


internal/services/keymanager/key_data_source.go
```diff
@@ -0,0 +1,57 @@
...
+	dsSchema["key_id"] = &schema.Schema{
+		Type:             schema.TypeString,
+		Required:         true,
+		Description:      "The ID of the key",
+		ValidateDiagFunc: verify.IsUUIDorUUIDWithLocality(),
+	}
+
+	return &schema.Resource{
+		ReadContext: dataSourceKeyRead,
+		Schema:      dsSchema,
```
remyleone: Use a SchemaFunc here also

estellesoulard: Done :heavy_check_mark:

Issue #3423, opened 2025-10-24
Add a write only attribute to all sensitive attributes all resources


remyleone:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

It would be helpful to have support in all resources that handle sensitive values a support for a [write only](https://developer.hashicorp.com/terraform/language/manage-sensitive-data/write-only) attribute.
This would provide an option to not write a secret in the state or plan files while providing similar features.

This change must be done with good documentation to show case how to migrate from one way to use those sensitive values to the next where they are not persisted in the state. Using a secret resource provided by Scaleway could be helpful to store those secrets.

### New or Affected Resource(s)

* scaleway_redis_cluster
* scaleway_applesilicon_server
* scaleway_baremetal_server
* scaleway_cockpit_grafana_user (might get deprecated and migrated to iam)
* scaleway_container (with `secret_environment_variables`)
* scaleway_container_namespace (with `secret_environment_variables`)
* scaleway_container_token
* scaleway_function (with `secret_environment_variables`)
* scaleway_function_namespace (with `secret_environment_variables`)
* scaleway_function_token
* scaleway_iam_api_key
* scaleway_iam_user
* scaleway_inference_model (secret)
* scaleway_iot_device
* scaleway_iot_network
* scaleway_iot_routes 
* scaleway_k8s_cluster (kubeconfig)
* scaleway_lb_certificate
* scaleway_nats_credentials
* scaleway_sqs_credentials
* scaleway_sns_credentials
* scaleway_sns_topic
* scaleway_sqs_queue
* scaleway_mongodb_instance
* scaleway_mongodb_user
* scaleway_object (`sse_customer_key`)
* scaleway_rdb_instance (`password`)
* scaleway_rdb_user (`password`)
* scaleway_config (`secret_key`)
* scaleway_secret_version

### References

* https://developer.hashicorp.com/terraform/plugin/framework/resources/write-only-arguments
* https://developer.hashicorp.com/terraform/plugin/sdkv2/resources/write-only-arguments

ndeybach:
Hi, 

[...]

PR #3538, opened 2025-12-15
feat(secret): add support for WriteOnly data argument in secret version

estellesoulard:
### Context
I was implementing key manager's `Encrypt` call as an `Ephemeral Resource`. This kind of resource seems like the most appropriate implementation choice (`Ephemeral Resources` are not persisted in state, and seems perfect for the use case of returning an encrypted payload).
To test this `Ephemeral Resource`, I needed a matching resource argument that I could transfer my `Ephemeral Resource`'s value into. This is necessary because `Ephemeral Resources` cannot be outputted nor used liberally in any other resource, therefore in the acceptance tests I had no way to check that the `Ephemeral Resource` properly worked. 
However, an `Ephemeral Resource`'s attribute value can be transferred into a `WriteOnly` argument, that I could later retrieve in the target resource's datasource. This is exactly what I did here :smile: 

### PR content: TLDR
This PR adds a `data_wo` argument to secret_version Resource that can be used alternatively to the classic `data` argument. 
Since this `data_wo` is a `WriteOnly` it is not stored in the state, and drifts cannot be detected. It is fundamentally incompatible with `ForceNew`. I bypassed this issue by adding a  `data_wo_version` with ForceNew, enabling a simili-ForceNew for `data_wo`: updating `data_wo` requires to update `data_wo_version`, which will then be properly caught by Terraform and trigger a recreation. 

### A word of caution
This PR is the first in a long line of WriteOnly attribute additions. (cf https://github.com/scaleway/terraform-provider-scaleway/issues/3423). It may be treated as a blueprint for the following ones: please raise any concern during your review :saluting_face:


docs/resources/secret_version.md
```diff
@@ -37,7 +37,8 @@ resource "scaleway_secret_version" "v1" {
 The following arguments are supported:
 
 - `secret_id` - (Required) The ID of the secret associated with the version.
-- `data` - (Required) The data payload of the secret version. Must not exceed 64KiB in size (e.g. `my-secret-version-payload`). Find out more on the [data section](/#data-information).
+- `data` - (Optional) The data payload of the secret version. Must not exceed 64KiB in size (e.g. `my-secret-version-payload`). Only one of `data` or `data_wo` should be specified. Find out more on the [data section](/#data-information).
+- `data_wo` - (Optional) The data payload of the secret version. Must not exceed 64KiB in size (e.g. `my-secret-version-payload`).  Only one of `data` or `data_wo` should be specified. `data_wo` will not be set in the Terraform state. Find out more on the [data section](/#data-information).
```
remyleone: It would be helpful to have a link that can send to the write only documentation as not all user are familiar with this notion

remyleone (dismissed):

PR #3539, opened 2025-12-15
feat(keymanager): add encrypt ephemeral resource

estellesoulard:
- Added Ephemeral Resources in provider
- Added tests for compulsory Ephemeral Resources descriptions (same as for actions)
- Added key manager Encrypt Ephemeral Resource

### DIFF
--- a/templates/ephemeral-resources/key_manager_encrypt.md.tmpl
+++ b/templates/ephemeral-resources/key_manager_encrypt.md.tmpl
@@ -1,4 +1,4 @@
-{{- /*gotype: github.com/hashicorp/terraform-plugin-docs/internal/provider.ActionTemplateType */ -}}
+{{- /*gotype: github.com/hashicorp/terraform-plugin-docs/internal/provider.ResourceTemplateType */ -}}
 ---
 subcategory: "Key Manager"
 page_title: "Scaleway: scaleway_key_manager_encrypt"


.markdownlint.yml
```diff
@@ -16,6 +16,7 @@ MD010: false
 MD012: false
 MD013: false
 MD022: false
+MD033: false
```
estellesoulard, line 19: Generated files (docs/ephemeral-resources/key_manager_encrypt.md) broke the markdown linter because of the `<a id="nestedatt--associated_data"></a>` line.  Error seemed safe enough to ignore (let me know if not)

templates/ephemeral-resources/key_manager_encrypt.md.tmpl
```diff
@@ -0,0 +1,11 @@
+{{- /*gotype: github.com/hashicorp/terraform-plugin-docs/internal/provider.ActionTemplateType */ -}}
```
estellesoulard: TODO

internal/services/keymanager/encrypt_ephemeral_resource_test.go
```diff
@@ -0,0 +1,60 @@
...
+import (
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/scaleway/terraform-provider-scaleway/v2/internal/acctest"
+)
+
+func TestAccEncryptEphemeralResource_Basic(t *testing.T) {
+	if acctest.IsRunningOpenTofu() {
+		t.Skip("Skipping TestAccEncryptEphemeralResource_Basic because testing Ephemeral Resources is not yet supported on OpenTofu")
```
estellesoulard, line 12: Test fails with opentofu, looks like the ephemeral resource is not properly evaluated (even when adding a dependency as I did here). My understanding is that ephemeral resources testing is not properly handled yet: https://github.com/opentofu/opentofu/issues/3557

remyleone (approved):

PR #3540, opened 2025-12-16
docs: add support for description for k8s resources

remyleone:


estellesoulard (dismissed):
LGTM (just the lint error `MD031` in the generated files, not sure if we can fix it in the generation or if we want to add an ignore)

PR #3545, opened 2025-12-18
feat(keymanager): add decrypt ephemeral resource

estellesoulard:
Add key manager Decrypt Ephemeral Resource (mostly the same as the Encrypt resource)
https://github.com/scaleway/terraform-provider-scaleway/issues/3336

### DIFF
--- a/internal/services/keymanager/decrypt_ephemeral_resource.go
+++ b/internal/services/keymanager/decrypt_ephemeral_resource.go
@@ -85,7 +85,6 @@ func (r *DecryptEphemeralResource) Schema(ctx context.Context, req ephemeral.Sch
 			"ciphertext": schema.StringAttribute{
 				Required:    true,
 				Description: "Ciphertext data to decrypt. Data size must be between 1 and 131071 bytes.",
-				Sensitive:   true,
 			},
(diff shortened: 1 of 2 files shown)


internal/services/keymanager/decrypt_ephemeral_resource.go
```diff
@@ -0,0 +1,192 @@
...
+				Required:    true,
+				Description: "Ciphertext data to decrypt. Data size must be between 1 and 131071 bytes.",
+				Sensitive:   true,
+			},
+			"associated_data": schema.ObjectAttribute{
+				Optional:    true,
+				Description: "Must match the associated_data value passed in the encryption request. Only supported by keys with a usage set to `symmetric_encryption`.",
+				AttributeTypes: map[string]attr.Type{
+					"value": types.StringType,
+				},
```
estellesoulard, line 94: i set this as an object attribute to match the api, but thinking back would it be cleaner to just have an optional string ? (I did use this pattern in the Encrypt Ephemeral Resource already, so if I change it here I will also have to update it in Encrypt)

yfodil (dismissed):

PR #3548, opened 2025-12-19
feat(cockpit): add exporter resource

jremy42:


internal/acctest/anonymize.go
```diff
@@ -0,0 +1,164 @@
+package acctest
```
remyleone, line 1: Could you add a test that read all cassettes and ensure that there are no leak?

PR #3549, opened 2025-12-22
docs: add support for description inside action schema

remyleone:


docs/actions/key_manager_key_rotate_action.md
```diff
@@ -1,10 +1,20 @@
 ---
-subcategory: "Key Manager"
-page_title: "Scaleway: scaleway_key_manager_key_rotate_action"
+# generated by https://github.com/hashicorp/terraform-plugin-docs
+page_title: "scaleway_key_manager_key_rotate_action Action - terraform-provider-scaleway"
+subcategory: ""
```
yfodil: With these changes, the subcategories are now empty. Is this intentional?

remyleone: https://github.com/hashicorp/terraform-plugin-docs/issues/156 I don't think it is yet solved. :/ I think we will need to keep the template for the time being

remyleone: fixed

PR #3550, opened 2025-12-23
feat(vpcgw): add refresh ssh keys action

yfodil:


internal/services/vpcgw/action_refresh_ssh_keys_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+					action "scaleway_vpcgw_refresh_ssh_keys_action" "main" {
+						config {
+							gateway_id = scaleway_vpc_public_gateway.main.id
+							wait = true
+						}
+					}
+				`,
+				Check: resource.ComposeTestCheckFunc(
+					isGatewaySSHKeysRefreshed(tt, "scaleway_vpc_public_gateway.main"),
```
estellesoulard: Maybe this check can be done with the datasource ?

remyleone (approved):

PR #3556, opened 2025-12-24
feat(keymanager): add sign ephemeral resource

estellesoulard:
[#3336](https://github.com/scaleway/terraform-provider-scaleway/issues/3336)


templates/ephemeral-resources/key_manager_sign.md.tmpl
```diff
@@ -0,0 +1,11 @@
+{{- /*gotype: github.com/hashicorp/terraform-plugin-docs/internal/provider.ResourceTemplateType */ -}}
+---
+subcategory: "Key Manager"
+page_title: "Scaleway: scaleway_key_manager_sign"
+---
+
+# scaleway_key_manager_sign (Ephemeral Resource)
```
remyleone: I think this could be obtained from `{{ .Name }}``

templates/ephemeral-resources/key_manager_sign.md.tmpl
```diff
@@ -0,0 +1,11 @@
...
+
+{{ .Description }}
+
```
remyleone, line 10: Could you also add an `{{ .Examples }}` https://github.com/hashicorp/terraform-plugin-docs?tab=readme-ov-file#conventional-paths 

By providing examples directly as file it should be possible to include them in the templates and improve the documentation.

remyleone: Here is an example of a template using Examples https://github.com/scaleway/terraform-provider-scaleway/blob/master/templates/resources/k8s_cluster.md.tmpl#L11-L20

estellesoulard: Done!

remyleone (approved):

PR #3582, opened 2026-01-07
feat(applesilicon): add support for identity

remyleone:


internal/services/applesilicon/server.go
```diff
@@ -293,6 +303,22 @@ func ResourceAppleSiliconServerRead(ctx context.Context, d *schema.ResourceData,
...
+
+	if res.Commitment != nil {
+		switch res.Commitment.Type {
+		case applesilicon.CommitmentTypeNone, applesilicon.CommitmentTypeDuration24h:
+			_ = d.Set("commitment", applesilicon.CommitmentTypeDuration24h.String())
+		case applesilicon.CommitmentTypeRenewedMonthly:
+			_ = d.Set("commitment", applesilicon.CommitmentTypeRenewedMonthly.String())
+		}
+	}
+
```
estellesoulard, line 321: I see this was added as a fix but I am not sure to understand. Was this missing in the previous implementation?

estellesoulard (approved):

internal/services/applesilicon/server.go
remyleone: Without it the identity import test fails because the state is non consistent. Those information were not included in the state before. Identity test checks for this kind of mistake making it really helpful to push it in all resources :)

PR #3606, opened 2026-01-14
feat(k8s): make security_group_id editable on existing pools

jtherin:


internal/services/k8s/pool.go
```diff
@@ -566,6 +565,10 @@ func ResourceK8SPoolUpdate(ctx context.Context, d *schema.ResourceData, m any) d
 
 	updateRequest.UpgradePolicy = upgradePolicyReq
 
+	if d.HasChange("security_group_id") {
```
remyleone, line 568: I think it would be helpful to update the documentation as well

PR #3612, opened 2026-01-16
docs(container): add example for how to reload `latest` tags

cyclimse:
Hello !

There's no super clear way to reload a container with the `latest` tag documented in our Terraform provider, which can cause confusion.

I've updated the docs to include a small example on using `registry_sha256`

I've also updated the example container resources to match what's automatically set in the Console.

Have a great day!

### DIFF
--- a/docs/resources/container.md
+++ b/docs/resources/container.md
@@ -11,6 +11,7 @@ Refer to the Serverless Containers [product documentation](https://www.scaleway.
 
 For more information on the limitations of Serverless Containers, refer to the [dedicated documentation](https://www.scaleway.com/en/docs/serverless-containers/reference-content/containers-limitations/).
 
+
 ## Example Usage
 
 ```terraform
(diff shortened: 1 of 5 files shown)


docs/resources/container.md
```diff
@@ -223,7 +261,7 @@ You can determine the computing resources to allocate to each container.
 The `memory_limit` (in MB) must correspond with the right amount of vCPU. Refer to the table below to determine the right memory/vCPU combination.
 
 | Memory (in MB) | vCPU |
-|----------------|------|
+| -------------- | ---- |
```
cyclimse: This information looks out of date, we can now modify both `memory_limit` and `cpu_limit` but this seems to indicate otherwise 🤔

thomas-tacquet (dismissed):

docs/resources/container.md
```diff
@@ -102,6 +102,44 @@ $ curl -H "X-Auth-Token: $(terraform output -raw secret_key)" \
 
 Keep in mind that you should revoke your legacy JWT tokens to ensure maximum security.
 
+### Deploying containers with mutable images
+
+When using mutable images (e.g., `latest` tag), you can use the `scaleway_container_registry_image` data source along with the `registry_sha256` argument to trigger container redeployments when the image is updated.
```
cyclimse:
```suggestion
When using mutable images (e.g., `latest` tag), you can use the `scaleway_registry_image_tag` data source along with the `registry_sha256` argument to trigger container redeployments when the image is updated.
```

remyleone (approved):

PR #3620, opened 2026-01-22
fix(domain): handle FQDN normalization for CNAME targets

yfodil:
Scaleway API canonicalizes hostname record targets (adds zone + trailing dot)

Normalize and suppress diffs to avoid Terraform plan diffs for CNAME/NS/MX records

### DIFF
--- a/internal/services/domain/record.go
+++ b/internal/services/domain/record.go
@@ -98,6 +98,13 @@ func recordSchema() map[string]*schema.Schema {
 			Type:        schema.TypeString,
 			Description: "The data of the record",
 			Required:    true,
+			// NOTE: For CNAME/NS/MX records, the Scaleway API normalizes the "data" field to an absolute FQDN with a trailing dot. Example:
+			//
+			//   config: data = "www"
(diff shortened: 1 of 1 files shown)


internal/services/domain/record.go
```diff
@@ -98,6 +98,13 @@ func recordSchema() map[string]*schema.Schema {
 			Type:        schema.TypeString,
 			Description: "The data of the record",
 			Required:    true,
+			DiffSuppressFunc: func(k, oldValue, newValue string, d *schema.ResourceData) bool {
```
remyleone, line 108: Could you add some documentation to explain this behavior

internal/services/domain/helpers.go
```diff
@@ -663,3 +663,36 @@ func normalizeRecordName(name, dnsZone string) string {
 
 	return name
 }
+
+func normalizeRecordData(data string, recordType domain.RecordType, dnsZone string) string {
```
remyleone: Could you add some unit test for those functions?

PR #3626, opened 2026-01-28
feat(secret): add secret_version ephemeral resource

estellesoulard:
https://github.com/scaleway/terraform-provider-scaleway/issues/2973

### DIFF
--- a/internal/services/secret/version_ephemeral_resource.go
+++ b/internal/services/secret/version_ephemeral_resource.go
@@ -98,7 +98,6 @@ func (r *VersionEphemeralResource) Schema(ctx context.Context, req ephemeral.Sch
 			},
 			"secret_name": schema.StringAttribute{
 				Optional:    true,
-				Computed:    true,
 				Description: "The name of the secret.  Either secret_id or secret_name must be specified.",
 				Validators: []validator.String{
 					stringvalidator.ExactlyOneOf(
@@ -255,6 +254,7 @@ func (r *VersionEphemeralResource) Open(ctx context.Context, req ephemeral.OpenR
 		}
 
 		secretID = foundSecret.ID
+		data.SecretID = types.StringValue(secretID)
 	case !data.SecretID.IsNull() && !data.SecretID.IsUnknown():
 		secretID = locality.ExpandID(data.SecretID.ValueString())
 	default:
--- a/internal/services/secret/testdata/ephemeral-resource-secret-version-basic.cassette.yaml
+++ b/internal/services/secret/testdata/ephemeral-resource-secret-version-basic.cassette.yaml
@@ -20,22 +20,22 @@ interactions:
         proto: HTTP/2.0
         proto_major: 2
         proto_minor: 0
-        content_length: 492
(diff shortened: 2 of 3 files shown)


docs/ephemeral-resources/secret_version.md
```diff
@@ -0,0 +1,36 @@
...
+
+Contrary to the secret_version data source, this is an [ephemeral resource](https://developer.hashicorp.com/terraform/plugin/framework/ephemeral-resources): it will not be persisted in plan or state artifacts.
+
+This ephemeral resource is particularly useful when you need to access secret values during Terraform execution without persisting them in the state file, especially if that secret's data has been set with the [write-only](https://developer.hashicorp.com/terraform/language/manage-sensitive-data/write-only) attribute [`data_wo`](https://registry.terraform.io/providers/scaleway/scaleway/latest/docs/resources/secret_version#data_wo-2). You can avoid exposing sensitive data in your Terraform state files, from the creation to the access of your secret_version resource.
+
+Refer to the Secret Manager [documentation](https://www.scaleway.com/en/docs/secret-manager/) and [API documentation](https://www.scaleway.com/en/developers/api/secret-manager/) for more information.
+
```
estellesoulard, line 15: add examples

internal/services/secret/version_ephemeral_resource.go
```diff
@@ -0,0 +1,318 @@
...
+	data.Status = types.StringValue(getResponse.Status.String())
+	if getResponse.CreatedAt != nil {
+		data.CreatedAt = types.StringValue(getResponse.CreatedAt.Format(time.RFC3339))
+	}
+
+	if getResponse.UpdatedAt != nil {
+		data.UpdatedAt = types.StringValue(getResponse.UpdatedAt.Format(time.RFC3339))
+	}
+
+	data.Data = types.StringValue(base64.StdEncoding.EncodeToString(accessResponse.Data))
```
yfodil, line 314: nit: 
```suggestion
	data.Data = types.StringValue(secret.Base64Encoded(accessResponse.Data))
```

internal/services/secret/version_ephemeral_resource.go
```diff
@@ -0,0 +1,318 @@
...
+var versionEphemeralResourceDescription string
+
+func (r *VersionEphemeralResource) Schema(ctx context.Context, req ephemeral.SchemaRequest, resp *ephemeral.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Description:         versionEphemeralResourceDescription,
+		MarkdownDescription: versionEphemeralResourceDescription,
+		Attributes: map[string]schema.Attribute{
+			"secret_id": schema.StringAttribute{
+				Optional:    true,
+				Computed:    true,
```
yfodil, line 89: When looking up by `secret_name`, the `secret_id` isn't set in the result I believe, is `computed` needed/wanted here?

internal/services/secret/version_ephemeral_resource.go
estellesoulard: I don't think we should do that: this ephemeral resource is equivalent to the secret version datasource, and we should follow its behaviour. In the datasource, I had attempted this change and had to [revert it](https://github.com/scaleway/terraform-provider-scaleway/commit/e3080edb599be4b8ff11bb8b2bd278e257cd86fa). because `secret.Base64Encoded` does not systematically encode as base64, which led to inconsistent behaviour depending on the content of `data` (for instance, if for any reason a customer had provided base64 in the first place).

internal/services/secret/version_ephemeral_resource.go
estellesoulard: Oops indeed, nice catch! I left the `secret_id` as `computed` but I added the missing `set` when we are doing a get by `name`. Also I removed the computed from secret_name, I realized it's not in the API return

yfodil (approved):

PR #3659, opened 2026-02-11
feat(rdb): Implement Identity for all RDB resources

jremy42:


internal/services/rdb/database.go
```diff
@@ -131,8 +133,9 @@ func ResourceRdbDatabaseCreate(ctx context.Context, d *schema.ResourceData, m an
 		return diag.FromErr(err)
 	}
 
-	d.SetId(ResourceRdbDatabaseID(region, instanceID, db.Name))
-	_ = d.Set("region", region)
+	if err := identity.SetRegionalIdentity(d, region, instanceID+"/"+db.Name); err != nil {
```
remyleone: It is not a good idea, we need to have a dedicated function in the identity package for multi-part id.

internal/services/rdb/database.go
```diff
@@ -175,13 +178,15 @@ func ResourceRdbDatabaseRead(ctx context.Context, d *schema.ResourceData, m any)
 		return diag.FromErr(err)
 	}
 
-	d.SetId(ResourceRdbDatabaseID(region, instanceID, database.Name))
-	_ = d.Set("instance_id", regional.NewID(region, instanceID).String())
+	if err := identity.SetRegionalIdentity(d, region, instanceID+"/"+database.Name); err != nil {
```
remyleone: It is not a good idea, we need to have a dedicated function in the identity package for multi-part id.

internal/services/rdb/privilege.go
```diff
@@ -122,7 +124,10 @@ func ResourceRdbPrivilegeCreate(ctx context.Context, d *schema.ResourceData, m a
 		return diag.FromErr(err)
 	}
 
-	d.SetId(ResourceRdbUserPrivilegeID(region, locality.ExpandID(instanceID), databaseName, userName))
+	compositeID := locality.ExpandID(instanceID) + "/" + databaseName + "/" + userName
```
remyleone: We should have a dedicated function for composite ID and not have something that piggy back the RegionalIdentity

internal/services/rdb/user.go
```diff
@@ -148,7 +150,10 @@ func ResourceUserCreate(ctx context.Context, d *schema.ResourceData, m any) diag
 		return diag.FromErr(err)
 	}
 
-	d.SetId(ResourceUserID(region, locality.ExpandID(instanceID), user.Name))
+	compositeID := locality.ExpandID(instanceID) + "/" + user.Name
```
remyleone: We should have a dedicated identity function for multi-part id

internal/services/rdb/instance.go
```diff
@@ -46,9 +47,11 @@ func ResourceInstance() *schema.Resource {
...
 			StateContext: schema.ImportStatePassthroughContext,
 		},
-		SchemaVersion: 0,
-		SchemaFunc:    instanceSchema,
-		CustomizeDiff: cdf.LocalityCheck("private_network.#.pn_id"),
+		SchemaVersion:    0,
+		SchemaFunc:       instanceSchema,
+		CustomizeDiff:    cdf.LocalityCheck("private_network.#.pn_id"),
+		Identity:         identity.DefaultRegional(),
+		ResourceBehavior: schema.ResourceBehavior{MutableIdentity: true},
```
remyleone, line 54: Could you add a test for this behaviour?

jremy42: The test already exists here:
https://github.com/scaleway/terraform-provider-scaleway/blob/master/internal/services/rdb/instance_test.go#L1588

PR #3668, opened 2026-02-16
feat(mongodb): add identity

jremy42:


internal/services/mongodb/user.go
```diff
@@ -42,6 +43,7 @@ func ResourceUser() *schema.Resource {
 		},
 		SchemaVersion: 0,
 		SchemaFunc:    userSchema,
+		Identity:      identity.DefaultRegional(),
```
remyleone: We have a composite identity here, I'm not in favour to use the same setup here than with the other. I would like to have an explicit multi part regional here.

PR #3672, opened 2026-02-17
feat(edge_services): add support for plan identity

yfodil:


internal/identity/identity.go
```diff
@@ -89,18 +103,55 @@ func SetRegionalIdentity(d *schema.ResourceData, region scw.Region, id string) e
...
 	}
 
-	d.SetId(value)
+	d.SetId(id)
 
 	return nil
 }
+
+// SetMultiPartIdentity sets identity attributes and constructs a composite ID from multiple parts
+func SetMultiPartIdentity(d *schema.ResourceData, values map[string]string, keyOrder ...string) error {
```
remyleone, line 124: Could you add testing for this function? Between the key order and the map I can see several potential bug in case the order and map size are different and so on.

PR #3676, opened 2026-02-17
feat(cockpit): implement Identity

jremy42:


internal/services/cockpit/source.go
```diff
@@ -110,7 +112,9 @@ func ResourceCockpitSourceCreate(ctx context.Context, d *schema.ResourceData, me
 		return diag.FromErr(err)
 	}
 
-	d.SetId(regional.NewIDString(region, res.ID))
+	if err := identity.SetRegionalIdentity(d, region, res.ID); err != nil {
```
remyleone: Could you use the region coming from res?

internal/services/cockpit/source.go
```diff
@@ -135,6 +139,10 @@ func ResourceCockpitSourceRead(ctx context.Context, d *schema.ResourceData, meta
 		return diag.FromErr(err)
 	}
 
+	if err := identity.SetRegionalIdentity(d, region, res.ID); err != nil {
```
remyleone: Could you use the region coming from res?

internal/services/cockpit/token.go
```diff
@@ -182,7 +184,10 @@ func ResourceCockpitTokenCreate(ctx context.Context, d *schema.ResourceData, m a
 	}
 
 	_ = d.Set("secret_key", res.SecretKey)
-	d.SetId(regional.NewIDString(region, res.ID))
+
+	if err := identity.SetRegionalIdentity(d, region, res.ID); err != nil {
```
remyleone: Could you use the region coming from res?

internal/services/cockpit/token.go
```diff
@@ -207,6 +212,10 @@ func ResourceCockpitTokenRead(ctx context.Context, d *schema.ResourceData, m any
 		return diag.FromErr(err)
 	}
 
+	if err := identity.SetRegionalIdentity(d, region, res.ID); err != nil {
```
remyleone: Could you use the region coming from res?

PR #3677, opened 2026-02-18
feat(instance): add support for image identity

Mia-Cross:


internal/services/instance/image.go
```diff
@@ -222,9 +224,12 @@ func ResourceInstanceImageCreate(ctx context.Context, d *schema.ResourceData, m
 		return diag.FromErr(err)
 	}
 
-	d.SetId(zonal.NewIDString(zone, res.Image.ID))
+	err = identity.SetZonalIdentity(d, zone, res.Image.ID)
```
remyleone: Could you use the zone coming from the res variable?

internal/services/instance/image.go
```diff
@@ -257,22 +281,12 @@ func ResourceInstanceImageRead(ctx context.Context, d *schema.ResourceData, m an
...
-	_ = d.Set("tags", image.Image.Tags)
-	_ = d.Set("public", image.Image.Public)
-	_ = d.Set("creation_date", types.FlattenTime(image.Image.CreationDate))
-	_ = d.Set("modification_date", types.FlattenTime(image.Image.ModificationDate))
-	_ = d.Set("from_server_id", image.Image.FromServer)
-	_ = d.Set("state", image.Image.State)
-	_ = d.Set("zone", image.Image.Zone)
-	_ = d.Set("project_id", image.Image.Project)
-	_ = d.Set("organization_id", image.Image.Organization)
+	err = identity.SetZonalIdentity(d, zone, image.Image.ID)
```
remyleone: Could you use the zone coming from the image?

PR #3684, opened 2026-02-18
feat(edge_services): add support for route stage identity

yfodil:


remyleone (commented):
Could you remove the line of this resource in the test file?

remyleone (commented):
Could you remove the exception related to this resource in the sdkv2_test file?

PR #3688, opened 2026-02-19
feat(domain): add Identity

jremy42:


internal/services/domain/record.go
```diff
@@ -431,8 +452,73 @@ func resourceDomainRecordRead(ctx context.Context, d *schema.ResourceData, m any
...
+
+	if len(res.Records) == 0 {
+		d.SetId("")
+
+		return nil
+	}
+
+	record := res.Records[0]
+
+	d.SetId(fmt.Sprintf("%s/%s", dnsZone, record.ID))
```
remyleone: Why is there no call to `identity.SetMultiPartIdentity` ?

internal/services/domain/record.go
```diff
@@ -526,7 +553,5 @@ func resourceDomainRecordDelete(ctx context.Context, d *schema.ResourceData, m a
 		return diag.FromErr(err)
 	}
 
-	d.SetId("")
```
remyleone, line 529: Why is it removed?

PR #3691, opened 2026-02-20
feat(lb): add support for backend identity

yfodil:


internal/services/lb/backend.go
```diff
@@ -458,7 +460,10 @@ func resourceLbBackendCreate(ctx context.Context, d *schema.ResourceData, m any)
 		return diag.FromErr(err)
 	}
 
-	d.SetId(zonal.NewIDString(zone, res.ID))
+	err = identity.SetZonalIdentity(d, zone, res.ID)
```
remyleone: Could you use the zone coming from the res?

PR #3709, opened 2026-02-23
feat(vpc): add support for list resource

remyleone:


internal/locality/regional/model.go
```diff
@@ -0,0 +1,19 @@
...
+	GetRegion() string
+}
+
+// RegionsToQuery determines regions to query
+func RegionsToQuery(model Model) []scw.Region {
+	var regionsToQuery []scw.Region
+	if model.GetRegion() == "all" {
+		regionsToQuery = scw.AllRegions
+	} else {
+		regionsToQuery = []scw.Region{scw.Region(model.GetRegion())}
```
estellesoulard: add support for several regions?

internal/services/vpc/vpc_list.go
```diff
@@ -0,0 +1,228 @@
...
+	response.Schema = schema.Schema{
+		Attributes: map[string]schema.Attribute{
+			"name": schema.StringAttribute{
+				Description: "Name of the vpc to list for",
+				Optional:    true,
+			},
+			"organization_id": schema.StringAttribute{
+				Description: "Organization ID of the VPC to list for",
+				Optional:    true,
+			},
```
estellesoulard: validate uuid?

internal/services/vpc/vpc_list.go
```diff
@@ -0,0 +1,228 @@
...
+			"project_id": schema.StringAttribute{
+				Description: "Project ID of the VPC to list for",
+				Optional:    true,
+			},
```
estellesoulard: validate uuid?

internal/services/vpc/vpc_list.go
```diff
@@ -0,0 +1,228 @@
...
+			"is_default": schema.BoolAttribute{
+				Description: "Whether the VPC is the default VPC",
+				Optional:    true,
+			},
+			"region": schema.StringAttribute{
+				Description: "Region of the VPC. Use 'all' to list VPCs from all regions",
+				Optional:    true,
+				Validators: []validator.String{
+					stringvalidator.OneOf(append(regional.AllRegions(), "all")...),
+				},
```
estellesoulard: extract to a function?

internal/services/vpc/vpc_list.go
```diff
@@ -0,0 +1,228 @@
...
+	"fmt"
+
+	"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
+	"github.com/hashicorp/terraform-plugin-framework/list"
+	"github.com/hashicorp/terraform-plugin-framework/list/schema"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-log/tflog"
+	"github.com/hashicorp/terraform-plugin-mux/tf5to6server/translate"
```
estellesoulard, line 14: waiting for upstream merge

internal/locality/regional/model.go
```diff
@@ -0,0 +1,19 @@
+package regional
+
+import "github.com/scaleway/scaleway-sdk-go/scw"
+
+type Model interface {
+	GetRegion() string
```
estellesoulard: return string slice?

internal/services/vpc/vpc_list.go
```diff
@@ -0,0 +1,228 @@
...
+				Region:         region,
+				Name:           data.Name.ValueStringPointer(),
+				Tags:           tags,
+				OrganizationID: data.OrganizationID.ValueStringPointer(),
+				ProjectID:      data.ProjectID.ValueStringPointer(),
+				IsDefault:      data.IsDefault.ValueBoolPointer(),
+				RoutingEnabled: data.RoutingEnabled.ValueBoolPointer(),
+			}
+
+			vpcs, err := r.vpcAPI.ListVPCs(listRequest, scw.WithContext(ctx))
```
estellesoulard: with all pages

internal/services/vpc/vpc_list_test.go
```diff
@@ -0,0 +1,107 @@
...
+func TestAccListVPCs_Basic(t *testing.T) {
+	tt := acctest.NewTestTools(t)
+	defer tt.Cleanup()
+
+	resource.ParallelTest(t, resource.TestCase{
+		ProtoV6ProviderFactories: tt.ProviderFactories,
+		Steps: []resource.TestStep{
+			{
+				Config: `
+					resource "scaleway_account_project" "main" {}
```
estellesoulard, line 26: ensure project deletion, check destroy

internal/services/vpc/vpc_list.go
remyleone: Without news from them we forked

internal/services/vpc/vpc_list.go
```diff
@@ -0,0 +1,248 @@
...
+		stream.Results = list.ListResultsStreamDiagnostics(diag.Diagnostics{
+			diag.NewErrorDiagnostic("Listing projects", "An error was encountered when listing projects: "+err.Error()),
+		})
+
+		return
+	}
+
+	var targets []listscw.RegionalFetchTarget
+
+	for _, r := range regions {
```
jremy42, line 167: What happens if neither `regions` nor `project_ids` is specified? The list would silently return 0 results.

yfodil: Maybe it should fall back to the provider's default region and project ID when these fields are null

internal/list/fetch.go
```diff
@@ -0,0 +1,49 @@
...
+
+var defaultFetchLimit = runtime.NumCPU()
+
+// FetchFunc defines a function that fetches items for a given target.
+type FetchFunc[T any, Target any] func(ctx context.Context, target Target) ([]T, error)
+
+// FetchConcurrently fetches items for the provided targets concurrently,
+// limiting the number of active goroutines to the default fetch limit.
+// It returns all fetched items combined into a single slice.
+func FetchConcurrently[T any, Target any](ctx context.Context, targets []Target, fetch FetchFunc[T, Target]) ([]T, error) {
```
yfodil: question: this function appends results from goroutines in arbitrary order, so listing the same VPCs twice can return them in different orders

should we consider sorting the result after fetching the data (by ID or creation date or ...?) for consistency?

internal/services/vpc/vpc_list.go
```diff
@@ -0,0 +1,248 @@
...
+	"github.com/scaleway/terraform-provider-scaleway/v2/internal/meta"
+)
+
+var (
+	_ list.ListResource                 = (*ListResource)(nil)
+	_ list.ListResourceWithConfigure    = (*ListResource)(nil)
+	_ list.ListResourceWithRawV6Schemas = (*ListResource)(nil)
+)
+
+type ListResource struct {
```
yfodil:
```suggestion
type VPCListResource struct {
```
The name is generic, will collide if a second list resource like private networks is added to the vpc package later

remyleone: The problem is that it will break a linter because the type now starts with the same as the package name. Either an exception in the linter is required or something that is avoid the issue like `ListResourceForVPC` there will be no collision if you use `GatewayListResource` or `PrivateNetworkListResource` for instance

internal/list/fetch.go
remyleone: Excellent remark. I'm not sure how much order will matter for importation though. That being said we can use a lexicographical order locality first, id second to guarantee and order. Because all components that are in identity will be comparable an order is possible.

internal/services/vpc/vpc_list.go
remyleone: Excellent remark I will think of a fallback

PR #3716, opened 2026-02-24
feat(tem): add identity for domain_validation

jremy42:


internal/verify/ip.go
```diff
@@ -30,8 +30,8 @@ func IsStandaloneIPorCIDR() schema.SchemaValidateDiagFunc {
 		if err != nil {
 			return diag.Diagnostics{diag.Diagnostic{
 				Severity:      diag.Error,
-				Summary:       "neither a valid IP address or CIDR notation: " + ip,
 				AttributePath: path,
+				Summary:       "neither a valid IP address or CIDR notation: " + ip,
```
remyleone, line 33: What problem this change helps solve?

jremy42: Reverted. This change was likely noise from a bad rebase

PR #3749, opened 2026-03-02
feat(autoscaling): add support for group identity

yfodil:


internal/services/autoscaling/instance_group.go
```diff
@@ -178,6 +183,17 @@ func ResourceInstanceGroupRead(ctx context.Context, d *schema.ResourceData, m an
...
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	return diags
+}
+
+func setInstanceGroupState(d *schema.ResourceData, group *autoscaling.InstanceGroup, zone scw.Zone) diag.Diagnostics {
 	_ = d.Set("name", group.Name)
 	_ = d.Set("template_id", zonal.NewIDString(zone, group.InstanceTemplateID))
```
remyleone: Could the zone be deduced from the group variable and avoid passing it as an extra argument? Is it possible to have groups that are cross zones?

PR #3750, opened 2026-03-02
feat(autoscaling): add support for policy identity

yfodil:


internal/services/autoscaling/instance_policy.go
remyleone: Could you use the zone in the policy variable and remove this argument from the function signature?

PR #3766, opened 2026-03-09
feat(mnq): add identity support to MNQ NATS resources

jremy42:


internal/services/mnq/nats_account.go
```diff
@@ -60,7 +62,9 @@ func ResourceMNQNatsAccountCreate(ctx context.Context, d *schema.ResourceData, m
 		return diag.FromErr(err)
 	}
 
-	d.SetId(regional.NewIDString(region, account.ID))
+	if err := identity.SetRegionalIdentity(d, region, account.ID); err != nil {
```
remyleone:
```suggestion
	if err := identity.SetRegionalIdentity(d, account.Region, account.ID); err != nil {
```

internal/services/mnq/nats_account.go
```diff
@@ -85,6 +89,10 @@ func ResourceMNQNatsAccountRead(ctx context.Context, d *schema.ResourceData, m a
 		return diag.FromErr(err)
 	}
 
+	if err := identity.SetRegionalIdentity(d, region, id); err != nil {
```
remyleone:
```suggestion
	if err := identity.SetRegionalIdentity(d, account.Region, account.ID); err != nil {
```

PR #3825, opened 2026-03-19
fix(instance): set public_ips in datasource as optional

estellesoulard:


internal/services/instance/servers_data_source_test.go
```diff
@@ -209,3 +209,121 @@ func TestAccDataSourceServers_PrivateIPs(t *testing.T) {
...
+					data "scaleway_instance_servers" "servers" {
+					}
+
+					data "scaleway_instance_servers" "server0" {
+						name = scaleway_instance_servers.server0.name
+					}
+
+					data "scaleway_instance_servers" "server1" {
+						name = scaleway_instance_servers.server1.name
+					}
```
estellesoulard: Running this test locally with outputs for each datasource, I get the following output:

```
servers = {
  "id" = "fr-par-1"
  "name" = tostring(null)
  "organization_id" = tostring(null)
  "project_id" = tostring(null)
  "servers" = tolist(null) /* of object */
  "tags" = tolist(null) /* of string */
  "zone" = tostring(null)
}
server0 = {
  "id" = "fr-par-1"
  "name" = "tf-acc-server-ips-0"
  "organization_id" = tostring(null)
  "project_id" = tostring(null)
  "servers" = tolist(null) /* of object */
  "tags" = tolist(null) /* of string */
  "zone" = tostring(null)
}
server1 = {
  "id" = "fr-par-1"
  "name" = "tf-acc-server-ips-1"
  "organization_id" = tostring(null)
  "project_id" = tostring(null)
  "servers" = tolist(null) /* of object */
  "tags" = tolist(null) /* of string */
  "zone" = tostring(null)
}
```

estellesoulard: If i remove the `ip_ids = [scaleway_instance_ip.ip1.id]` line from all server resources, the outputs are complete. (given there is NO instance with a public ip on the target organization)

PR #3835, opened 2026-03-23
feat(vpc): add support for vpc peering connectors

yfodil:


internal/services/vpc/connector.go
```diff
@@ -0,0 +1,207 @@
...
+			Elem: &schema.Schema{
+				Type: schema.TypeString,
+			},
+		},
+		"vpc_id": {
+			Type:        schema.TypeString,
+			Optional:    true,
+			Computed:    true,
+			ForceNew:    true,
+			Description: "The VPC in which to create the private network",
```
estellesoulard: discrepancy between the schema description and the description in the docs  (from what I see in the create I think it should be required)

internal/services/vpc/connector.go
```diff
@@ -0,0 +1,207 @@
...
+		},
+		"target_vpc_id": {
+			Type:        schema.TypeString,
+			Optional:    true,
+			Computed:    true,
+			ForceNew:    true,
+			Description: "The VPC in which to create the private network",
```
estellesoulard: discrepancy between the schema description and the description in the docs

internal/services/vpc/connector.go
```diff
@@ -0,0 +1,207 @@
...
+			Type:        schema.TypeList,
+			Optional:    true,
+			Description: "The tags associated with the VPC connector",
+			Elem: &schema.Schema{
+				Type: schema.TypeString,
+			},
+		},
+		"vpc_id": {
+			Type:        schema.TypeString,
+			Optional:    true,
```
estellesoulard: marked required in docs, but optional here? (from what I see in the create I think it should be required)

internal/services/vpc/connector.go
```diff
@@ -0,0 +1,207 @@
...
+			Computed:    true,
+			ForceNew:    true,
+			Description: "The VPC in which to create the private network",
+		},
+		"target_vpc_id": {
+			Type:        schema.TypeString,
+			Optional:    true,
```
estellesoulard: marked required in docs vs optional here

PR #3840, opened 2026-03-26
feat(domain): Added ressources for external domain management

lde:
Due to the way actual dns management zones work at Scaleway test is broken. and can't be fixed. even if resource is working well. For testing root zone must be hosted outside Scaleway. 

> [!IMPORTANT]
> This PR depend on this pr : https://github.com/scaleway/scaleway-sdk-go/pull/3005 I made.

### DIFF
--- a/docs/resources/domain_external_domain.md
+++ b/docs/resources/domain_external_domain.md
@@ -1,0 +1,54 @@
+---
+subcategory: "Domains and DNS"
+page_title: "Scaleway: scaleway_domain_external_domain"
+---
+
+# Resource: scaleway_domain_external_domain
+
+The `scaleway_domain_external_domain` resource allows you to register an external domain name to be managed by Scaleway.
+This is useful when you want to use Scaleway DNS for a domain registered with another registrar.
+
+Once registered, you will receive a `validation_token` that you must add as a TXT record to your domain's current DNS configuration to prove ownership.
+After ownership is verified, you can point your domain's NS records to the `ns_servers` provided by Scaleway.
+
+Refer to the [Domains and DNS documentation](https://www.scaleway.com/en/docs/network/domains-and-dns/) and the [API documentation](https://developers.scaleway.com/) for more details.
+
+## Example Usage
+
+### Register an External Domain
+
+The following example registers an external domain and retrieves its validation token.
+
+```terraform
+resource "scaleway_domain_external_domain" "example" {
+  domain = "example.com"
+}
+
+output "validation_token" {
+  value = scaleway_domain_external_domain.example.validation_token
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+- `domain` (Required, String): The domain name to be registered.
+- `project_id` (Optional, String): The Scaleway project ID.
+
+## Attributes Reference
+
+In addition to all arguments above, the following attributes are exported:
+
+- `id`: The domain name.
+- `ns_servers`: List of NS servers for the domain once it is validated.
+- `status`: The status of the domain (e.g., `checking`, `active`).
+- `validation_token`: The validation token to be added as a TXT record to the domain's DNS to verify ownership.
+
+## Import
+
+To import an existing external domain, use:
+
+```bash
+terraform import scaleway_domain_external_domain.example example.com
+```
--- a/docs/resources/domain_external_domain_validated.md
+++ b/docs/resources/domain_external_domain_validated.md
@@ -1,0 +1,78 @@
+---
+subcategory: "Domains and DNS"
+page_title: "Scaleway: scaleway_domain_external_domain_validated"
+---
+
+# Resource: scaleway_domain_external_domain_validated
+
+The `scaleway_domain_external_domain_validated` resource allows you to wait until an external domain is validated by Scaleway.
+This resource is used to block the Terraform execution until the validation token has been correctly added to the external domain's DNS records and verified by Scaleway.
+
+
+~> **WARNING:** This resource implements a part of the validation workflow. It does not represent a real-world entity in Scaleway, therefore changing or deleting this resource on its own has no effect. It must be used with a `scaleway_domain_external_domain` resource. 
+
+
+Once the domain is validated, its status becomes `active`, and it can be used within Scaleway services.
+
+Refer to the [Domains and DNS documentation](https://www.scaleway.com/en/docs/network/domains-and-dns/) and the [API documentation](https://developers.scaleway.com/) for more details.
+
+## Example Usage
+
+### Wait for an External Domain to be Validated
+
+The following example registers an external domain and then waits for it to be validated.
+
+```terraform
+
+resource "scaleway_domain_external_domain" "example" {
+  domain = "scw.example.com"
+}
+resource "scaleway_domain_record" "validation" {
+  dns_zone= "example.com"
+  name = "_scaleway-challenge.scw"
+  type = "TXT"
+  data = scaleway_domain_external_domain.example.validation_token
+}
+
+# Wait for domain validation (default timeout is 30 minutes)
+resource "scaleway_domain_external_domain_validated" "example" {
+  domain = scaleway_domain_external_domain.example.domain
+}
+
+# After validation, we can use the domain's NS servers
+resource "scaleway_domain_record" "NS" {
+  for_each = {ns0=scaleway_domain_external_domain_validated.example.ns_servers[0],
+    ns1=scaleway_domain_external_domain_validated.example.ns_servers[1]}
+  dns_zone= "example.com"
+  name = "scw"
+  type = "NS"
+  data = each.value
+}
+
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+- `domain` (Required, String): The domain name to be validated.
+- `project_id` (Optional, String): The Scaleway project ID.
+
+## Attributes Reference
+
+In addition to all arguments above, the following attributes are exported:
+
+- `id`: The domain name.
+- `ns_servers`: List of NS servers for the domain once it is validated.
+
+## Timeouts
+
+The `timeouts` block allows you to specify a custom duration for waiting until the domain is validated.
+Default timeout is 30 minutes.
+
+- `create`: Custom duration for the validation process.
+
+
+```bash
+terraform import scaleway_domain_external_domain_validated.example example.com
+```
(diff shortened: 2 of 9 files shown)


internal/services/domain/external_domain.go
```diff
@@ -0,0 +1,116 @@
...
+	return &schema.Resource{
+		CreateContext: resourceExternalDomainCreate,
+		ReadContext:   resourceExternalDomainRead,
+		DeleteContext: resourceExternalDomainDelete,
+
+		Importer: &schema.ResourceImporter{
+			StateContext: schema.ImportStatePassthroughContext,
+		},
+
+		Schema: map[string]*schema.Schema{
```
jremy42, line 25: Please follow the current domain resource pattern (`zone` / `registration`):
- `SchemaFunc` instead of inline `Schema`
- `Identity: identity.DefaultGlobal()`
- `Timeouts` if documented
- `organization_id: account.OrganizationIDSchema()`

New SDKv2 resources in this package should not use the older `Schema:` style.

internal/services/domain/external_domain.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+
+	domainName := d.Get("domain").(string)
+
+	_, err = registrarAPI.RegisterExternalDomain(&domain.RegistrarAPIRegisterExternalDomainRequest{
+		Domain:    domainName,
+		ProjectID: projectID,
+	}, scw.WithContext(ctx))
+	if err != nil {
+		return diag.FromErr(err)
```
jremy42, line 68: `RegisterExternalDomainResponse` already returns `ValidationToken` / `OrganizationID` / `ProjectID`. Prefer using the create response (at least to seed `validation_token`) instead of discarding it with `_`.

Also set the ID via `identity.SetGlobalIdentity(d, domainName)` rather than only `d.SetId`.

internal/services/domain/external_domain.go
```diff
@@ -0,0 +1,116 @@
...
+
+	return resourceExternalDomainRead(ctx, d, m)
+}
+
+func resourceExternalDomainRead(ctx context.Context, d *schema.ResourceData, m any) diag.Diagnostics {
+	registrarAPI := NewRegistrarDomainAPI(m)
+	resp, err := registrarAPI.GetDomain(&domain.RegistrarAPIGetDomainRequest{Domain: d.Id()}, scw.WithContext(ctx))
+	if err != nil {
+		if httperrors.Is404(err) {
+			d.SetId("")
```
jremy42, line 80: On 404, please log a warning before removing from state, e.g.
`log.Printf("[WARN] domain external domain (%s) not found, removing from state", d.Id())`
(see common review items).

internal/services/domain/external_domain.go
```diff
@@ -0,0 +1,116 @@
...
+			return nil
+		}
+		return diag.FromErr(err)
+	}
+	_ = d.Set("domain", resp.Domain)
+	_ = d.Set("project_id", resp.ProjectID)
+	if len(resp.DNSZones) > 0 {
+		_ = d.Set("ns_servers", resp.DNSZones[0].NsDefault)
+	}
+	_ = d.Set("status", resp.Status)
```
jremy42, line 90: Use `resp.Status.String()` like `zone.go`.

Also always `d.Set("validation_token", ...)` (including empty string) so the token does not go stale after validation.

For `ns_servers` (TypeList), check the `d.Set` error.

internal/services/domain/external_domain.go
```diff
@@ -0,0 +1,116 @@
...
+	if resp.ExternalDomainRegistrationStatus != nil && resp.ExternalDomainRegistrationStatus.ValidationToken != "" {
+		_ = d.Set("validation_token", resp.ExternalDomainRegistrationStatus.ValidationToken)
+	}
+
+	return nil
+}
+
+func resourceExternalDomainDelete(ctx context.Context, d *schema.ResourceData, m any) diag.Diagnostics {
+	if d.Get("status") == "pending" {
```
jremy42, line 99: `pending` is not a valid `DomainStatus` in the SDK. External domains use values like `checking` / `active` / `creating` / …

This guard is effectively dead code (and checks stale state via `d.Get` instead of the API). Please remove it or compare against real SDK constants after a fresh Get.

internal/services/domain/external_domain_validated.go
```diff
@@ -0,0 +1,62 @@
...
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/scaleway/terraform-provider-scaleway/v2/internal/services/account"
+)
+
+func ResourceExternalDomainValidated() *schema.Resource {
+	return &schema.Resource{
+		CreateContext: resourceExternalDomainValidatedCreate,
+		ReadContext:   resourceExternalDomainRead,
```
jremy42, line 14: **Bug:** `ReadContext: resourceExternalDomainRead` reuses the parent resource Read, which sets `status` / `validation_token` — attributes that are not in this schema.

Please give `*_validated` its own Read (and ideally a `validated` computed bool, like TEM/mailbox domain validation).

internal/services/domain/external_domain_validated.go
```diff
@@ -0,0 +1,62 @@
...
+				Computed:    true,
+			},
+		},
+	}
+}
+
+func resourceExternalDomainValidatedCreate(ctx context.Context, d *schema.ResourceData, m any) diag.Diagnostics {
+	registrarAPI := NewRegistrarDomainAPI(m)
+	domainName := d.Get("domain").(string)
+	_, err := waitForExternalDomainValidation(ctx, registrarAPI, domainName, defaultDomainRegistrationTimeout)
```
jremy42, line 50: Use `d.Timeout(schema.TimeoutCreate)` instead of hardcoding `defaultDomainRegistrationTimeout`, and declare Create/Delete timeouts on the resource (wait-only resources still expose Create timeout).

internal/services/domain/external_domain_validated.go
```diff
@@ -0,0 +1,62 @@
...
+		Timeouts: &schema.ResourceTimeout{
+			Default: schema.DefaultTimeout(defaultDomainRegistrationTimeout),
+		},
+
+		Importer: &schema.ResourceImporter{
+			StateContext: schema.ImportStatePassthroughContext,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"domain": {
```
jremy42, line 25: Same as the parent resource: prefer `SchemaFunc` + `Identity: identity.DefaultGlobal()` + `SetGlobalIdentity`.

Also add `depends_on = [scaleway_domain_record.validation]` in docs/examples so Create does not race the TXT challenge.

internal/services/domain/external_domain_validated_test.go
```diff
@@ -0,0 +1,101 @@
...
+  name       = "_scaleway-challenge.%s"
+  type       = "TXT"
+  data       = scaleway_domain_external_domain.example.validation_token
+ provider    = scaleway.alt
+}
+
+resource "scaleway_domain_external_domain_validated" "example" {
+  domain = scaleway_domain_external_domain.example.domain
+}
+
```
jremy42, line 93: This `provider "scaleway" { alias = ... }` block triggers AT004 in our acceptance-test lint rules (configs must not declare providers).

Also the recorded cassette returns 403 (`parent domain already registered`), so replay cannot pass. External-domain tests need a FQDN whose parent is **not** registered at Scaleway (and a way to publish the TXT challenge without an in-config provider alias).

Please also add:
- `CheckDestroy` ✅ (already there)
- an Exists check
- `ImportState` / `ImportStateVerify` steps (Importer is declared)

docs/resources/domain_external_domain.md
```diff
@@ -0,0 +1,54 @@
...
+
+- `domain` (Required, String): The domain name to be registered.
+- `project_id` (Optional, String): The Scaleway project ID.
+
+## Attributes Reference
+
+In addition to all arguments above, the following attributes are exported:
+
+- `id`: The domain name.
+- `ns_servers`: List of NS servers for the domain once it is validated.
```
jremy42, line 44: Docs/schema mismatch: this page should be generated from `templates/resources/domain_external_domain.md.tmpl` via `make docs`.

Also missing from the resource schema today: `organization_id`, and any `timeouts` block (if you document timeouts, implement them on the resource).

docs/resources/domain_external_domain_validated.md
```diff
@@ -0,0 +1,78 @@
...
+}
+resource "scaleway_domain_record" "validation" {
+  dns_zone= "example.com"
+  name = "_scaleway-challenge.scw"
+  type = "TXT"
+  data = scaleway_domain_external_domain.example.validation_token
+}
+
+# Wait for domain validation (default timeout is 30 minutes)
+resource "scaleway_domain_external_domain_validated" "example" {
```
jremy42, line 38: Please add `depends_on = [scaleway_domain_record.validation]` here so validation waiting cannot start before the TXT record exists.

Also consider exporting a computed `validated` attribute for consistency with `scaleway_tem_domain_validation` / `scaleway_mailbox_domain_validation`.

jremy42 (changes requested):
Thanks for the contribution — the overall split (`external_domain` + wait-only `*_validated`) matches the TEM/mailbox validation pattern and is the right direction.

Before we can merge, the implementation needs to be aligned with our current domain resource conventions (see `zone.go` / `registration.go` / `mailbox/domain_validation.go`) and the [common review items](https://github.com/scaleway/terraform-provider-scaleway/blob/main/.github/contributing/common_reviews_items.md). CI is also red on lint/docs/tests.

### Blockers
1. Align resources with `SchemaFunc` + `Identity: identity.DefaultGlobal()` + `identity.SetGlobalIdentity` (not raw `Schema:` / `d.SetId` only).
2. `*_validated` must not reuse `resourceExternalDomainRead` (it sets attributes that are not in its schema).
3. Bump `scaleway-sdk-go` in `go.mod` (PR depends on https://github.com/scaleway/scaleway-sdk-go/pull/3005).
4. Fix/remove the broken cassette (403 parent already registered) and avoid AT004 (`provider` blocks in test configs).
5. Add `templates/resources/*.md.tmpl` and keep docs in sync (`make docs`); docs currently claim attributes/timeouts that are missing from the schema.
6. Add `ImportState` test steps (Importer is declared).

Also please rebase on `main` instead of merge commits, and prefer American English / consistent naming in descriptions.

Happy to re-review once these are addressed.

PR #3852, opened 2026-03-30
feat(k8s): add pool specs validation

Mia-Cross:
This PR adds validation for the following specs of a k8s pool:
- pool `root_volume_type` must be compatible with the type of storage supported by the `node_type`
- pool `root_volume_size` must be at least 20GB, and lower than the max capacity of the `node_type` when using local storage.
- pool `size` must be greater than 0 on mutualized clusters, except if the cluster already has at least another node on another pool. (So far the provider did not allow size 0 pools so we had to change the default value of `min_size`)

It also documents the recommended usage of the `create_before_destroy` lifecycle rule to reduce the impact of failures in a pool's update when the requested change requires the resource's replacement.

### DIFF
--- a/internal/services/k8s/pool.go
+++ b/internal/services/k8s/pool.go
@@ -4,10 +4,13 @@ import (
 	"context"
 	_ "embed"
 	"fmt"
+	"strings"
 	"time"
 
+	"github.com/hashicorp/go-cty/cty"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/scaleway/scaleway-sdk-go/api/instance/v1"
 	"github.com/scaleway/scaleway-sdk-go/api/k8s/v1"
 	"github.com/scaleway/scaleway-sdk-go/scw"
 	"github.com/scaleway/terraform-provider-scaleway/v2/internal/dsf"
@@ -19,6 +22,11 @@ import (
 	"github.com/scaleway/terraform-provider-scaleway/v2/internal/services/ipam"
 	"github.com/scaleway/terraform-provider-scaleway/v2/internal/types"
 	"github.com/scaleway/terraform-provider-scaleway/v2/internal/verify"
+)
+
+const (
+	nodeMinVolumeSize = 20 * scw.GB
+	nodeMaxVolumeSize = 10 * scw.TB
 )
 
 //go:embed descriptions/pool.md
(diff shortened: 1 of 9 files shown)


internal/services/k8s/pool.go
```diff
@@ -628,3 +662,150 @@ func ResourceK8SPoolCustomDiff(_ context.Context, diff *schema.ResourceDiff, _ a
...
+				diag.Diagnostic{
+					Severity:      diag.Error,
+					Summary:       "Invalid root volume size",
+					Detail:        fmt.Sprintf("requested size must be at least 20GB, got: %d,", *req.RootVolumeSize/scw.GB),
+					AttributePath: cty.GetAttrPath("root_volume_size"),
+				},
+			}
+		}
+
+		if req.RootVolumeType == k8s.PoolVolumeTypeLSSD {
```
remyleone: Is there specific checks to perform for sbs volume types? If so, I would be more confortable with a switch case with a case for lssd, a bssd and default that returns an error in case we are not in those cases.

internal/services/k8s/pool.go
```diff
@@ -628,3 +662,150 @@ func ResourceK8SPoolCustomDiff(_ context.Context, diff *schema.ResourceDiff, _ a
...
+			return diag.Diagnostics{
+				diag.Diagnostic{
+					Severity: diag.Warning,
+					Summary:  "Could not validate root volume specs",
+					Detail:   "Server type had no volume constraints to check",
+				},
+			}
+		}
+
+		if req.RootVolumeSize != nil && *req.RootVolumeSize < 20*scw.GB {
```
remyleone: Could the 20*scw.GB be extracted and set up as a const in the package with the right documentation?

internal/services/k8s/pool.go
```diff
@@ -628,3 +667,171 @@ func ResourceK8SPoolCustomDiff(_ context.Context, diff *schema.ResourceDiff, _ a
...
+				},
+			}
+		}
+
+		if req.RootVolumeSize != nil && *req.RootVolumeSize < nodeMinVolumeSize {
+			return diag.Diagnostics{
+				diag.Diagnostic{
+					Severity:      diag.Error,
+					Summary:       "Invalid root volume size",
+					Detail:        fmt.Sprintf("requested size must be at least 20GB, got: %d,", *req.RootVolumeSize/scw.GB),
```
louisportay: maybe reuse `nodeMinVolumeSize` in the message ?

internal/services/k8s/pool.go
Mia-Cross: New checks have been added for sbs volumes

PR #3874, opened 2026-03-31
feat(interlink): add partner datasources

yfodil:


docs/data-sources/interlink_partner.md
```diff
@@ -0,0 +1,45 @@
...
+data "scaleway_interlink_partner" "by_name" {
+  name = "Equinix"
+}
+```
+
+
+
+## Argument Reference
+
+- `partner_id` - (Optional) The ID of the partner. Conflicts with `name`.
```
estellesoulard: maybe for clarity specify it could be either uuid or regional uuid ?

internal/services/interlink/partner_data_source.go
```diff
@@ -0,0 +1,148 @@
...
+	partnerID, idExists := d.GetOk("partner_id")
+	if idExists {
+		return dataSourcePartnerReadByID(ctx, d, m, partnerID.(string))
+	}
+
+	return dataSourcePartnerReadByFilters(ctx, d, m)
+}
+
+func dataSourcePartnerReadByID(ctx context.Context, d *schema.ResourceData, m any, partnerID string) diag.Diagnostics {
+	api, region, err := interlinkAPIWithRegion(d, m)
```
estellesoulard, line 81: partner_id accepts a regional uuid but the region is never derived from it, while some other resources with a localized uuid base the region by priority on : the region argument > the uuid region > the provider region (like in [StartJobDefinitionAction](https://github.com/scaleway/terraform-provider-scaleway/blob/02ec3ccfc284ac637fe51fc14ef4e962592adaa9/internal/services/jobs/job_definition_start_action.go#L150) for example)

internal/services/interlink/partner_data_source_test.go
```diff
@@ -0,0 +1,32 @@
+package interlink_test
+
+import (
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/scaleway/terraform-provider-scaleway/v2/internal/acctest"
+)
+
+func TestAccDataSourceInterlinkPartner_ByName(t *testing.T) {
```
estellesoulard, line 10: add test by id?

internal/services/interlink/partners_data_source_test.go
```diff
@@ -0,0 +1,29 @@
+package interlink_test
+
+import (
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/scaleway/terraform-provider-scaleway/v2/internal/acctest"
+)
+
+func TestAccDataSourceInterlinkPartners_Basic(t *testing.T) {
```
estellesoulard, line 10: add test by ids?

yfodil: Pops datasources are not implemented yet, will add this test later

PR #3936, opened 2026-04-10
fix(mnq): tests serial and sns project name

jremy42:


remyleone (changes requested):
Could you ensure that the test executes quickly either with parallel tests or with retry that are smaller, the test suite is now taking 6 minutes instead of something quick

remyleone (approved):

PR #3993, opened 2026-05-12
feat(object): add support for bucket `noncurrent_versions`

pypaut:
Add support for `noncurrent_version_transition` and `noncurrent_version_expiration` blocks, both part of the `lifecycle_rule` block.


pypaut:
Just rebased `main` and re-edited each commits to sign with GPG.

remyleone (approved):

PR #3998, opened 2026-05-13
fix(object): add support for bucket encryption region

pypaut:
Add optional field `region` to the `scaleway_object_bucket_server_side_encryption_configuration` resource.


remyleone (approved):

vdhpieter:
@pypaut when will this be released?

PR #4005, opened 2026-05-19
feat(object): add support for bucket SSE-KMS

pypaut:
- Update the `scaleway_bucket_server_side_configuration` schema, with parsing
- Add tests
- Give up on adding a `CustomizeDiff` validation function, was too complicated and already handled by the backend
- Add examples to the documentation


examples/resources/scaleway_object_bucket_server_side_encryption_configuration/resource-kms.tf
```diff
@@ -0,0 +1,18 @@
+resource "scaleway_object_bucket" "test" {
```
remyleone, line 1: The template should use those examples to generate a documentation

remyleone (changes requested):

examples/resources/scaleway_object_bucket_server_side_encryption_configuration/resource-kms.tf
pypaut: Ah yes forgot the little `make docs`

remyleone (approved):

PR #4008, opened 2026-05-20
docs(object): add bucket examples for lifecycle rules

pypaut:


Monitob:
Perfect

examples/resources/scaleway_object_bucket/resource.tf
```diff
@@ -0,0 +1,77 @@
+resource "scaleway_object_bucket" "main" {
```
remyleone: This would need to be picked up by the template to after being rendered in the documentation

remyleone: Here is an example of a PR that use examples in the template file for the documentation https://github.com/scaleway/terraform-provider-scaleway/pull/4006/changes

remyleone (changes requested):

remyleone (approved):

PR #4013, opened 2026-05-20
feat(iam): add user list resource

estellesoulard:
https://github.com/scaleway/terraform-provider-scaleway/issues/3884


internal/services/iam/user.go
```diff
@@ -294,3 +286,25 @@ func resourceIamUserDelete(ctx context.Context, d *schema.ResourceData, m any) d
 
 	return nil
 }
+
+func setUserState(d *schema.ResourceData, user *iam.User) diag.Diagnostics {
```
estellesoulard:
```suggestion
func setUserState(d *schema.ResourceData, user *iam.User) {
```

internal/services/iam/user_data_source.go
```diff
@@ -83,35 +76,39 @@ func DataSourceIamUserRead(ctx context.Context, d *schema.ResourceData, m any) d
...
-			return diag.Errorf("no user found with the email %s", d.Get("email"))
+		if userID == nil {
+			return diag.Errorf("no user found with the email %s", email)
 		}
 	}
 
 	d.SetId(userID.(string))
 
-	err := d.Set("user_id", userID)
+	err = d.Set("user_id", userID)
```
pypaut, line 99: What is the difference between `d.Set("user_id", userID)` and `d.SetId(userID.(string))`?

internal/services/iam/user_list_test.go
```diff
@@ -0,0 +1,121 @@
...
+			},
+			{
+				Query: true,
+				Config: `
+					list "scaleway_iam_user" "by_mfa" {
+						provider = scaleway
+
+						config {
+							organization_id = scaleway_iam_user.user1.organization_id
+							mfa              = false
```
pypaut, line 111: Could you add a test with `mfa = true`, to make sure it actually filters out the user?
And same with users using MFA.

internal/services/iam/user_data_source.go
remyleone: Set is for a standard attribute,  SetId is for the legacy Id system in terraform.

remyleone (approved):

Issue #3985, opened 2026-05-06
scaleway_object_bucket_server_side_encryption_configuration does not support project_id


vdhpieter:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform Version

```
OpenTofu v1.11.5
on darwin_arm64
+ provider registry.opentofu.org/cyrilgdn/postgresql v1.26.0
+ provider registry.opentofu.org/hashicorp/helm v3.1.1
+ provider registry.opentofu.org/hashicorp/kubernetes v3.1.0
+ provider registry.opentofu.org/hashicorp/random v3.8.1
+ provider registry.opentofu.org/scaleway/scaleway v2.74.0
```

### Affected Resource(s)

* scaleway_object_bucket_server_side_encryption_configuration

### Terraform Configuration Files

```hcl
resource "scaleway_object_bucket_server_side_encryption_configuration" "test" {
  bucket = scaleway_object_bucket.test.id

  rule {
    apply_server_side_encryption_by_default {
      sse_algorithm = "AES256"
    }
  }
}
```

### Debug Output

https://gist.github.com/vdhpieter/f42cf3f9f9c7bd257348eea4d4987dd4

### Expected Behavior

I should be able to specify the project id when configuring server side encryption.

### Actual Behavior

I'm not able to configure a project ID so when I try to configure sever side encryption for a bucket in not the default project I get a 403.

### Steps to Reproduce

1. `terraform apply`

### Other

I was doubting a lot between feature request vs bug but given almost all other resources support a project id when needed I'm leaning more to the bug side.

kirby0025:
Hello,
Not sure if it's exactly the same issue, but the result is similar at least : I have a 403 when trying to add a scaleway_object_bucket_server_side_encryption_configuration resource to a bucket.

It may be different because the error I get is a SignatureDoesNotMatch : 
[...]

PR #4015, opened 2026-05-21
feat(object): add support for encryption project_id

pypaut:
Address #3985.


pypaut:
Got too complicated and I don't want to deal with a rebase. Closing and continuing on #4020.

PR #4020, opened 2026-05-22
feat(object): enhance support for "project_id" in bucket & sub-resources

pypaut:
Addresses #3985

* Add identity, add `project_id` to resource ID of the bucket, its SSE, and its ACL
* Add `project_id` to the SSE schema
* Add test cases handling and checking the `project_id`
* Remove "acl" from the resource ID of the ACL resource


pypaut:
Tests broken in the CI but working locally. Updated the cassettes for each breaking test, no luck.

pypaut:
The changes in this PR got out of scope. I might need to start over

internal/locality/regional/ids.go
```diff
@@ -10,8 +10,9 @@ import (
 
 // ID represents an ID that is linked with a region, eg fr-par/11111111-1111-1111-1111-111111111111
 type ID struct {
-	ID     string
-	Region scw.Region
+	ID        string
+	ProjectID string
```
remyleone, line 14: Revert this change, many id do not require a project id to be identified. Therefore this should not be here. An alternative could be to have a dedicated struct for bucket but I'm not even sure it is required

internal/services/object/bucket.go
```diff
@@ -307,36 +309,46 @@ func bucketSchema() map[string]*schema.Schema {
...
 
 	if err != nil {
 		return diag.FromErr(err)
 	}
 
-	d.SetId(regional.NewIDString(region, bucketName))
+	projectId := d.Get("project_id").(string)
+
+	if projectId != "" {
+		err = identity.SetRegionalIdentity(d, region, bucketName+"@"+projectId)
```
remyleone, line 344: We should have a SetRegionalIdentityWithProject in this context to be able to clearly identify the resource that are using this format. Similar to a function like SetMultiPartIdentity

internal/services/object/bucket.go
```diff
@@ -745,15 +757,19 @@ func resourceObjectBucketRead(ctx context.Context, d *schema.ResourceData, m any
...
-	acl, err := s3Client.GetBucketAcl(ctx, &s3.GetBucketAclInput{
-		Bucket: aws.String(bucketName),
-	})
+	projectId, diags, ok := setProjectId(ctx, d, bucketName, s3Client, &diags)
+	if !ok {
+		return diags
+	}
+
+	if projectId != "" {
+		err = identity.SetRegionalIdentity(d, region, bucketName+"@"+projectId)
```
remyleone, line 766: We should have a SetRegionalIdentityWithProject in this context to be able to clearly identify the resource that are using this format

internal/services/object/bucket_acl.go
```diff
@@ -427,20 +438,33 @@ func flattenBucketACLAccessControlPolicyOwner(owner *s3Types.Owner) []any {
 func resourceBucketACLRead(ctx context.Context, d *schema.ResourceData, m any) diag.Diagnostics {
 	expectedBucketOwner := d.Get("expected_bucket_owner")
 
-	conn, region, bucket, acl, err := s3ClientWithRegionWithNameACL(ctx, d, m, d.Id())
+	conn, region, bucketName, err := s3ClientWithRegionAndName(ctx, d, m, d.Id())
 	if err != nil {
 		return diag.FromErr(err)
 	}
 
+	// Rm "@<PROJECT_ID>"
```
remyleone, line 446: Why is this comment present?

internal/services/object/bucket_acl.go
```diff
@@ -427,20 +438,33 @@ func flattenBucketACLAccessControlPolicyOwner(owner *s3Types.Owner) []any {
...
+	parts := strings.Split(bucketName, "@")
```
remyleone, line 447: This should be in a dedicated function

internal/services/object/bucket_acl.go
```diff
@@ -465,34 +482,44 @@ func resourceBucketACLRead(ctx context.Context, d *schema.ResourceData, m any) d
...
 }
 
 func resourceBucketACLUpdate(ctx context.Context, d *schema.ResourceData, m any) diag.Diagnostics {
-	conn, region, bucket, acl, err := s3ClientWithRegionWithNameACL(ctx, d, m, d.Id())
+	conn, _, bucketName, err := s3ClientWithRegionAndName(ctx, d, m, d.Id())
 	if err != nil {
 		return diag.FromErr(err)
 	}
 
+	// Rm "@<PROJECT_ID>"
```
remyleone, line 509: Why is this comment present ?

internal/services/object/bucket_server_side_encryption_configuration.go
```diff
@@ -81,11 +82,12 @@ func bucketServerSideEncryptionConfigurationSchema() map[string]*schema.Schema {
 				},
 			},
 		},
+		"project_id": account.ProjectIDSchema(),
```
remyleone, line 85: It should be at region level of indendation, not at rule level of identation, I would suggest moving it next to region.

remyleone (changes requested):

internal/services/object/bucket_server_side_encryption_configuration.go
pypaut: It is at `region` level of indentation 👍🏻

PR #4076, opened 2026-06-10
chore: fix MD047

remyleone:


pypaut (changes requested):
Missing updates in the `.md.tmpl` files

pypaut (approved):
Ok I got it wrong. The descriptions are actually used by the templates to build the final `.md` files.

PR #4103, opened 2026-06-15
feat(k8s): upgrade pool separately from cluster

### LINKED ISSUE #2993: Expose `upgrade_pools` when updating the version of a `scaleway_k8s_cluster`
### Community Note

### Description

Hello!

As of today, when we bump the version of a K8s cluster, `upgrade_pools` is always sent as `True`. This parameter is defined here in the K8s API: https://www.scaleway.com/en/developers/api/kubernetes/#path-clusters-upgrade-a-cluster

This is probably a healthy default, but depending on the exact Kubernetes setup being used, it may cause issues. Typically, some local data can be lost in the process, which is why there's a warning when enabling this parameter from the console. So users may want to pass `upgrade_pools: False` to the underlying API call.

The current workaround is to do the upgrade outside of Terraform, and then apply the changes to reconcile the configuration with the infra. This is OK, but it would be nice to be able to do everything from Terraform.

Thanks!

* scaleway_k8s_cluster

### Potential Terraform Configuration

(Please validate this with the K8s team ^^)

```hcl
resource "scaleway_k8s_cluster" "cluster" {
  ...
  
  upgrade_pools_on_version_changes = false // true by default, optional to keep existing behavior
}
```

### References

* #0000

Mia-Cross:


templates/resources/k8s_cluster.md.tmpl
```diff
@@ -8,16 +8,27 @@ page_title: "Scaleway: scaleway_k8s_cluster"
 
 {{ .Description }}
 
-{{ if .HasExamples }}
 ## Example Usage
 
-{{ range .ExampleFiles -}}
```
remyleone, line 14: You lose the ability to range meaning you will have to include all files explicitly

remyleone (approved):

PR #4109, opened 2026-06-17
feat(block): add snapshot identity

pypaut:


internal/services/block/snapshot.go
```diff
@@ -206,6 +211,11 @@ func ResourceBlockSnapshotRead(ctx context.Context, d *schema.ResourceData, m an
 
 	_ = d.Set("tags", snapshot.Tags)
 
+	err = identity.SetZonalIdentity(d, zone, snapshot.ID)
```
remyleone:
```suggestion
	err = identity.SetZonalIdentity(d, snapshot.Zone, snapshot.ID)
```

internal/services/block/snapshot.go
```diff
@@ -286,3 +296,19 @@ func ResourceBlockSnapshotDelete(ctx context.Context, d *schema.ResourceData, m
...
+
+func setSnapshotState(resourceData *schema.ResourceData, snapshot *block.Snapshot) {
+	if snapshot == nil {
+		return
+	}
+
+	_ = resourceData.Set("name", snapshot.Name)
+	_ = resourceData.Set("project_id", snapshot.ProjectID)
+	_ = resourceData.Set("tags", snapshot.Tags)
+
```
remyleone, line 299: I think snapshot.Zone is missing here as well as many other fields. Could you ensure that all field are fill up here.

remyleone (changes requested):

internal/services/block/snapshot.go
remyleone: Why is `setSnapshotState` not used here?

remyleone (changes requested):

remyleone (approved):

PR #4113, opened 2026-06-18
fix(domain): support import by domain name and handle archived tasks

jremy42:


docs/resources/domain_registration.md
```diff
@@ -136,10 +136,22 @@ Each contact block supports the following attributes:
...
+To import an existing domain registration, use the domain name:
+
+```bash
+terraform import scaleway_domain_registration.test <project_id>/<domain_name>
+```
+
+For a multi-domain registration, list every domain name separated by commas:
+
+```bash
+terraform import scaleway_domain_registration.test <project_id>/<domain1.com>,<domain2.com>
```
remyleone: I'm not sure to understand why project_id is required here. We don't use it in other resources such as servers or volumes, why would it be helpful here?

remyleone (changes requested):

docs/resources/domain_registration.md
jremy42: You're right, it wasn't needed — I've dropped it from the import.

jremy42: @remyleone

docs/resources/domain_registration.md
```diff
@@ -136,10 +136,24 @@ Each contact block supports the following attributes:
...
-To import an existing domain registration, use:
+To import an existing domain registration, use the domain name:
 
 ```bash
-terraform import scaleway_domain_registration.test <project_id>/<task_id>
+terraform import scaleway_domain_registration.test <domain_name>
 ```
 
-You can use the [scaleway_domain_registration](../data-sources/domain_registration.md) data source to look up the `task_id` and `project_id` by domain name.
+For a multi-domain registration, list every domain name separated by commas:
```
elwinar: Maybe a note about ordering being important (or not) could avoid someone asking questions later.

internal/services/domain/helpers.go
```diff
@@ -421,29 +390,73 @@ func FindTaskByDomain(ctx context.Context, registrarAPI *domain.RegistrarAPI, do
...
+			return nil, fmt.Errorf("invalid domain format in ID: %s", id)
+		}
+
+		return names, nil
+	}
+
+	pageSize := uint32(1000)
+
+	listTasksResponse, err := registrarAPI.ListTasks(&domain.RegistrarAPIListTasksRequest{
+		PageSize: &pageSize,
```
elwinar: Can be written `new(1000)` since 1.26 I think.

templates/resources/domain_registration.md.tmpl
```diff
@@ -137,10 +137,24 @@ Each contact block supports the following attributes:
...
 ```bash
-terraform import scaleway_domain_registration.test <project_id>/<task_id>
+terraform import scaleway_domain_registration.test <domain_name>
 ```
 
-You can use the [scaleway_domain_registration](../data-sources/domain_registration.md) data source to look up the `task_id` and `project_id` by domain name.
+For a multi-domain registration, list every domain name separated by commas:
+
+```bash
+terraform import scaleway_domain_registration.test <domain1.com>,<domain2.com>
```
estellesoulard, line 149: I might be missing it but I don't see any test on multi-domain import. If this is indeed the case could we add one?

templates/resources/domain_registration.md.tmpl
```diff
@@ -137,10 +137,24 @@ Each contact block supports the following attributes:
...
+
+Importing by task ID is also supported, as long as the registration task still exists:
+
+```bash
+terraform import scaleway_domain_registration.test <task_id>
+```
+
+The `project_id` is resolved automatically from the API and does not need to be provided in the import ID. A legacy `<project_id>/<domain_name>` or `<project_id>/<task_id>` format is still accepted for backward compatibility.
+
+Registration tasks are archived after some time; once archived, import by domain name instead. You can use the [scaleway_domain_registration](../data-sources/domain_registration.md) data source to look up the `task_id` by domain name.
```
estellesoulard: nit: i would switch these two sentences, otherwise we start by describing a case in which the task_id is no longer useable, then we explain how to fetch it (which then seems irrelevant in this case).
 
```suggestion
You can use the [scaleway_domain_registration](../data-sources/domain_registration.md) data source to look up the `task_id` by domain name. Registration tasks are archived after some time; once archived, import by domain name instead. 
```

templates/resources/domain_registration.md.tmpl
jremy42: We currently only have a single domain available in staging for these registration tests, so we can't reliably record a multi-domain import acceptance test yet...

estellesoulard: That's unfortunate :cry:

remyleone (approved):

PR #4132, opened 2026-06-25
fix(k8s): remove default values for new autoscaler fields

Mia-Cross:
It was reported that the default values that Terraform fills in for `skip_nodes_with_local_storage` and `log_level` do not always correspond to the default values from the API, which can cause unexpected changes (and Terraform wants to replace the resource since the attribute triggers a ForceNew).
To avoid this, this PR removes the default values from Terraform and lets the API decide.

### DIFF
--- a/internal/services/k8s/cluster.go
+++ b/internal/services/k8s/cluster.go
@@ -1094,14 +1094,12 @@ func autoscalerConfigSchema() *schema.Resource {
 			"skip_nodes_with_local_storage": {
 				Type:        schema.TypeBool,
 				Optional:    true,
-				Default:     true,
 				ForceNew:    true,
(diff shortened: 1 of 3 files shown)


internal/services/k8s/cluster.go
```diff
@@ -1094,14 +1094,12 @@ func autoscalerConfigSchema() *schema.Resource {
 			"skip_nodes_with_local_storage": {
 				Type:        schema.TypeBool,
 				Optional:    true,
-				Default:     true,
 				ForceNew:    true,
 				Description: "If true, the autoscaler will never delete nodes with pods with local storage, e.g. EmptyDir or HostPath, defaults to true.",
```
remyleone: the description should be changed if the default is not present anymore. Also the documentation should be updated as well.

remyleone (changes requested):

remyleone (approved):

PR #4141, opened 2026-06-29
fix: bound 403 retry by a 2-minute IAM propagation timeout

jremy42:
## Summary

IAM is eventually consistent: just after a resource is created, a granted permission can intermittently return 403 (even after a prior 200), causing flaky acceptance tests. This bounds the 403 retry by a 2-minute propagation window — a 403 that outlives it is treated as a real permission error.

## Changes

- `transport.RetryOn403` / `RetryOn403Value` are bounded by `IAMPropagationTimeout` (2 min) instead of a fixed retry count; they retry only on HTTP 403 and respect `ctx`. On VCR replay the wait collapses to 0.
- `rdb` `retryOn403` now delegates to `transport.RetryOn403`.
- Acceptance `Check` functions in cockpit & rdb call the API directly (outside provider CRUD) and were not guarded; their reads are now wrapped with `transport.RetryOn403`.

### DIFF
--- a/internal/services/rdb/waiters.go
+++ b/internal/services/rdb/waiters.go
@@ -2,46 +2,12 @@ package rdb
 
 import (
 	"context"
-	"errors"
-	"net/http"
 	"time"
 
 	"github.com/scaleway/scaleway-sdk-go/api/rdb/v1"
 	"github.com/scaleway/scaleway-sdk-go/scw"
 	"github.com/scaleway/terraform-provider-scaleway/v2/internal/transport"
 )
-
-var (
-	maxRetriesOnForbidden = 3
-	waitTime              = 1 * time.Second
-)
-
-// Mitigate transient permission issue during provisioning,
-// caused by IAM permissions propagation delta.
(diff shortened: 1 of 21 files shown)


internal/services/rdb/waiters.go
```diff
@@ -2,45 +2,18 @@ package rdb
...
-			case <-time.After(waitTime):
-				continue
-			}
-		}
-
-		return lastErr
-	}
-
-	return lastErr
+	return transport.RetryOn403(ctx, fn)
```
estellesoulard: nit: since this function no longer does anything more than `transport.RetryOn403`, maybe we could directly call `transport.RetryOn403` instead of wrapping it ?

estellesoulard (dismissed):

internal/services/rdb/waiters.go
jremy42: done, removed the wrapper and call transport.RetryOn403 directly

estellesoulard (approved):

PR #4147, opened 2026-07-01
fix(cockpit): grafana user datadog nightly

jremy42:


internal/services/cockpit/exporter_test.go
```diff
@@ -19,6 +19,10 @@ import (
 const defaultOrgIDPlaceholder = "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"
 
 func TestAccCockpitExporter_Basic_Datadog(t *testing.T) {
+	if *acctest.UpdateCassettes && os.Getenv("TF_TEST_DATADOG_API_KEY") == "" {
```
remyleone: Could all occurrences of this variable be pointing to the env package ? It would be helpful to visualize where this environment variable is used.

remyleone (changes requested):

remyleone (approved):

Issue #4118, opened 2026-06-19
Add support for key manager externally generated key


estellesoulard:
Add support for importing [externally generated key](https://www.scaleway.com/en/developers/api/key-manager/keys#import-key-material) to Key Manager. 
Since we are importing a key_material to a pre-existing key, this could be setup as a Terraform Action.
`key_material` is sensitive and should be a write only attribute (probably as well as `salt` attribute).

PR #4149, opened 2026-07-01
feat(keymanager): add key_material resource

estellesoulard:
https://github.com/scaleway/terraform-provider-scaleway/issues/4118

API has a create and delete function for this, it seemed relevant as a resource. No GET, so no need for a datasource nor import func.


docs/resources/key_manager_key_material.md
```diff
@@ -0,0 +1,87 @@
...
+
+
+## Argument Reference
+
+The following arguments are supported:
+
+- `key_id` - (Required, ForceNew) The ID of the key to import key material into. The key's origin must be external (UUID format). Can be a plain UUID or a regional ID.
+- `key_material` - (Optional, Sensitive, ForceNew) The key material to import. The key material is a random sequence of bytes used to derive a cryptographic key. Can be provided as raw bytes or a base64-encoded string (the provider will automatically normalize the input).
+- `key_material_wo` - (Optional, Write-only) The key material to import in write-only mode. The key material is a random sequence of bytes used to derive a cryptographic key. Must be provided as a base64-encoded string. The key material will not be stored in the Terraform state. Either `key_material` or `key_material_wo` must be specified.
+- `key_material_wo_version` - (Optional) Version number to track changes to the write-only key material. Increment this value to recreate the resource with new key material. Required when using `key_material_wo`.
```
Mia-Cross: Shouldn't this field be marked as ForceNew (I believe it recreates the resource) ?

internal/services/keymanager/key_material_resource.go
```diff
@@ -0,0 +1,424 @@
...
+		if decodeErr != nil {
+			resp.Diagnostics.AddError("Invalid salt format", decodeErr.Error())
+
+			return
+		}
+
+		importReq.Salt = &saltBytes
+	}
+
+	_, err = r.keyManagerAPI.ImportKeyMaterial(importReq)
```
Mia-Cross: Maybe we could pass the context ?

internal/services/keymanager/key_material_resource.go
```diff
@@ -0,0 +1,424 @@
...
+
+		return
+	}
+
+	deleteReq := &key_manager.DeleteKeyMaterialRequest{
+		Region: region,
+		KeyID:  keyID,
+	}
+
+	err = r.keyManagerAPI.DeleteKeyMaterial(deleteReq)
```
Mia-Cross: Maybe we could pass the context ?

docs/resources/key_manager_key_material.md
estellesoulard: Oops absolutely! Nice catch

estellesoulard: I reworked the doc and schema so that the relevant non-WriteOnly attributes are ForceNew (RequiresReplace() is the framework equivalent)

Mia-Cross (approved):

PR #4166, opened 2026-07-03
doc(container): add details about the form of `public_endpoint`

mormotomyia:
While using the new terraform version and fixing all deprecation notices, I noticed a breaking change in the behavior of the `public_endpoint`.

This update reflects the current behavior.

**I don't think this is a good change** - this makes creating domain records quite a hassle, as you have to pry apart the https:// from the domain by hand.

I would prefer having a dedicated domain record that can be used as a reference.


Mia-Cross:
Hello @mormotomyia,

The changes on `public_endpoint` / `domain_name` were required to stay consistent with the changes made in the Containers API. However, more information about the new form of the field is indeed welcome, thank you for your contribution !

I've made a few changes to comply with our CI, we'll merge your work as soon as it runs with success :+1:

mormotomyia:
Thanks for having a look!

I urge you to consider keeping the current field, as the change requires every user of the terraform provider to manipulate this field for DNS usage.

You're now putting this burden on the user - who will know less about your product design then you.
This is a counterintuitive breaking change that more people will run into.

Mia-Cross (approved):

Mia-Cross:
We will consider keeping the field in Terraform when the API stops returning it, thank you for your feedback !

PR #4183, opened 2026-07-08
fix(transport): jittered backoff for API rate limits

Spazzy757:
## Background

We hit sustained HTTP 429 rate limiting from the Scaleway API when applying with Terraform's default concurrency.

While diagnosing, I captured a **real** 429 from the Scaleway API by bursting concurrent read-only requests:

```
HTTP/2 429
x-envoy-ratelimited: true
server: Scaleway API Gateway (fr-par-3;edge03)
```

Scaleway's API Gateway is Envoy-based, and its rate-limit filter returns a bare 429 with **no timing guidance at all** — no `Retry-After`, no `X-RateLimit-*` headers. So header-driven backoff cannot help here; client-side backoff is the only lever.

## Problem

The retry transport left `go-retryablehttp`'s `DefaultBackoff` in place. Its fallback exponential backoff has **no jitter**. Under Terraform's concurrent execution (default `-parallelism=10`), every throttled request backs off in lock-step (2s → 4s → 8s …) and retries at the same instant — a thundering herd that immediately re-triggers the limit.

## Change

- Replace the backoff with a jittered exponential strategy (full jitter across `[min, min*2^n]`, capped at `RetryWaitMax`), wired via `Client.Backoff`. This spreads concurrent retries out and breaks the herd.
- Raise `RetryMax` from 3 to 5 so sustained throttling gets more attempts.
- `Retry-After` is still honored when present (standard, zero cost) in case an individual product API sets it — though the gateway does not.
- Add unit tests covering `Retry-After` parsing, jitter bounds/variance, the `x-envoy-ratelimited` fallback path, the `RetryWaitMax` cap, and zero-min safety.

`go build`, `go vet`, `gofmt`, and the transport tests all pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/internal/transport/retry.go
+++ b/internal/transport/retry.go
@@ -5,7 +5,10 @@ import (
 	"context"
 	"errors"
 	"io"
+	"math"
+	"math/rand"
 	"net/http"
+	"strconv"
 	"time"
 
 	"github.com/hashicorp/go-retryablehttp"
@@ -45,7 +48,7 @@ func NewRetryableTransportWithOptions(defaultTransport http.RoundTripper, option
 	c.HTTPClient = &http.Client{Transport: defaultTransport}
 
 	// Defaults
-	c.RetryMax = 3
+	c.RetryMax = 5
 	c.RetryWaitMax = 2 * time.Minute
 	c.Logger = logging.L
 	c.RetryWaitMin = time.Second * 2
@@ -56,6 +59,7 @@ func NewRetryableTransportWithOptions(defaultTransport http.RoundTripper, option
 
 		return retryablehttp.DefaultRetryPolicy(ctx, resp, err)
 	}
+	c.Backoff = rateLimitBackoff
 
 	// If ErrorHandler is not set, retryablehttp will wrap http errors
 	c.ErrorHandler = func(resp *http.Response, err error, _ int) (*http.Response, error) {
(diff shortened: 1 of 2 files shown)


internal/transport/retry.go
```diff
@@ -90,6 +94,66 @@ func NewRetryableTransport(defaultTransport http.RoundTripper) http.RoundTripper
 	return NewRetryableTransportWithOptions(defaultTransport, RetryableTransportOptions{})
 }
 
+// rateLimitBackoff computes how long to wait before the next retry.
+//
+// Scaleway's API Gateway (Envoy) returns 429 responses carrying no timing
```
remyleone, line 99: Hello I think new headers are now available:

Three headers are included with every response:
• x-ratelimit-limit – the limit, followed by the window in seconds. A limit of 20/h results in `20, 20;w=3600`.
• x-ratelimit-remaining – the number of requests remaining in the current window.
• x-ratelimit-reset – the number of seconds until the window resets.

A fourth header appears with a 429 response:
• retry-after – the number of seconds to wait before making another request.

Example based on a limit of 50 requests per second:
x-ratelimit-limit: 50, 50;w=1
x-ratelimit-remaining: 49
x-ratelimit-reset: 1

You can now handle rate limiting properly, without having to guess when the next request will be possible.

remyleone: Maybe the code could be simplified by taking those headers into account.

Spazzy757: just saw this! ill update it!

Issue #4184, opened 2026-07-08
Terraform function to get the real id from composite id


Nautigsam:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Add a Terraform function akin to `region_from_id` but to extract the real id from the composite id.

### New or Affected Resource(s)

All resources. I currently need it to get the real id of a secret to fetch it with the API at instance startup.

### Potential Terraform Configuration

```hcl
locals {
  SECRET_DB_PASSWORD = real_id_from_id(scaleway_secret.db_password.id)
}
```

estellesoulard:
Hello @Nautigsam! I have just added a `id_from_regional_id` function :). It will be available in the next provider release. Hope it helps!

Nautigsam:
That's great thanks!

PR #4186, opened 2026-07-08
feat: add support for id_from_regional_id function

estellesoulard:
https://github.com/scaleway/terraform-provider-scaleway/issues/4184


internal/functions/region_from_id_test.go
```diff
@@ -125,22 +125,13 @@ func TestAccProviderFunction_Region_From_ID(t *testing.T) {
...
-}
+				resource "scaleway_secret" "main" {
+					name = "terraform_test_region_from_id"
+				}
 
-output "region" {
-  value = provider::scaleway::region_from_id(scaleway_secret.main.id)
-}
+				output "region" {
+				value = provider::scaleway::region_from_id(scaleway_secret.main.id, null)
```
estellesoulard, line 133: About this change: I could not make this function in this current state, it seems like this test was faulty? 
I have tried to twist this in many directions to no avail. I can now quite confidently say that terraform functions cannot have optional parameters in the way initially presented in this test.

Mia-Cross: I experienced the same issues with the original form of the test, but this way works fine 👍

Mia-Cross (approved):

PR #4190, opened 2026-07-09
fix(instance): add support for listing all servers in scaleway_instance_servers data source

remyleone:
…ce_servers data source


estellesoulard (dismissed):

internal/services/instance/servers_data_source.go
```diff
@@ -197,7 +197,7 @@ func DataSourceInstanceServersRead(ctx context.Context, d *schema.ResourceData,
 		Name:    types.ExpandStringPtr(d.Get("name")),
 		Project: types.ExpandStringPtr(d.Get("project_id")),
 		Tags:    types.ExpandStrings(d.Get("tags")),
-	}, scw.WithContext(ctx))
+	}, scw.WithContext(ctx), scw.WithAllPages())
```
pypaut: Could you add an acceptance test that validates this specific addition? I.e. which would not pass without before this update.

pypaut (dismissed):

internal/services/instance/servers_data_source.go
remyleone: I would need to create more than 50 servers for it. Therefore no I don't think the cost of this test in regard to its scope justify it

PR #4206, opened 2026-07-20
feat(instance): add support for templates

Mia-Cross:


internal/services/instance/instance_template.go
```diff
@@ -0,0 +1,606 @@
...
+
+	tmpl, err := r.api.CreateTemplate(createReq, scw.WithContext(ctx))
+	if err != nil {
+		resp.Diagnostics.AddError("Failed to create Instance Template", err.Error())
+
+		return
+	}
+
+	state := flattenInstanceTemplate(ctx, tmpl, req, &resp.Diagnostics)
+	resp.Diagnostics.Append(resp.State.Set(ctx, new(state))...)
```
estellesoulard: Reading the docs, `new allocates a new, initialized variable and returns a pointer to it`.
Isn't the intention to set the state with the var initialized l.319 ? Something like:

```suggestion
	resp.Diagnostics.Append(resp.State.Set(ctx, &state)...)
```

Maybe I am misunderstanding though

internal/services/instance/helpers_framework.go
```diff
@@ -0,0 +1,253 @@
...
+
+func expandPrivateNetworks(ctx context.Context, networks types.Set, d *diag.Diagnostics) privateNetworkSpecs {
+	rawIDsList := scwtypes.ExpandRawIDSet(ctx, networks, "private_networks", d)
+
+	pnIDs := make([]string, 0, len(rawIDsList))
+	_ = append(pnIDs, rawIDsList...)
+
+	return privateNetworkSpecs{
+		privateNetworkIDs: rawIDsList,
+	}
```
estellesoulard: append() returns the updated slice, so here the call does not appear to be used? Should this be:
```suggestion
	updatedIDsList = append(pnIDs, rawIDsList...)

	return privateNetworkSpecs{
		privateNetworkIDs: updatedIDsList,
	}
```
?

internal/services/instance/instance_template.go
Mia-Cross: You're right, I changed it when I was grasping at straws while troubleshooting a bug, and forgot to change it back, thanks !

internal/services/instance/helpers_framework.go
Mia-Cross: The linter made me replace `pnIDs = append(pnIDs, rawIDsList...)` with this (-_-)' Nice catch 👍

internal/services/instance/template_test.go
```diff
@@ -0,0 +1,710 @@
...
+			{
+				ResourceName:      "scaleway_instance_template.main",
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
+		},
+	})
+}
+
+func TestAccInstanceTemplateResource_Filesystems(t *testing.T) {
```
remyleone, line 731: Could you add tests that show case an update in security_group_id, placement_group_id and windows_rdp_ssh_key_id?

internal/services/instance/template.go
```diff
@@ -0,0 +1,609 @@
...
+
+	if !plan.PublicIPV6Count.Equal(state.PublicIPV6Count) {
+		publicIPV6Count := plan.PublicIPV6Count.ValueInt32()
+		updateReq.PublicIPV6Count = new(uint32(publicIPV6Count))
+		hasChanges = true
+	}
+
+	if !plan.Volumes.Equal(state.Volumes) {
+		updateReq.UpdateVolumes = expandVolumes(ctx, plan.Volumes, &resp.Diagnostics).ToUpdateRequest()
+		hasChanges = true
```
remyleone, line 548: should a block like:
```go
if resp.Diagnostics.HasError() {
			return
		}
```
be also present?

internal/services/instance/template.go
```diff
@@ -0,0 +1,609 @@
...
+			return
+		}
+
+		resp.Diagnostics.AddError("Failed to read Instance Template", err.Error())
+
+		return
+	}
+
+	newState := flattenTemplate(ctx, tmpl, req, &resp.Diagnostics)
+	resp.Diagnostics.Append(resp.State.Set(ctx, new(newState))...)
```
estellesoulard: I think this is the same issue as in my previous comment (`new()` reinits `newState` so l.443 will be ignored)  https://github.com/scaleway/terraform-provider-scaleway/pull/4206#discussion_r3831484213

internal/services/instance/template.go
```diff
@@ -0,0 +1,638 @@
...
+	}
+
+	if tmpl.CreatedAt != nil {
+		model.CreatedAt = types.StringValue(tmpl.CreatedAt.Format("2006-01-02T15:04:05Z07:00"))
+	} else {
+		model.CreatedAt = types.StringNull()
+	}
+
+	if tmpl.UpdatedAt != nil {
+		model.UpdatedAt = types.StringValue(tmpl.UpdatedAt.Format("2006-01-02T15:04:05Z07:00"))
```
remyleone, line 390: Would it be easier to use `time.RFC3339` here?

remyleone (approved):

PR #4214, opened 2026-07-22
fix(keymanager): mark next_rotation_at as computed to stop rotation schedule drift

eschmar:
## What

- Mark `rotation_policy.next_rotation_at` on `scaleway_key_manager_key` as `Computed` (in addition to `Optional`) and validate it as an RFC 3339 date.
- On update, only send `next_rotation_at` to the API when it is explicitly set in the configuration, so changing `rotation_period` lets Key Manager recompute the schedule instead of re-sending the value carried over from state.
- Add an acceptance test covering: no diff on re-plan, `ImportStateVerify`, a `rotation_period`-only update, and an explicitly configured `next_rotation_at` round-tripping unchanged.
- Document `next_rotation_at` and its future-date constraint. Also correct the docs to list `rotation_period` as Required inside the block: the schema has always had `Required: true` there — no schema change — the docs just wrongly said Optional, and the API confirms a `rotation_policy` without `rotation_period` is rejected.

## Why

When a key has a `rotation_policy` without an explicit `next_rotation_at`, every plan showed `next_rotation_at -> null` because the attribute was `Optional` but not `Computed`. Each apply then called `UpdateKey` with `NextRotationAt = nil`, which the API interprets as "reschedule to now + rotation_period" — so every apply silently postponed the next rotation, and a regularly-applied key would effectively never rotate.

With this change the API-computed timestamp is retained in state when not configured, producing an empty plan and leaving the rotation schedule untouched.

Co-authored with Claude Code, feel free to discard silently if that is an issue. Hope this fits into the greater picture.

### DIFF
--- a/docs/resources/key_manager_key.md
+++ b/docs/resources/key_manager_key.md
@@ -91,7 +91,8 @@ The following arguments are supported:
     - `scaleway_kms` (default)
     - `external`
 - `rotation_policy` (Block, Optional) – Rotation policy for the key:
-    - `rotation_period` (String, Optional) – The period between key rotations (e.g., `"720h"` for 30 days).
+    - `rotation_period` (String, Required) – The period between key rotations (e.g., `"720h"` for 30 days).
+    - `next_rotation_at` (String, Optional) – The date and time of the next scheduled rotation, in RFC 3339 format. If not set, it is computed by the Key Manager API from `rotation_period`.
 
 ## Attributes Reference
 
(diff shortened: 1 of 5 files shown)


estellesoulard (approved):

estellesoulard:
Thank you very much for your contribution @eschmar !

estellesoulard:
Since we have a requirement on commits being signed we cannot merge it directly: I have cherry picked your commit and repushed it over at https://github.com/scaleway/terraform-provider-scaleway/pull/4217 , will merge from there :)

PR #4218, opened 2026-07-22
feat(annotations): add key resource and ds

estellesoulard:


internal/services/annotations/key_resource.go
```diff
@@ -0,0 +1,349 @@
...
+					stringplanmodifier.RequiresReplace(),
+					stringplanmodifier.UseStateForUnknown(),
+				},
+			},
+		},
+	}
+}
+
+func (r *AnnotationsKeyResource) IdentitySchema(ctx context.Context, req resource.IdentitySchemaRequest, resp *resource.IdentitySchemaResponse) {
+	resp.IdentitySchema = identityschema.Schema{
```
estellesoulard, line 90: Note: I started reworking the identity package to have a more homogeneous approach with a common interface in framework and sdkv2, but the diff was causing lots of changes across namespaces and getting way too large. It is also a whole topic in and of itself. To avoid any straying, I decided that I'll rework the identities in a future PR.

remyleone (approved):

PR #4220, opened 2026-07-23
fix: make region and zone computed attributes

### LINKED ISSUE #4201: `region` attribute omitted on resource causes downstream error at plan time
### Community Note

---

Disclaimer: this issue was redacted with the help of an LLM.

### Terraform Version

Terraform v1.15.7 / scaleway/scaleway 2.79.0

### Affected Resource(s)

* `scaleway_object_bucket` (and any resource using `regional.Schema()` for its `region` attribute)

```hcl
resource "scaleway_object_bucket" "app" {
  name          = "my-bucket"
  force_destroy = true
}

resource "coolify_environment_variable" "worker" {
  for_each = {
    STORAGE_REGION = scaleway_object_bucket.app.region
  }
  application_uuid = coolify_application.worker.uuid
  key              = each.key
  value            = each.value
}
```

### Expected Behavior

When `region` is omitted on the bucket, the provider should inject the default region (set on the provider itself) into the attribute, making it a known value at plan time — or at minimum, mark it as `Computed` so Terraform treats it as `(unknown after apply)` rather than `null`. Either way, downstream references via `for_each` / `each.value` should work.

### Actual Behavior

`scaleway_object_bucket.app.region` is `null` at plan time because `region` is `Optional` but **not** `Computed`. When this `null` value flows into a downstream resource via `for_each` → `each.value`, the downstream provider (here `coolify-terraform/coolify`, but the issue is provider-agnostic) sees `null` on a `Required` attribute and fails with:

```
Error: Missing Configuration for Required Attribute
  The attribute "value" is required, but no configuration was provided.
```

### Steps to Reproduce

1. Create a `scaleway_object_bucket` without setting `region`
2. Reference `scaleway_object_bucket.this.region` in a `for_each` or `values()` map on another resource
3. Run `terraform plan`

### Important Factoids

The root cause is in `internal/locality/regional/schemas.go`:

```go
func Schema() *schema.Schema {
    return &schema.Schema{
        Type:             schema.TypeString,
        Description:      "The region you want to attach the resource to",
        Optional:         true,
        ForceNew:         true,
        ValidateDiagFunc: verify.ValidateStringInSliceWithWarning(allRegions(), "region"),
        DiffSuppressFunc: locality.SuppressSDKNullAssignment,
    }
}
```

`Optional` without `Computed` causes the Terraform SDK v2 to assign `null` to the attribute when the user omits it. By contrast, `Computed` attributes get the sentinel `(unknown after apply)`, which Terraform correctly propagates through expressions and allows downstream `Required` attributes to remain pending until apply.

The fix would be to add `Computed: true` to the schema returned by `regional.Schema()`. This is safe because:

1. When the user provides a region explicitly (`region = "fr-par"`), `Optional` still takes precedence and the user's value is used.
2. When omitted, the provider will compute the region (from the provider config's default region) during Create/Read — exactly as it already does at runtime via `meta.ExtractRegion()`.
3. Existing configurations that omit `region` will see a one-time update in state as Terraform stores the computed value, but there is no plan diff or resource recreation.

### References

* Related discussion: `internal/locality/regional/schemas.go` line 20
* The only resource currently using `regional.ComputedSchema()` is `scaleway_lb` (`internal/services/lb/lb.go`), showing the pattern already exists in the codebase.

pypaut:
Fixes #4201 

- Make `region` a computed field in the default schema function
- Don't use the default schema function in the providers' schema, because "computed" cannot be added to the "framework" provider, thus creating a diff between the two providers


internal/services/object/bucket.go
```diff
@@ -280,7 +280,7 @@ func bucketSchema() map[string]*schema.Schema {
 				},
 			},
 		},
-		"region":     regional.Schema(),
+		"region":     regional.ComputedOptionalSchema(),
```
remyleone: Why only this attribute should be changed? What is specific about the region attribute for this resource?

internal/locality/regional/schemas.go
```diff
@@ -36,3 +36,17 @@ func Schema() *schema.Schema {
 		DiffSuppressFunc: locality.SuppressSDKNullAssignment,
 	}
 }
+
+// ComputedOptionalSchema returns a standard schema for a region, albeit
+// Computed and Optional
+func ComputedOptionalSchema() *schema.Schema {
```
remyleone: I think the documentation should be updated too.

remyleone (changes requested):

internal/services/object/bucket.go
pypaut: You mean why in this resource only? Or rather why only the `region` in this resource?

This PR aims at fixing #4201.

pypaut: Ended up applying to all the resources using `regional.Schema()`, with a little fix in [extractor.go](https://github.com/scaleway/terraform-provider-scaleway/pull/4220/changes#diff-944e6d3c1ffcdad0e424365b9cf1e00ee8f21c75d6772d55b247e1ce4d0b43bd) which was pinned down by all the RDB acceptance tests (no test "errors" per say, but a segfault).

internal/locality/regional/schemas.go
pypaut: Specified `Computed` and `Optional` to all the relevant resources.

provider/sdkv2.go
```diff
@@ -118,8 +116,16 @@ func SDKProvider(config *Config) plugin.ProviderFunc {
...
+				"region": {
+					Type:        schema.TypeString,
+					Optional:    true,
+					Description: "The region you want to attach the resource to",
+				},
+				"zone": {
+					Type:        schema.TypeString,
+					Optional:    true,
+					Description: "The zone you want to attach the resource to",
+				},
```
estellesoulard: why this change on the zone field ? the zonal schema was not changed in this PR, and it had additional validation, ForceNew, etc.

pypaut: Editing the `regional.Schema()` created issue with the provider configuration at runtime, which motivated me to separate the provider's `region` from the resources'. I thus did the same for `zone` for consistency, but removing the additional fields doesn't seem relevant. I will add them back 👍🏻

internal/locality/zonal/schemas.go
remyleone, line 34: Should Computed be also added here as well?

pypaut: Yes indeed. 🤌

remyleone (approved):

PR #4222, opened 2026-07-27
chore(tests): add namespaces checks on all test listings

estellesoulard:
- add missing namespaces to test lists
- add validation to ensure no namespace is missing from tests
- opensearch: rerecord cassettes
- s2s: fix tests (ip ranges were overlapping when running all tests)
- autoscaling: temporarily skip tests that use /instance-template and 501 (cc @Mia-Cross)
- datalab: skip tests until timeouts are fixed on API side (SPK-936)


internal/services/autoscaling/instance_group_data_source_test.go
```diff
@@ -8,6 +8,7 @@ import (
 )
 
 func TestAccDataSourceInstanceGroup_Basic(t *testing.T) {
+	t.Skip("autoscaling /instance-templates endpoint deprecated, skip until v1alpha2")
```
estellesoulard, line 11: Are these autoscaling skips ok with you @Mia-Cross ? I added them because rerecording the cassettes failed with 501s and I saw this API is undergoing transformations. Otherwise I can add an exception for the whole namespace in the tests matrix while v1alpha2 is in the works

estellesoulard: Discussed and confirmed :+1:

.github/scripts/check_tests_matrix.sh
```diff
@@ -0,0 +1,78 @@
...
+    service=$(basename "$service_path")
+    if ! is_excluded "$service"; then
+      echo "$service"
+    fi
+  done
+}
+
+# Check a matrix for missing services
+# Arguments: workflow_file matrix_services (newline-separated) matrix_name
+check_matrix() {
```
pypaut, line 33: What is a matrix exactly?

estellesoulard: We're talking about the product matrix at .github/workflows/acceptance-tests.yaml and .github/workflows/nightly.yml .

.github/workflows/acceptance-tests.yaml
```diff
@@ -17,12 +17,14 @@ jobs:
...
           - applesilicon
           - audittrail
+          - autoscaling
           - az
           - baremetal
           - billing
           - block
           - cockpit
           - container
+          # - datalab // timeouts, skipping until SPK-936 is fixed
```
pypaut: Why not use `t.Skip()` like some other tests?

pypaut (changes requested):

.github/workflows/acceptance-tests.yaml
estellesoulard: Both ways work. Here I'm only skipping the acctests and not the nightly, while `t.Skip` would skip all testing. Let me know if `t.Skip` is preferred !

pypaut: Okay bear with me... what is the difference between nighlies and acceptance tests? I thought they simply designated the job which runs acceptance tests each night.

estellesoulard: Acceptance tests run with the existing cassettes (no resource is created, API calls are mocked). Nightly tests run and actually make API calls without using the cassettes, they make actual API calls, create resources, etc

estellesoulard: I switched to `t.Skip`s :saluting_face:

pypaut: Ok clear, thank you. If the product is broken anyway, I believe it is better not to play with their actual API indeed

pypaut (approved):

Mia-Cross (approved):

PR #4226, opened 2026-07-30
fix(inference): use model_id in basic example

pypaut:


remyleone (changes requested):
Could you extract this example, put it in the examples folder and include it using a template ?

remyleone (approved):

PR #4252, opened 2026-08-07
feat(iam): add ephemeral api_key lifecycle handling

estellesoulard:
https://github.com/scaleway/terraform-provider-scaleway/issues/4320


internal/services/iam/api_key_ephemeral_resource.go
```diff
@@ -153,46 +196,229 @@ func (r *ApiKeyEphemeralResource) Open(ctx context.Context, req ephemeral.OpenRe
...
+			path.Root("organization_id"),
+			"Organization ID is required",
+			"Configure a default organization",
+		)
+
+		return
 	}
 
-	var err error
+	hasAnnotationIdentifier := !data.AnnotationIdentifier.IsNull() && !data.AnnotationIdentifier.IsUnknown()
```
estellesoulard, line 216: handle case where hasAnnotation was present but has been removed: should delete the annotation. + Add test.

PR #4260, opened 2026-08-14
fix(vpc): deprecate enable_routing and enable_custom_routes_propagation

yfodil:


estellesoulard (changes requested):
Bump the sdk-go with the API deprecation ? (Otherwise LGTM 🥳 )

remyleone (dismissed):

PR #4264, opened 2026-08-17
chore(object): ensure test buckets cleanup

estellesoulard:
- ensure all buckets follow sweeper namings
- ensure all tests creating buckets have the appropriate CheckDestroy


pypaut (changes requested):
While I agree with the new names, I don't understand how they were an issue for resource cleanup after tests. Can you explain?

Edit: `ensure all buckets follow sweeper namings` means that the tool that cleans the test resources filter on names, which you have fixed.

And all the bucket tests seem to have a valid `CheckDestroy`. Ok 🤌

pypaut (approved):

PR #4267, opened 2026-08-18
fix: #4266 k8s pool startup_taints overwrote taints on create

### LINKED ISSUE #4266: scaleway_k8s_pool: startup_taints overwrites taints on pool create (req.Taints instead of req.StartupTaints)
### Community Note

### Terraform Version

```
Terraform v1.15.7
provider registry.terraform.io/scaleway/scaleway v2.80.0
```

`master` is affected identically — same file, same line numbers, as of today.

### Affected Resource(s)

* `scaleway_k8s_pool`
* `scaleway_k8s_cluster` (same shape via `CreateClusterRequestPoolConfig`, not verified)

```terraform
resource "scaleway_k8s_pool" "runner" {
  cluster_id  = scaleway_k8s_cluster.this.id
  name        = "runner-az1"
  node_type   = "GP1-XS"
  size        = 0
  min_size    = 0
  max_size    = 6
  autoscaling = true

  taints {
    key    = "node"
    value  = "runner"
    effect = "NoSchedule"
  }

  startup_taints {
    key    = "example.com/registry-config"
    value  = "pending"
    effect = "NoSchedule"
  }
}
```

### Debug Output

Not attached — this is a code-path bug found by reading the provider source, and reproducing it live means deliberately creating a node pool whose regular taints are silently dropped. Happy to attach a `TF_LOG=DEBUG` gist showing the `CreatePool` request body if that is wanted; the relevant part is that the JSON carries `taints` populated from the `startup_taints` block and `startup_taints` empty.

### Expected Behavior

A pool created with both `taints` and `startup_taints` gets both: `taints` reconciled by Scaleway, `startup_taints` applied once at node join and not reconciled — which is what the two fields are documented to mean, and what the *update* path already does.

### Actual Behavior

On **create**, `startup_taints` overwrites `taints`, and `StartupTaints` is never set on the request.

https://github.com/scaleway/terraform-provider-scaleway/blob/v2.80.0/internal/services/k8s/pool.go#L432-L439

```go
	if taints, ok := d.GetOk("taints"); ok {
		req.Taints = expandCoreV1Taints(taints)
	}

	if startupTaints, ok := d.GetOk("startup_taints"); ok {
		req.Taints = expandCoreV1Taints(startupTaints)   // <-- req.Taints, should be req.StartupTaints
	}
```

`k8s.CreatePoolRequest` does have the right field, so this looks like a one-word typo rather than a missing capability:

```go
// StartupTaints: kubernetes taints applied at node creation but not reconciled afterwards.
StartupTaints []*CoreV1Taint `json:"startup_taints"`
```
— [`scaleway-sdk-go api/k8s/v1/k8s_sdk.go`, `CreatePoolRequest`](https://github.com/scaleway/scaleway-sdk-go/blob/master/api/k8s/v1/k8s_sdk.go)

So the pool comes up with the startup taints installed as **regular, reconciled** taints, and the regular taints gone entirely.

The **update** path is correct — `resourceK8SPoolUpdate` calls `SetPoolTaints` (`pool.go:734`) and `SetPoolStartupTaints` (`pool.go:745`) separately — so adding `startup_taints` to a pool that already exists behaves exactly as documented. Only create is affected.

### Steps to Reproduce

1. `terraform apply` the configuration above against an empty state, so the pool is **created** rather than updated.
2. Inspect the pool: `scw k8s pool get <pool-id>`.
3. Observe `taints` contains `example.com/registry-config=pending:NoSchedule` and does **not** contain `node=runner:NoSchedule`; `startup_taints` is empty.
4. `terraform plan` now shows a corrective diff, and a second `terraform apply` repairs it through the update path.

### Important Factoids

Why this is worth more than a cosmetic fix — the two consequences compound, and neither is visible without looking at the pool:

1. **The pool's real taint silently disappears.** In the example above `node=runner:NoSchedule` is gone, so arbitrary workloads schedule onto nodes meant to be dedicated.
2. **The startup taint becomes un-removable.** The whole point of `startup_taints` is "hold this node until something in-cluster clears it". Installed as a regular taint, Scaleway reconciles it back after every removal, so the in-cluster remover can never win and the nodes never become schedulable for the workload they were created for.

A second apply heals both, so the damage window is between the two applies — but it opens without warning, because it is not only reachable from a fresh pool. Any ForceNew field on an existing pool triggers a replacement, which goes through create: `node_type`, `zone`, `name`, `container_runtime`, `root_volume_type`, `public_ip_disabled`, `placement_group_id`. A routine node-type bump on a pool that uses both blocks is enough.

Suggested fix is the one-line assignment. A regression test asserting that a pool created with both blocks comes back with both would catch it — the existing coverage presumably exercises them separately, which is why the update path is right and this is not.

### References

* #4031 — the PR that added `labels`, `taints` and `startup_taints`

peter-svensson:
Fixes #4266.

### The fix

`ResourceK8SPoolCreate` assigns `startup_taints` to `req.Taints` instead of `req.StartupTaints`, so the second block overwrites the first:

```go
	if taints, ok := d.GetOk("taints"); ok {
		req.Taints = expandCoreV1Taints(taints)
	}

	if startupTaints, ok := d.GetOk("startup_taints"); ok {
		req.Taints = expandCoreV1Taints(startupTaints)   // before
	}
```

A pool created with both blocks therefore comes up with its regular taints **gone**, and the startup taints installed as **reconciled** taints — which inverts what `startup_taints` is for, since Scaleway then re-adds them after every removal and nothing in-cluster can clear them.

`resourceK8SPoolUpdate` was already correct (`SetPoolTaints` and `SetPoolStartupTaints` are separate calls), so only create is affected, and a second `terraform apply` repairs it through the update path.

### The test

`TestAccPool_TaintsAndLabels` could not have caught this, which is worth fixing along with the bug:

* **step 1 created** the pool with `taints` only, asserting `startup_taints.# == 0`
* **step 2 updated** it to add `startup_taints`

So every `startup_taints` assertion ran against the update path — the one that works — and the create path was only ever exercised without the second block, where the offending `if` never fires.

This PR adds a `startup_taints` block to step 1 and asserts both lists after creation. On `main` that step fails (`taints.#` comes back `0`); with the fix it passes.

### Acceptance test recording

I have not re-recorded `internal/services/k8s/testdata/pool-taints-and-labels.cassette.yaml`, so that cassette no longer matches the modified step 1 and will need re-recording on your side.

I would rather not record it from here: doing so means creating a real pool whose regular taints get silently dropped, and the only project I have available is one running production workloads that rely on a dedicated-node taint. `.github/contributing/acceptance_test.md` says best-effort tests are acceptable and that you will run them, so I have followed that — happy to adjust the test shape if you would prefer a separate test function with its own cassette instead of extending this one.

### Verified

* `go build ./...`, `go vet ./internal/services/k8s/`, `gofmt` clean.
* `go test -run TestAccPool_TaintsAndLabels ./internal/services/k8s/` compiles and skips without `TF_ACC`.
* The bug was found on `v2.80.0` and reproduced by reading `main`; both carry the same line.
* `k8s.CreatePoolRequest` does have a `StartupTaints` field, so this is the intended assignment rather than a missing capability.

### Scope

Only `scaleway_k8s_pool` exposes `startup_taints` — `grep -rn 'startup_taints\|StartupTaints' internal/services/k8s/*.go` shows no other resource is affected. (My issue speculated `scaleway_k8s_cluster` might share the shape via `CreateClusterRequestPoolConfig`; it does not, and I will correct that there.)

### DIFF
--- a/internal/services/k8s/pool.go
+++ b/internal/services/k8s/pool.go
@@ -435,7 +435,7 @@ func ResourceK8SPoolCreate(ctx context.Context, d *schema.ResourceData, m any) d
 	}
 
 	if startupTaints, ok := d.GetOk("startup_taints"); ok {
-		req.Taints = expandCoreV1Taints(startupTaints)
+		req.StartupTaints = expandCoreV1Taints(startupTaints)
 	}
 
 	// Validate pool configuration
--- a/internal/services/k8s/pool_test.go
+++ b/internal/services/k8s/pool_test.go
@@ -858,6 +858,12 @@ func TestAccPool_TaintsAndLabels(t *testing.T) {
 							value = "value2"
 							effect = "NoExecute"
 						}
+
+						startup_taints {
+							key = "startup-key1"
+							value = "startup-value1"
+							effect = "NoSchedule"
+						}
 					}`,
 				Check: resource.ComposeTestCheckFunc(
 					testAccCheckK8SClusterExists(tt, "scaleway_k8s_cluster.main"),
@@ -877,7 +883,12 @@ func TestAccPool_TaintsAndLabels(t *testing.T) {
 						"value":  "value2",
 						"effect": "NoExecute",
 					}),
-					resource.TestCheckResourceAttr("scaleway_k8s_pool.main", "startup_taints.#", "0"),
+					resource.TestCheckResourceAttr("scaleway_k8s_pool.main", "startup_taints.#", "1"),
+					resource.TestCheckTypeSetElemNestedAttrs("scaleway_k8s_pool.main", "startup_taints.*", map[string]string{
+						"key":    "startup-key1",
+						"value":  "startup-value1",
+						"effect": "NoSchedule",
+					}),
 				),
 			},
 			{
--- a/internal/services/k8s/testdata/pool-taints-and-labels.cassette.yaml
+++ b/internal/services/k8s/testdata/pool-taints-and-labels.cassette.yaml
@@ -10,64 +10,64 @@ interactions:
         host: api.scaleway.com
         headers:
             User-Agent:
-                - scaleway-sdk-go/v1.0.0-beta.7+dev (go1.26.0; linux; amd64) terraform-provider/develop terraform/terraform-tests
+                - scaleway-sdk-go/v1.0.0-beta.7+dev (go1.26.5; linux; amd64) terraform-provider/develop terraform/terraform-tests
         url: https://api.scaleway.com/k8s/v1/regions/fr-par/versions
         method: GET
       response:
         proto: HTTP/2.0
         proto_major: 2
         proto_minor: 0
-        content_length: 4355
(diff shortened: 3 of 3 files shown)


Mia-Cross:
Hello again,

The CI failed on its first run, but fixing it won't be a big deal and you won't have to do anything more. 

First, we use cassettes to record API interactions, so every time we modify a test, we have to re-record the test's cassette for the CI to replay. I added a commit with the cassette so this should be fixed :+1: 

Then, the rumdl linter caught some new problems that are not related to your changes so I opened a different PR to fix them (#4270). Once it has been merged, you can update your branch and it should be good to go !

Mia-Cross (approved):

Mia-Cross:
Never mind, the lint check is not mandatory so LGTM !

PR #4277, opened 2026-08-21
chore: sort namespaces alphabetically in framework.go

Mia-Cross:
This PR sorts all list of Resources, DataSources, Actions, etc. by alphabetical order of the namespaces for more clarity.

Also a pre-commit check picked up that the condition `err == nil` was always true, since a block `if err != nil`, that ends with a function exit, is called immediately before it.


pypaut (approved):
✨

PR #4281, opened 2026-08-21
feat(vpcgw): add srn attribute

estellesoulard:


internal/services/vpcgw/ip.go
```diff
@@ -149,6 +154,7 @@ func setIPState(d *schema.ResourceData, ip *vpcgw.IP) diag.Diagnostics {
 	_ = d.Set("zone", ip.Zone)
 	_ = d.Set("tags", ip.Tags)
 	_ = d.Set("reverse", ip.Reverse)
+	_ = d.Set("srn", ip.Srn)
```
pypaut, line 157: Could you add tests on this value?

pypaut (changes requested):

pypaut (approved):

PR #4282, opened 2026-08-21
feat(ipam/flexibleip): add srn attribute

estellesoulard:


internal/services/ipam/ip.go
```diff
@@ -309,6 +314,7 @@ func setIPAMIPState(d *schema.ResourceData, ip *ipam.IP, privateNetworkID string
 	_ = d.Set("updated_at", types.FlattenTime(ip.UpdatedAt))
 	_ = d.Set("is_ipv6", ip.IsIPv6)
 	_ = d.Set("region", ip.Region)
+	_ = d.Set("srn", ip.Srn)
```
pypaut, line 317: Could you add some tests on this value?

pypaut (changes requested):

pypaut (approved):

PR #4283, opened 2026-08-21
feat(interlink): add srn attribute

estellesoulard:


internal/services/interlink/routing_policy.go
```diff
@@ -186,6 +191,7 @@ func setRoutingPolicyState(d *schema.ResourceData, policy *interlink.RoutingPoli
 	_ = d.Set("is_ipv6", policy.IsIPv6)
 	_ = d.Set("prefix_filter_in", prefixFilterIn)
 	_ = d.Set("prefix_filter_out", prefixFilterOut)
+	_ = d.Set("srn", policy.Srn)
```
pypaut, line 194: Is it possible to add some tests to check for this value?

pypaut (changes requested):

pypaut (approved):

PR #4299, opened 2026-08-27
feat(autoscaling): add support for autoscaling groups

Mia-Cross:


internal/services/autoscaling/group.go
```diff
@@ -0,0 +1,562 @@
...
+		Zone:                          zone,
+		ProjectID:                     projectID,
+		Name:                          scwtypes.ExpandOrGenerateString(data.Name.ValueString(), "tf-asg"),
+		TemplateID:                    *templateID,
+		Tags:                          scwtypes.ExpandStringList(ctx, data.Tags, &resp.Diagnostics),
+		ScalingPolicySpec:             expandScalingPolicy(data.ScalingPolicy),
+		LoadBalancerConfigurationSpec: expandLoadBalancerConfiguration(ctx, data.LoadBalancerConfiguration, &resp.Diagnostics),
+	}
+
+	if resp.Diagnostics.HasError() {
```
remyleone, line 323: How can resp can gain errors between line 309 and this one?

internal/services/autoscaling/group.go
```diff
@@ -0,0 +1,562 @@
...
+	}
+
+	group, err := r.api.CreateGroup(createReq, scw.WithContext(ctx))
+	if err != nil {
+		resp.Diagnostics.AddError("Failed to create Autoscaling Group", err.Error())
+
+		return
+	}
+
+	groupSummaries, err := r.api.ListGroups(&autoscaling.ListGroupsRequest{
```
remyleone: After CreateGroup succeeds, the code calls ListGroups purely to emit a warning if the group isn't found in the requested zone (the Group response has no Zone field, hence the workaround). If that ListGroups call errors (transient API error, rate limit, etc.), the handler returns without writing the created group to state:

```go
group, err := r.api.CreateGroup(createReq, scw.WithContext(ctx))
// ...
groupSummaries, err := r.api.ListGroups(...)  // only used for a warning
if err != nil {
    resp.Diagnostics.AddError("Failed to list Autoscaling Groups", err.Error())
    return  // group exists server-side but is NOT saved to state
}
```

The group now exists on the API side but not in Terraform state. On the next terraform apply, Terraform will attempt to create it again, producing a duplicate that the user must clean up manually. The verification is only there to emit a non-fatal warning, so its failure should not abort state save. Suggested fix: on ListGroups error, emit the warning and fall through to flattenGroup/resp.State.Set using group from the create response and the requested zone (or just skip the verification on error). As a side note, this `ListGroups(..., WithAllPages())` after every create is the only resource in the repo doing this and is an extra paginated call worth reconsidering.

remyleone (dismissed):

internal/services/autoscaling/group.go
Mia-Cross: Line 320, a pointer to resp.Diagnostics is passed to expandLoadBalancerConfiguration, and is supposed to pick up errors that could occur while parsing `load_balancer_configuration`.

internal/services/autoscaling/group.go
Mia-Cross: I've modified the code to avoid returning before setting the state (send only warnings), and also prevent fetching all pages every time.

remyleone (approved):

PR #4325, opened 2026-09-03
chore(acceptance_tests): set default organization_id and project_id

estellesoulard:
Ensure the acceptance_tests environment is similar to the nightly environment in terms of defaults.

Additionally:
- rerecorded some cassettes (unrelated)
- fixed IAM SAML tests (previous implementation used a cert generated on the fly, which did not make it possible to replay the tests with existing cassettes. To circumvent the issue I hardcoded a test dummy cert)

This PR is blocking https://github.com/scaleway/terraform-provider-scaleway/pull/4252 (for which tests will simply not be able to run in the CI's acceptance tests without a default organization_id because of the added side-dependency to Annotations. Additionally, a few namespaces are impacted by the lack of defaults in the acctests, having to skip tests in this env, or fold them to make them work, which is not desirable at all)


.github/workflows/acceptance-tests.yaml
```diff
@@ -154,6 +154,8 @@ jobs:
           SCW_DEBUG: 0
           SCW_ACCESS_KEY: "SCWXXXXXXXXXXXXXFAKE"
           SCW_SECRET_KEY: "11111111-1111-1111-1111-111111111111"
+          SCW_DEFAULT_ORGANIZATION_ID: ${{ secrets.SCW_DEFAULT_ORGANIZATION_ID }}
```
estellesoulard, line 157: According to the docs, [we do not set SCW_DEFAULT_ORGANIZATION_ID as mandatory.](https://registry.terraform.io/providers/scaleway/scaleway/latest/docs#arguments-reference). 
Either: add it as mandatory in the docs, or remove it here (but since organization_id is essential to some core calls, I'd rather go with the former).

estellesoulard: Fixed by https://github.com/scaleway/terraform-provider-scaleway/pull/4332 : after discussion, SCW_DEFAULT_ORGANIZATION_ID will now be documented as mandatory.

pypaut (approved):

pypaut (approved):

remyleone (approved):

PR #4327, opened 2026-09-03
chore: add support for hashicorp skills

remyleone:


jremy42:
One suggestion: could we also register them in `AGENTS.md`? I think that would make the skills visible to agents, right?

remyleone:
Agents are supposed to find them on their own. Also it can be manually launched in most agents.

jremy42 (approved):

PR #4340, opened 2026-09-09
fix(rdb): increase engine upgrade timeout and document timeout recovery

jremy42:


internal/services/rdb/instance_upgrade_cleanup.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/scaleway/scaleway-sdk-go/scw"
+	"github.com/scaleway/terraform-provider-scaleway/v2/internal/httperrors"
+	"github.com/scaleway/terraform-provider-scaleway/v2/internal/locality/regional"
+)
+
+// deleteReplacedRDBInstance waits for the replaced (old) instance then deletes it.
+// A 404 is treated as success (already gone).
+func deleteReplacedRDBInstance(ctx context.Context, api *rdb.API, region scw.Region, instanceID string, timeout time.Duration) error {
+	_, err := waitForRDBInstance(ctx, api, region, instanceID, timeout)
+	if err != nil && !httperrors.Is404(err) {
```
elwinar: Shouldn't the 404 error lead to an early return ? 🤔 If the 404 indicates that the RDBInstance is already gone (as per the comment), you don't have anything to wait for ?

internal/services/rdb/instance_upgrade_cleanup.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	regionalID, ok := raw.(string)
+	if !ok || regionalID == "" {
+		return
+	}
+
+	region, instanceID, err := regional.ParseID(regionalID)
+	if err != nil {
+		tflog.Warn(ctx, fmt.Sprintf("Invalid replaced_from_instance_id %q, clearing attribute: %v", regionalID, err))
```
elwinar: Maybe it's the conventional way to handle it here, but I'm not a huge fan of removing it here, as it hides away an error somewhere in the system, assuming there is never a valid reason to have an invalid regional id.

internal/services/rdb/instance_upgrade_cleanup.go
```diff
@@ -0,0 +1,79 @@
...
+
+	if err := deleteReplacedRDBInstance(ctx, api, region, instanceID, timeout); err != nil {
+		tflog.Warn(ctx, fmt.Sprintf("Failed to cleanup replaced instance %s: %v", regionalID, err))
+
+		return
+	}
+
+	_ = d.Set("replaced_from_instance_id", "")
+
+	tflog.Info(ctx, "Successfully cleaned up replaced RDB instance "+regionalID)
```
elwinar: You should probably use a sprintf here to consistency with the other logs.

internal/services/rdb/instance_upgrade_cleanup.go
```diff
@@ -0,0 +1,79 @@
...
+		tflog.Warn(ctx, fmt.Sprintf("Invalid replaced_from_instance_id %q, clearing attribute: %v", regionalID, err))
+
+		_ = d.Set("replaced_from_instance_id", "")
+
+		return
+	}
+
+	tflog.Info(ctx, "Resuming cleanup of replaced RDB instance "+regionalID)
+
+	if err := deleteReplacedRDBInstance(ctx, api, region, instanceID, timeout); err != nil {
```
elwinar: Personal convention, but I dislike having side-effects in a if statement, it makes it harder to read both the happy path and the error handling.

internal/services/rdb/instance_upgrade_cleanup.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/scaleway/scaleway-sdk-go/api/rdb/v1"
+	"github.com/scaleway/scaleway-sdk-go/scw"
+	"github.com/scaleway/terraform-provider-scaleway/v2/internal/httperrors"
+	"github.com/scaleway/terraform-provider-scaleway/v2/internal/locality/regional"
+)
+
+// deleteReplacedRDBInstance waits for the replaced (old) instance then deletes it.
+// A 404 is treated as success (already gone).
+func deleteReplacedRDBInstance(ctx context.Context, api *rdb.API, region scw.Region, instanceID string, timeout time.Duration) error {
+	_, err := waitForRDBInstance(ctx, api, region, instanceID, timeout)
```
elwinar: Both this and the api.WaitForInstance function share the same parameters, almost the same name, error message. If they are different, maybe there is a naming issue.

internal/services/rdb/instance_upgrade_cleanup.go
jremy42: Good catch — fixed. On 404 we now return nil immediately instead of continuing to wait/delete.

internal/services/rdb/instance_upgrade_cleanup.go
jremy42: Agreed. We no longer clear replaced_from_instance_id on parse error — we warn and skip so the bad value stays visible in state.

internal/services/rdb/instance_upgrade_cleanup.go
jremy42: Makes sense for consistency

internal/services/rdb/instance_upgrade_cleanup.go
jremy42: Done

internal/services/rdb/instance_upgrade_cleanup.go
jremy42: Fair point. We now use only waitForRDBInstance for both waits, with comments: ready-before-delete vs wait-until-gone.

internal/services/rdb/instance.go
```diff
@@ -960,6 +965,13 @@ func readInstanceIntoState(ctx context.Context, d *schema.ResourceData, m any) d
 }
 
 func ResourceRdbInstanceRead(ctx context.Context, d *schema.ResourceData, m any) diag.Diagnostics {
+	rdbAPI, _, _, err := NewAPIWithRegionAndID(m, d.Id())
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	resumeReplacedRDBInstanceCleanup(ctx, d, rdbAPI, d.Timeout(schema.TimeoutDelete))
```
estellesoulard: I'm not sure to understand why this function is called from the `Read` and not only in the `Update`. Is the scenario:
- first apply, start update
- update timeouts on RDB side because of a large restore operation
- second apply, `Update` is not triggered so we have to catch it from the `Read`?

internal/services/rdb/instance_upgrade_cleanup.go
```diff
@@ -0,0 +1,83 @@
...
+		return fmt.Errorf("failed to delete old instance %s: %w", instanceID, err)
+	}
+
+	// Wait until the deleted instance disappears from the API.
+	_, err = waitForRDBInstance(ctx, api, region, instanceID, timeout)
+	if err != nil && !httperrors.Is404(err) {
+		return fmt.Errorf("error waiting for old instance %s deletion: %w", instanceID, err)
+	}
+
+	return nil
```
estellesoulard: Shouldn't we only return nil on 404 in this case? Here we would return nil if the resource is found, which is the opposite of  what the comment describes l.41

internal/services/rdb/instance.go
jremy42: Yes, that's the scenario: after a timed-out upgrade there may be no config diff, so Update won't run. Cleanup in Read lets us finish deleting the old instance on refresh/plan.

jremy42:
@estellesoulard  After discussion, I simplified the approach. Instead of resuming cleanup / tracking the old instance ID in state, we now only increase the default timeouts.update to 60m, document the blue/green timeout behavior, and return a clear error message if Terraform times out while waiting for the engine upgrade (the Scaleway workflow may still continue in the background).

docs/resources/rdb_instance.md
```diff
@@ -341,6 +343,67 @@ are of the form `{region}/{id}`, e.g. `fr-par/11111111-1111-1111-1111-1111111111
     - `forced_at` - Time when Scaleway-side maintenance will be applied.
     - `is_applicable` - Whether the maintenance can be applied by the user.
 
+## Engine upgrade timeout recovery
+
+Use this procedure when a blue/green `engine` upgrade times out in Terraform. The Scaleway upgrade may still finish in the background (snapshot, restore, endpoint migration). **Do not re-apply an engine change** while an upgrade is still running.
```
estellesoulard: `**Do not re-apply an engine change** while an upgrade is still running.` 
Since this sentence is really scary 😅 , can we specify what would happen in this scenario? (say if I already made that mistake and I want to know how much trouble I'm in)

jremy42: Fair point 😅 I clarified it: if you re-apply while the upgrade is still running and state still points to the old instance, Terraform can just kick off another upgrade and create a new extra instance. That’s the main risk.

elwinar (approved):

remyleone (approved):

PR #4343, opened 2026-09-10
feat(messageq): add deployment and user resources

jremy42:
## Summary
- Add `scaleway_messageq_deployment` and `scaleway_messageq_user` resources with regional/composite identity
- Add datasources for deployment, version, node type and certificate authority
- Include acceptance tests with VCR cassettes, examples and documentation


internal/services/messageq/types_test.go
```diff
@@ -0,0 +1,117 @@
+//nolint:testpackage // Tests need access to unexported messageq flatten/expand helpers.
```
remyleone, line 1: I would prefer to have flatten and expand helpers that are exported. I don't understand why messageq is special enough to desserve an exception.

provider/framework.go
```diff
@@ -255,6 +256,7 @@ func (p *ScalewayProvider) DataSources(_ context.Context) []func() datasource.Da
 		iam.NewScimDataSource,
 		iam.NewScimTokenDataSource,
 		kafka.NewVersionDataSource,
+		messageq.NewVersionDataSource,
```
remyleone, line 259: Why is there a mix between framework resources and SDK resources? Please use framework if possible.

internal/services/messageq/user.go
```diff
@@ -0,0 +1,538 @@
...
+}
+
+func (r *UserResource) Read(ctx context.Context, req resource.ReadRequest, resp *resource.ReadResponse) {
+	var (
+		state    userResourceModel
+		identity userResourceIdentityModel
+	)
+
+	resp.Diagnostics.Append(req.Identity.Get(ctx, &identity)...)
+	identityAvailable := !resp.Diagnostics.HasError() &&
```
remyleone, line 279: Could this long condition be extracted in a function and tested?

remyleone (approved):

PR #4344, opened 2026-09-10
feat(partner): add partner API resource

estellesoulard:
- Add resource for partner organization

A few notes:
- Partner organizations cannot be deleted. In the delete method, I opted to only remove the resource from the state.
- Since organizations cannot be deleted, and users must have unique email per API constraint, each test run requires a random email set at runtime in the test configuration. This had unfortunate consequences on cassettes replay, and I explored a few solutions to harmonize the test config and the cassettes. I finally opted to stay as close to the current `sensitiveFields` anonymization process, all while adding an optional scoping mechanism to fields per namespace. This impacted a few bricks in the anonymization process, like the leakchecks,  which now require verifying the namespace when necessary.
- Partner API is not available for the regular testing organization, hence the dedicated testing client and variables.


internal/services/partner/organization_resource_test.go
```diff
@@ -0,0 +1,221 @@
...
+					resource.TestCheckResourceAttr("scaleway_partner_organization.main", "owner_lastname", "Smith"),
+					resource.TestCheckResourceAttr("scaleway_partner_organization.main", "customer_id", "customer-456"),
+					resource.TestCheckResourceAttr("scaleway_partner_organization.main", "comment", "Updated organization"),
+				),
+			},
+		},
+	})
+}
+
+func TestAccPartnerOrganizationResource_WithPhoneNumber(t *testing.T) {
```
estellesoulard, line 116: Leaving this PR as draft for now:
API currently has a bug where some attributes are not handled properly. They are actually not relevant and will be removed in a later version. The following attributes might be removed: `siren_number`, `comment`, and `picture_link`.

Rerecord cassettes once this is fixed.

estellesoulard: Actually, I believe it's an OK compromise on our side to just leave these attributes out for now, and proceed with merging these resources. The attributes can be re-added later.

internal/services/partner/organization_resource_test.go
```diff
@@ -0,0 +1,240 @@
...
+func IsPartnerOrganizationLocked(tt *acctest.TestTools) resource.TestCheckFunc {
+	return func(s *terraform.State) error {
+		for _, rs := range s.RootModule().Resources {
+			if rs.Type != "scaleway_partner_organization" {
+				continue
+			}
+
+			partnerAPI := partnerSDK.NewAPI(meta.ExtractScwClient(tt.Meta))
+
+			err := retry.RetryContext(context.Background(), DestroyWaitTimeout, func() *retry.RetryError {
```
remyleone: Use tt.t.Context()

internal/services/partner/organization_resource.go
```diff
@@ -0,0 +1,396 @@
...
+			},
+			"email": schema.StringAttribute{
+				Required:            true,
+				MarkdownDescription: "The email of the new organization owner.",
+			},
+			"organization_name": schema.StringAttribute{
+				Required:            true,
+				MarkdownDescription: "The name of the organization you want to create. Usually the company name.",
+			},
+			"partner_id": schema.StringAttribute{
```
jremy42, line 88: Would it make sense to validate `partner_id` with `verify.IsStringUUID()`? That's the Framework pattern we use elsewhere (e.g. billing/IAM). Happy to leave it if you'd rather keep this loose for now.

```go
Validators: []validator.String{
    verify.IsStringUUID(),
},
```

internal/services/partner/organization_resource.go
```diff
@@ -0,0 +1,396 @@
...
+		MarkdownDescription: partnerOrganizationResourceDescription,
+		Attributes: map[string]schema.Attribute{
+			"id": schema.StringAttribute{
+				Computed:            true,
+				MarkdownDescription: "The ID of the organization resource.",
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
+				},
+			},
+			"email": schema.StringAttribute{
```
jremy42, line 77: Same idea for `email`: the format is known, and SDKv2 already has `verify.IsEmail()`. There's no Framework twin yet — maybe worth adding something like `verify.IsStringEmail()` (or another Framework validator) and wiring it here? Not a blocker if you'd rather follow up separately.

```go
Validators: []validator.String{
    verify.IsStringEmail(), // or equivalent
},
```

internal/services/partner/organization_resource_test.go
jremy42: Tiny nit: would it be better to pass `tt.T.Context()` here instead of `context.Background()`? That matches what we usually do in acceptance tests.

```go
err := retry.RetryContext(tt.T.Context(), DestroyWaitTimeout, func() *retry.RetryError {
```

internal/services/partner/organization_resource.go
```diff
@@ -0,0 +1,396 @@
...
+			"organization_name": schema.StringAttribute{
+				Required:            true,
+				MarkdownDescription: "The name of the organization you want to create. Usually the company name.",
+			},
+			"partner_id": schema.StringAttribute{
+				Optional:            true,
+				Computed:            true,
+				MarkdownDescription: "Your personal partner_id. This is the same as your Organization ID.",
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
```
jremy42, line 93: Small suggestion on `partner_id`: since it's `Optional` + `Computed` and has `RequiresReplace()`, it might be worth adding `UseStateForUnknown()` as well — same pattern as `organization_id` on the annotations key resource.

Without it, when `partner_id` is omitted (filled from the default org), the next plan can treat it as unknown again, and combined with `RequiresReplace()` that may look like a change and trigger a replace. Less visible when `partner_id` is set explicitly in HCL.

```go
PlanModifiers: []planmodifier.String{
    stringplanmodifier.RequiresReplace(),
    stringplanmodifier.UseStateForUnknown(),
},
```

A second plan-only step on `TestAccPartnerOrganizationResource_DefaultPartnerID` would be a nice way to catch a unexpected replace here.

internal/services/partner/organization_resource.go
estellesoulard: Oops yes I was indeed a bit light on validation here! Fixed

internal/services/partner/organization_resource.go
estellesoulard: Nice catch! Fixed and added relevant test.

internal/services/partner/organization_resource.go
```diff
@@ -0,0 +1,405 @@
...
+			},
+			"owner_firstname": schema.StringAttribute{
+				Required:            true,
+				MarkdownDescription: "The first name of the new organization owner.",
+			},
+			"owner_lastname": schema.StringAttribute{
+				Required:            true,
+				MarkdownDescription: "The last name of the new organization owner.",
+			},
+			"phone_number": schema.StringAttribute{
```
remyleone, line 108: phone_number is Optional without Computed, but Update only sends it when non-null. If a user removes it from config after setting it, the API keeps the old value, state keeps it, and Terraform shows a permanent diff. Fix: make it Optional: true, Computed: true (like partner_id), or send empty string on removal.

internal/services/partner/organization_resource.go
```diff
@@ -0,0 +1,405 @@
...
+	} else {
+		data.PhoneNumber = types.StringNull()
+	}
+
+	if organization.CreatedAt != nil {
+		data.CreatedAt = types.StringValue(organization.CreatedAt.String())
+	}
+
+	if organization.LockedAt != nil {
+		data.LockedAt = types.StringValue(organization.LockedAt.String())
```
remyleone, line 402: LockedAt has no else branch to set null when the API returns nil. A previously-locked then unlocked org keeps a stale timestamp. PhoneNumber handles this correctly a few lines above.

estellesoulard: Oops indeed, fixed

internal/services/partner/organization_resource.go
estellesoulard: made it optional computed

remyleone (approved):

PR #4358, opened 2026-09-15
ci(tem): sweep leftover TEM test DNS zones after revoking domains

jremy42:
## Summary
- Keep stable DNS subdomain names in TEM acceptance tests.
- Sweep leftover TEM test DNS zones on `scaleway-terraform.com` after revoking TEM domains so nightly runs no longer collide on leftovers.


internal/services/tem/testfuncs/sweep.go
```diff
@@ -46,4 +49,61 @@ func testSweepDomain(_ string) error {
...
+
+func isTEMTestDNSZone(subdomain string) bool {
+	switch subdomain {
+	case "test-blockedlist",
+		"test-autoconfig",
+		"test-autoconfig-update",
+		"webhook-test",
+		"test-reputation",
+		"validation-validation",
+		"validation-timeout":
```
estellesoulard: Wouldn't it be more efficient and future-proof to harmonize the test resources namings with proper prefixing instead, and use the `acctest.IsTestResource` here to only delete test resources like in other sweepers ?

jremy42: Good catch — done.

- Renamed TEM DNS subdomains to `tf-test-tem-*` so they match `acctest.IsTestResource`
- Sweeper now uses `acctest.IsTestResource` for both TEM domains and DNS zones (no hardcoded list)
- Updated the related cassettes accordingly

internal/services/tem/testfuncs/sweep.go
```diff
@@ -2,12 +2,15 @@ package temtestfuncs
 
 import (
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	domainSDK "github.com/scaleway/scaleway-sdk-go/api/domain/v2beta1"
```
estellesoulard: I'm just realizing now: why are these domain sweepers in `tem` namespace,  shouldn't they be in a dedicated `domain` sweeper?

internal/services/tem/testfuncs/sweep.go
```diff
@@ -44,6 +45,48 @@ func testSweepDomain(_ string) error {
 			}
 		}
 
+		return nil
+	})
+	if err != nil {
+		return err
+	}
+
+	return sweepTEMTestDNSZones()
```
estellesoulard: Once moved to the domain sweep.go, the `sweepTEMTestDNSZones` sweep function could be called as an independent sweeper for  `scaleway_domain_zone` (instead of calling it from another sweeper)

internal/services/tem/testfuncs/sweep.go
jremy42: done

internal/services/tem/testfuncs/sweep.go
jremy42: done

remyleone (approved):

PR #4364, opened 2026-09-16
feat(object): add support for custom s3 endpoint

pypaut:


internal/meta/extractors.go
```diff
@@ -72,6 +72,79 @@ func ExtractRegion(d terraformResourceData, m any) (scw.Region, error) {
 	return "", regional.ErrRegionNotFound
 }
 
+// ExtractS3UsePathStyle will retrieve the "s3_use_path_style" flag value from
+// the Terraform config file, in the "provider" block.
+func ExtractS3UsePathStyle(d terraformResourceData, m any) (s3UsePathStyle, ok bool) {
+	if d != nil {
+		rawConfigS3UsePathStyle, ok := GetRawConfigForKey(d, "s3_use_path_style", cty.Bool)
+		if ok && rawConfigS3UsePathStyle != "" {
```
estellesoulard: Unsure, but isn't `rawConfigS3UsePathStyle` a boolean there?

provider/framework.go
```diff
@@ -150,6 +182,28 @@ func modelToFrameworkConfig(model *ScalewayProviderModel) *meta.FrameworkProvide
...
 	}
 
+	if !model.Endpoints.IsNull() && !model.Endpoints.IsUnknown() {
+		config.Endpoints = make(map[string]string)
+
+		var endpoints []EndpointModel
+
+		diags := model.Endpoints.ElementsAs(ctx, &endpoints, false)
+		if diags.HasError() {
+			return config
```
estellesoulard: If we return here, no error is raised but S3UsePathStyle is not set even if present?

internal/meta/extractors.go
pypaut: Yes indeed. Well it's an `any`, so I replaced the check with `!= nil`.

provider/framework.go
pypaut: Indeed 😅. Switched strategy by skipping endpoints in case of error.

remyleone (approved):

PR #4366, opened 2026-09-16
feat(partner): add partner API actions

estellesoulard:
- Add Lock, Unlock, and Request Admin Role actions.


internal/services/partner/organization_lock_action_test.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+	tt, orgID := newPartnerTestTools(t)
+	defer tt.Cleanup()
+
+	email := partnerEmail(t)
+
+	resource.ParallelTest(t, resource.TestCase{
+		ProtoV6ProviderFactories: tt.ProviderFactories,
+		CheckDestroy:             IsPartnerOrganizationLocked(tt),
```
pypaut, line 32: Don't we want something like `IsPartnerOrganizationDeleted` instead?

internal/services/partner/organization_request_admin_role_action_test.go
```diff
@@ -0,0 +1,74 @@
...
+	tt, orgID := newPartnerTestTools(t)
+	defer tt.Cleanup()
+
+	email := partnerEmail(t)
+	newAdminEmail := partnerEmail(t)
+	password := partnerPassword(t)
+
+	resource.ParallelTest(t, resource.TestCase{
+		ProtoV6ProviderFactories: tt.ProviderFactories,
+		CheckDestroy:             IsPartnerOrganizationLocked(tt),
```
pypaut, line 39: Same here, expecting `IsPartnerOrganizationDeleted` as a test cleanup

internal/services/partner/organization_unlock_action_test.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	tt, orgID := newPartnerTestTools(t)
+	defer tt.Cleanup()
+
+	email := partnerEmail(t)
+
+	resource.ParallelTest(t, resource.TestCase{
+		ProtoV6ProviderFactories: tt.ProviderFactories,
+		CheckDestroy:             IsPartnerOrganizationLocked(tt),
```
pypaut, line 32: It's-a-me again

internal/services/partner/organization_unlock_action_test.go
```diff
@@ -0,0 +1,99 @@
...
+		partnerAPI := partnerSDK.NewAPI(meta.ExtractScwClient(tt.Meta))
+
+		organization, err := partnerAPI.GetOrganization(&partnerSDK.GetOrganizationRequest{
+			OrganizationID: rs.Primary.ID,
+		}, scw.WithContext(context.Background()))
+		if err != nil {
+			return fmt.Errorf("failed to get organization: %w", err)
+		}
+
+		if organization.LockedBy == partnerSDK.OrganizationLockedByPartner {
```
pypaut, line 93: What if the organization was locked by something else?

internal/services/partner/organization_request_admin_role_action_test.go
```diff
@@ -0,0 +1,74 @@
...
+						config {
+							organization_id = scaleway_partner_organization.main.id
+							username        = "testuser"
+							email           = "%s"
+							password        = "%s"
+						}
+					}
+				`, email, orgID, newAdminEmail, password),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttrSet("scaleway_partner_organization.main", "id"),
```
pypaut, line 69: Is there any way to actually check something like `HasAdminRole`?

internal/services/partner/organization_unlock_action_test.go
pypaut, line 93: Shouldn't we also check that `organization.LockedAt == nil`?

pypaut (changes requested):

internal/services/partner/organization_unlock_action_test.go
estellesoulard: no because a lock is still enforced by Scaleway at that point (since the created organization has just been created and not been verified by user email confirmation)

estellesoulard: The organization is locked by something else at that point. The Lock/Unlock call only impacts the Lock by the calling Partner. In this test, once created, the org is immediately locked by Scaleway (until the email of the newly created organization user is confirmed). Then the Partner locks it. Once we unlock it, it is still locked by Scaleway. But this Scaleway lock is beyond the scope of what the Partner can act on

internal/services/partner/organization_request_admin_role_action_test.go
estellesoulard: Partner organization cannot be deleted

internal/services/partner/organization_request_admin_role_action_test.go
estellesoulard: Only the owner is returned, not the admins :/ and `GET`ting the organization seems unfortunately nightmarish since this is a whole new separated organization for which we do not have access nor tokens from the get go

internal/services/partner/organization_request_admin_role_action_test.go
pypaut: Then the tests will create new organizations, without cleaning after? Does it mean we have to clean manually from time to time?

internal/services/partner/organization_unlock_action_test.go
pypaut: Ok, so all we can do is "check that the org is not locked by the partner", after the unlock. 👌🏻

internal/services/partner/organization_request_admin_role_action_test.go
pypaut: Ok :(

internal/services/partner/organization_request_admin_role_action_test.go
estellesoulard: I'm checking to see if we have any possibility to actually delete the organizations, but from the look of it it seems we are doomed to see them pile up :hidethepain:. Leaving this thread open until I have actual confirmation from the dedicated team.

internal/services/partner/organization_request_admin_role_action.go
```diff
@@ -0,0 +1,168 @@
...
+			},
+			"email": schema.StringAttribute{
+				Required:    true,
+				Description: "The member email.",
+				Validators: []validator.String{
+					verify.IsStringEmail(),
+				},
+			},
+			"password": schema.StringAttribute{
+				Required:    true,
```
pypaut, line 88:
```suggestion
				Required:    true,
				Sensitive:   true,
```

internal/services/partner/organization_request_admin_role_action.go
```diff
@@ -0,0 +1,168 @@
...
+)
+
+var (
+	_ action.Action              = (*PartnerOrganizationRequestAdminRoleAction)(nil)
+	_ action.ActionWithConfigure = (*PartnerOrganizationRequestAdminRoleAction)(nil)
+)
+
+type PartnerOrganizationRequestAdminRoleAction struct {
+	partnerAPI *partner.API
+	meta       *meta.Meta
```
pypaut: This field seems to be used for the `partnerAPI` creation only, during `Configure`. Can we drop it from the structure?

pypaut (changes requested):

internal/services/partner/organization_request_admin_role_action.go
estellesoulard: There's no `Sensitive` for `actions`' attributes :sweat_smile: . I studied `WriteOnly` as an alternative but it has an [existing limitation ](https://github.com/hashicorp/terraform/issues/37978)that makes it incompatible with actions as well. However, `actions` do not store anything in the state, so that seems harmless enough for me to leave it as is imo.

internal/services/partner/organization_request_admin_role_action.go
estellesoulard: Oops nice catch! I removed it from all 3 action structs.

pypaut (approved):

internal/services/partner/organization_request_admin_role_action_test.go
estellesoulard: As discussed, this is not possible for now, so let's merge as is (but I'll keep an eye out for future cleanup possibilities)

PR #4367, opened 2026-09-17
fix(cockpit): provision Grafana via IAM for grafana sync action test

jremy42:
## Summary
- Replace deprecated `scaleway_cockpit_grafana_user` setup in `TestAccActionCockpitGrafanaSyncDataSources_Basic` with IAM Grafana provisioning (`GetGrafana` + `GET /api/org` with `X-Auth-Token`) so the test can be rerecorded after Grafana users API stubbing.
- Rerecord the cassette (no password / grafana users interactions).
- Update action docs to describe IAM first-access provisioning.


docs/actions/cockpit_grafana_sync_data_sources.md
```diff
@@ -7,7 +7,7 @@ page_title: "Scaleway: scaleway_cockpit_grafana_sync_data_sources"
 
 The [`scaleway_cockpit_grafana_sync_data_sources`](https://registry.terraform.io/providers/scaleway/scaleway/latest/docs/actions/cockpit_grafana_sync_data_sources) action triggers the synchronization of all Cockpit data sources and the alert manager across regions into Grafana.
 
-Grafana must be provisioned for the project before running this action. With IAM authentication, access Grafana at least once for the project; the legacy `scaleway_cockpit_grafana_user` resource also provisions Grafana.
+Grafana must be provisioned for the project before running this action. With IAM authentication, access Grafana at least once for the project (for example `GET {grafana_url}/api/org` with header `X-Auth-Token: <secret_key>`).
```
estellesoulard, line 10: Would it be possible/worth adding a replacement action for that purpose, like an `activate_grafana_action` that makes this call from within terraform to spare users from having to perform a manual action?

jremy42: Good point. The first-access GET /api/org with IAM is what replaces the provisioning side-effect of the deprecated scaleway_cockpit_grafana_user. A dedicated activate_grafana action (or baking that call into grafana_sync_data_sources) would be a better UX than documenting a manual curl. I’d keep that as a follow-up — this PR is only unblocking the acceptance test after the Grafana users API stubbing.

estellesoulard: Works for me :) let's keeps that as a follow up, it could be a good addition imo.

estellesoulard (approved):

jremy42:
Closing as superseded by #4375 (`fix(cockpit): remove scaleway_cockpit_grafana_user after CreateGrafanaUser EOL`), which already landed the IAM Grafana provisioning for this test (including `/api/org` activation and a checked `Body.Close`). No remaining unique changes to keep from this PR.

